PR #1, opened 2015-10-09
Add index-services script

allait:
### Add index-services script
Updated version of the index_services.py scripts from
digitalmarketplace-api repository. New version:

* supports retries for all list services and indexing requests
* takes environment name instead of the two API endpoints
* can use any index name instead of the `g-cloud` default
* uses `dmutils` logging format

#### Add dmscripts.env helper to get API endpoint for given stage
Takes a stage name and returns the full API endpoint for Data or
Search APIs.

#### Add helper dmscript.logging functions
Used to configure 'dmutils' and 'script' logger with dmutils logging
handler. Logging levels can be modified for any additional loggers.

### DIFF
--- a/dmscripts/logging.py
+++ b/dmscripts/logging.py
@@ -1,0 +1,43 @@
+from __future__ import absolute_import
+
+import sys
+import logging
+from logging import DEBUG, INFO, WARNING, ERROR, CRITICAL  # noqa
+
+from dmutils.logging import CustomLogFormatter
+
+LOG_FORMAT = '%(asctime)s %(name)s %(levelname)s %(message)s'
+
+
+def configure_logger(log_levels=None):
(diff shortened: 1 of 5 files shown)


dmscripts/logging.py
```diff
@@ -0,0 +1,43 @@
...
+        logging.getLogger(logger_name).addHandler(handler)
+        logging.getLogger(logger_name).setLevel(level)
+
+    return logging.getLogger('script')
+
+
+def merge_log_levels(added_log_levels):
+    log_levels = {
+        'dmutils': INFO,
+        'script': INFO,
```
robyoung: Should this be `dmscripts`?

robyoung:
:+1:

PR #2, opened 2015-10-12
Update dmutils to 8.5.0

allait:
New SearchAPIClient index and alias related methods are required by
the index-services script.


robyoung:
:disappointed:  :+1:

PR #3, opened 2015-10-12
Skip index creation when alias exists with the same name

allait:
index-services is used for nightly index updates, during which it
uses the main alias as the request index. Trying to create an index
fails in this case since an alias with the same name already exists.

Index creation is skipped when --index is set to an alias name. Any
other error response during index creation will abort the script.


TheDoubleK:
:+1: looks good.

PR #4, opened 2015-10-19
Update generate user list script with interested suppliers endpoint

TheDoubleK:
Updates the user-list generation to use the new endpoint for getting suppliers registered for a framework.

Depends on this Utils PR being merged first:
 - [x] https://github.com/alphagov/digitalmarketplace-utils/pull/190

### DIFF
--- a/dmscripts/generate_user_email_list.py
+++ b/dmscripts/generate_user_email_list.py
@@ -1,18 +1,19 @@
 import csv
-import sys
-import multiprocessing
 from multiprocessing.pool import ThreadPool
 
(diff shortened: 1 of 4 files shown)


robyoung:
This is going to run considerably slower without the thread pool. That shouldn't be a problem but it may get annoying.

TheDoubleK:
But the thread pool was because there was an API call to check if each supplier is registered I think?  Now the script has a list of all suppliers so shouldn't need threading to make all those calls?  Or have I missed something?

allait:
:+1:

PR #5, opened 2015-10-27
Script to update API with pass/fail for suppliers

TheDoubleK:
For use when the pass/fail results for suppliers come in from CCS.

The script here may be subject to change as we don't yet know the format of data we'll get from CCS - this assumes a basic CSV with just `supplier_id` and `pass/fail` per row.  If we also get supplier name it would be good to modify the script to check the supplier name matches the ID before updating, as a sanity check.

### DIFF
--- a/dmscripts/insert_framework_results.py
+++ b/dmscripts/insert_framework_results.py
@@ -1,0 +1,32 @@
+from __future__ import unicode_literals
+import unicodecsv as csv
+
+from dmutils.apiclient.errors import HTTPError
+
+
+def insert_result(client, supplier_id, framework_slug, result, user):
+    try:
+        client.set_framework_result(supplier_id, framework_slug, result, user)
+        return "OK: {}\n".format(supplier_id)
+    except HTTPError as e:
+        return "Error inserting result for {} ({}): {}\n".format(supplier_id, result, str(e))
+
+
+def insert_results(client, output, framework_slug, filename, user):
+    with open(filename, 'rb') as csvfile:
+        reader = csv.reader(csvfile)
+        for index, row in enumerate(reader, start=1):
+            try:
+                supplier_id = int(row[0])
+                result = row[1].strip().lower()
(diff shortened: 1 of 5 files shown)


scripts/insert-results-for-framework.py
```diff
@@ -0,0 +1,30 @@
+"""
```
pcraig3: Probably `scripts/insert-results-for-framework.py` -> `scripts/insert-framework-results.py` so as to keep everything the same.

tests/test_insert_framework_results.py
```diff
@@ -0,0 +1,46 @@
...
+        call("OK: 234\n"),
+        call("OK: 345\n"),
+        call("OK: 456\n"),
+        call("Error: Result must be 'pass' or 'fail', not 'yes'; Bad line: 5\n"),
+        call("Error: invalid literal for int() with base 10: 'Company Name'; Bad line: 6\n"),
+        call("OK: 678\n")
+        ], any_order=False
+    )
+
+def test_hhtp_error_handling(mock_data_client):
```
pcraig3: http?

pcraig3:
Looks good once https://github.com/alphagov/digitalmarketplace-utils/pull/194 has gone through.

:+1:

scripts/insert-results-for-framework.py
TheDoubleK: Good spot - I missed that one when renaming.

dmscripts/insert_framework_results.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import unicode_literals
+import unicodecsv as csv
```
pcraig3: Also, this package isn't in the requirements.  Newest version is `unicodecsv==0.14.1`.

TheDoubleK: It wasn't there until very recently :)

PR #6, opened 2015-10-28
[#104991116] Generate framework agreement data for CCS

TheDoubleK:
See: https://www.pivotaltracker.com/story/show/104991116

There are some oddities here due to the way this has been written and re-written multiple times to get to this point. The script does what it should, but I'm happy to take comments to make it neater/better... as long as it's not a complete re-write (bearing in mind this is for one-time use and will not need to generalise to DOS or any other framework, as nobody should be doing PDF file generation from CSV files ever again in the future unless something goes horribly wrong).

* Rather than doing all the API calls itself the script uses output files from two earlier scripts in the API to build up the data - `generate_lots_export_for_ccs_sourcing` and `generate_declarations_export_for_ccs_sourcing` (these are the `<lots_file>` and `<declaration_file>` in the script command-line parameters).

* But it does now also do some API calls to check whether or not a supplier has been accepted onto the framework, so it also needs an API URL and token.

* The `check_lots_csv` and `check_declarations_csv` methods are left over from Martyn's original shot at this.  They're probably not necessary and could be deleted, but as we've got them it doesn't seem like a bad idea to do a basic sanity check on the file formats before proceeding.

* There are some tests.  If you would really like to see more then let me know.


dmscripts/generate_framework_agreement_data.py
```diff
@@ -0,0 +1,191 @@
...
+        sanitised_supplier_name = sanitised_supplier_name.replace(bad_char, '')
+    while '__' in sanitised_supplier_name:
+        sanitised_supplier_name = sanitised_supplier_name.replace('__', '_')
+    return "{}-{}".format(supplier_id, sanitised_supplier_name)
+
+
+def supplier_is_on_framework(client, supplier_id):
+    try:
+        framework_interest = client.get_supplier_framework_info(supplier_id, 'g-cloud-7')
+        if framework_interest['frameworkInterest']['onFramework']:
```
robyoung: `frameworkInterest` is correct (in that it's what the data API returns) but it shouldn't be.

TheDoubleK: What should it be do you reckon? We can fix the naming up before it starts being used too much. `supplierFramework` seems quite obscure.

robyoung: Yea, fair enough.

robyoung:
:+1:

robyoung:
Still :+1:

PR #8, opened 2015-11-02
Fix lots in export

TheDoubleK:
This is the version of the script that I used to generate the final files for CCS that they are happy with.

Compared to the previous version it:
* fixes the numbering of lots (PaaS and SaaS were the wrong way round)
* also generates a file for failed suppliers
* changes the key to have supplier name first

This now uses the new sanitise supplier name method and so depends on this being merged first: 
 - [ ] https://github.com/alphagov/digitalmarketplace-utils/pull/203


robyoung:
Could this make use of the sanitise function introduced in https://github.com/alphagov/digitalmarketplace-utils/pull/203 ?

TheDoubleK:
It could, and I did think that, but wasn't sure it was worth making the change given that we should never need to run this script again.  But that might have been me being lazy too, it will be a quick fix.

robyoung:
:+1:

PR #9, opened 2015-11-05
[#106095430] Script to all get users on framework

pcraig3:
This pull request adds a script that returns all the users who made an application to G-Cloud 7.
Specifically, it can return:

- All users of suppliers who made an application
- All users of suppliers who passed their application
- All users of suppliers who have not returned their framework agreement

Doesn't do anything else, but could be easily extended if it needed to.

[>> Story on Pivotal](https://www.pivotaltracker.com/story/show/106095430)


scripts/generate-user-email-list-applications.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+if __name__ == '__main__':
+    arguments = docopt(__doc__)
+
+    data_api_client = DataAPIClient(arguments['<data_api_url>'], arguments['<data_api_token>'])
+    output = open(arguments.get('--filename'), 'w+') if arguments.get('--filename') else sys.stdout
+    framework_slug = 'g-cloud-7'
+    on_framework = True if arguments.get('--on_framework') else None
+    agreement_returned = False if arguments.get('--agreement_not_returned') else None
```
robyoung: I think these will automatically be boolean by virtue of them being defined as flags.

dmscripts/generate_user_email_list_applications.py
```diff
@@ -0,0 +1,110 @@
+import csv
```
robyoung: Have you seen what @TheDoubleK has done over [here](https://github.com/alphagov/digitalmarketplace-scripts/blob/master/dmscripts/insert_framework_results.py#L2). `unicodecsv` allows us to handle unicode consistently between Python 2 and 3.

dmscripts/generate_user_email_list_applications.py
```diff
@@ -0,0 +1,110 @@
...
+            return supplier_framework.get('frameworkInterest')
+        except HTTPError:
+            return None
+
+    # filters Nones and entries with { 'onFramework': None } from a list
+    return [
+        supplier_framework for supplier_framework
+        in execute_concurrently(get_supplier_framework_info, enumerate(suppliers))
+        if (supplier_framework is not None) and (supplier_framework.get('onFramework') is not None)
+    ]
```
robyoung: This is a really nice, simple, obvious implementation!

dmscripts/generate_user_email_list_applications.py
```diff
@@ -0,0 +1,110 @@
...
+
+        return supplier_users
+
+    callback = partial(find_users_given_supplier_frameworks, len(supplier_frameworks))
+    # flattens a list of lists
+    return [
+        user for users_for_one_supplier
+        in execute_concurrently(callback, enumerate(supplier_frameworks))
+        for user in users_for_one_supplier
+    ]
```
robyoung: As is this

scripts/generate-user-email-list-applications.py
```diff
@@ -0,0 +1,24 @@
...
+from dmutils.apiclient import DataAPIClient
+from dmscripts.generate_user_email_list_applications import list_users
+
+
+if __name__ == '__main__':
+    arguments = docopt(__doc__)
+
+    data_api_client = DataAPIClient(arguments['<data_api_url>'], arguments['<data_api_token>'])
+    output = open(arguments.get('--filename'), 'w+') if arguments.get('--filename') else sys.stdout
+    framework_slug = 'g-cloud-7'
```
TheDoubleK: It might be good to pass `framework_slug` in as another parameter.  It seems wrong to add any new hard-coded `g-cloud-7` stuff while we are trying to unpick it from everywhere else.

dmscripts/generate_user_email_list_applications.py
TheDoubleK: Better to look here - for proper Python 2 / Python 3 compatibility I think you need to do conditional importing: https://github.com/alphagov/digitalmarketplace-scripts/blob/master/dmscripts/generate_framework_agreement_data.py#L7

dmscripts/generate_user_email_list_applications.py
```diff
@@ -0,0 +1,110 @@
...
+        Returns a supplier_framework or None if no supplier_framework entry exists
+        :param enumerated_supplier: tuple formatted (index, supplier)
+        :return: a supplier_framework or None
+        """
+        index, supplier = enumerated_supplier
+        if index % 100 == 0:
+            print('~{} suppliers'.format(index))
+        try:
+            supplier_framework = data_api_client.get_supplier_framework_info(supplier['id'], framework_slug)
+            return supplier_framework.get('frameworkInterest')
```
TheDoubleK: The `frameworkInterest` key has now changed now to `supplierFrameworks`.

scripts/generate-user-email-list-applications.py
pcraig3: Yeah, but we want 3 states, not 2.

For `agreement_returned`:

- `False` returns only those suppliers who haven't returned frameworks
- `True` would return only those suppliers who have returned frameworks 
- `None` doesn't discriminate.

So we can't really rely on the boolean logic for these, unless we change what `True` and `False` mean.

dmscripts/generate_user_email_list_applications.py
pcraig3: Not for the route I'm calling: https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/suppliers.py#L324

scripts/generate-user-email-list-applications.py
pcraig3: :weary: <sub>fine.</sub>

robyoung:
:+1:

TheDoubleK:
I still think we should drop the hard-coded  `g-cloud-7`, but we can always fix it later I guess...

PR #10, opened 2015-11-11
Add script to check for blacklisted words.

carolinegreen:
Add script to check for blacklisted words within g-cloud services

### DIFF
--- a/scripts/bad_words.py
+++ b/scripts/bad_words.py
@@ -1,0 +1,122 @@
+"""
+
+Usage:
+    scripts/bad_words.py <data_api_url> <data_api_token> <bad_words_path> <framework_slug>
+    <output_dir>
+"""
+
+import sys
+import os
+if sys.version_info > (3, 0):
+    import csv
+else:
+    import unicodecsv as csv
+import six
+import re
+from docopt import docopt
+from dmutils.apiclient import DataAPIClient
+
+
+def main(data_api_url, data_api_token, bad_words_path, framework_slug, output_dir):
+    client = DataAPIClient(data_api_url, data_api_token)
+    bad_words = get_bad_words(bad_words_path)
+    suppliers = get_suppliers(client, framework_slug)
+    check_services_with_bad_words(output_dir, framework_slug, client, suppliers, bad_words)
+
+
+def get_suppliers(client, framework_slug):
+    suppliers = client.find_framework_suppliers(framework_slug)
+    suppliers = suppliers["supplierFrameworks"]
+    if (framework_slug == "g-cloud-6"):
+        suppliers_on_framework = suppliers
+    else:
+        suppliers_on_framework = [supplier for supplier in suppliers if supplier["onFramework"]]
+    return suppliers_on_framework
+
+
+def get_draft_services(client, supplier_id, framework_slug):
+    services = client.find_draft_services(supplier_id, framework=framework_slug)
+    services = services["services"]
+    submitted_services = [service for service in services if service["status"] == "submitted"]
+    return submitted_services
+
+
+def get_services(client, supplier_id, framework_slug):
+    services = client.find_services(supplier_id, framework=framework_slug)
+    services = services["services"]
+    return services
+
+
+def get_bad_words(bad_words_path):
+    with open(bad_words_path) as file:
+        lines = file.readlines()
+        return [line.strip() for line in lines if not (line.startswith("#") or line.isspace())]
+
+
+def check_services_with_bad_words(output_dir, framework_slug, client, suppliers, bad_words):
+    if not os.path.exists(output_dir):
+        os.makedirs(output_dir)
+    with open('{}/{}-services-with-blacklisted-words.csv'.format(
+            output_dir, framework_slug), 'w') as csvfile:
+        fieldnames = [
+            'Supplier ID',
+            'Framework',
+            'Service ID',
+            'Service Title',
+            'Service Description',
+            'Blacklisted Word Location',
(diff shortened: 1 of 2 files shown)


scripts/bad_words.py
```diff
@@ -0,0 +1,67 @@
...
+
+def get_services_with_bad_words(client, suppliers, bad_words):
+    for supplier in suppliers:
+        if supplier["frameworkSlug"] == "g-cloud-6":
+            services = get_services(client, supplier["supplierId"], supplier["frameworkSlug"])
+        else:
+            services = get_draft_services(client, supplier["supplierId"], supplier["frameworkSlug"])
+        for service in services:
+            for key in service:
+                if isinstance(service[key], six.string_types):
```
quis: This will miss keys like `serviceFeatures`, which are lists of user-submitted strings.

scripts/bad_words.py
```diff
@@ -0,0 +1,67 @@
...
+                    for word in bad_words:
+                        if re.search(r"\b{}\b".format(word), service[key], re.IGNORECASE):
```
quis: `if word in service[key].lower()` would be simpler here.

scripts/bad_words.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def get_suppliers(client, framework_slug):
+    suppliers = client.find_framework_suppliers(framework_slug)
+    suppliers = suppliers["supplierFrameworks"]
+    if (framework_slug == "g-cloud-6"):
+        suppliers_on_framework = suppliers
+    else:
+        suppliers_on_framework = [supplier for supplier in suppliers if supplier["onFramework"]]
+    return suppliers_on_framework
```
quis: I think this could be
``` Python
return [
    supplier for supplier in suppliers if supplier["onFramework"] in [True, None]
]
```
without having to check the framework slug.

scripts/bad_words.py
```diff
@@ -0,0 +1,67 @@
...
+    services = services["services"]
+    submitted_services = [service for service in services if service["status"] == "submitted"]
+    return submitted_services
+
+
+def get_services(client, supplier_id, framework_slug):
+    services = client.find_services(supplier_id, framework=framework_slug)
+    services = services["services"]
+    return services
+
```
quis: `get_services` and `get_draft_services` could be one method that calls different methods on the API client depending on the framework slug.

scripts/bad_words.py
carolinegreen: Will look at that!

scripts/bad_words.py
carolinegreen: Wasn't sure which options were available.  As I couldn't find ones which were false, but if it's True on g-cloud-7, and just None on g-cloud-6, I can change it to that.  Thanks.

scripts/bad_words.py
carolinegreen: Yeah I wasn't sure if at some point, you wouldn't want to look at draft services for g-cloud-7 or other g-clouds either which was why I separated it out into two methods.  Thought it offered more flexibility.

robyoung:
One thing that would make it easier to pass the output of this to CCS is to make it output in CSV. [This script](https://github.com/alphagov/digitalmarketplace-scripts/blob/master/dmscripts/generate_framework_agreement_data.py) is a good example for outputting CSV in a way that can handle unicode in both Python 2 and Python 3.

quis:
You need to update `dmutils` to 13.2.0 to get the `apiclient.find_framework_suppliers` method.

scripts/bad_words.py
robyoung: We have G-Cloud 7 suppliers that have `onFramework` set to `None` so I think it's safest as it is.

carolinegreen: Okay, will leave it.

carolinegreen:
@robyoung  Wasn't sure what was being done with it, makes sense.  Will put it into a CSV.
@quis Why do we need the find_framework_suppliers method?

quis:
@carolinegreen Sorry, when I said update dmutils I meant in `requirements.txt`

carolinegreen:
@quis Oh right, it wasn't being kept as an old version for any particular reason.  Will update it.

scripts/bad_words.py
```diff
@@ -0,0 +1,101 @@
...
+                services = get_draft_services(client, supplier["supplierId"], supplier["frameworkSlug"])
+            for service in services:
+                for key in service:
+                    if isinstance(service[key], six.string_types):
+                        for word in bad_words:
+                            if re.search(r"\b{}\b".format(word), service[key], re.IGNORECASE):
+                                service_features = ""
+                                for feature in service["serviceFeatures"]:
+                                    service_features += feature + ","
+                                service_features = service_features[:-1]
```
robyoung: We don't actually ever check the service features for the bad words.

quis: Would it be better to flatten the whole service JSON and check that, or loop through each key and log that?

@carolinegreen `serviceFeatures` is one of a few keys that contain user-submitted content. Sorry I didn’t make that clear in my last comment.

carolinegreen:
It's going through all the keys now, outputting the service title, the key where the word was found, and the context for the word which I think is clearer.

carolinegreen:
It's going through all the keys now, outputting the service title, the key where the word was found, and the context surrounding the word found which I think is clearer.

robyoung:
Generated the file with production data and loaded it fine in Google Spreadsheets :+1:

PR #12, opened 2015-11-12
Make g cloud 7 live

robyoung:
This is the script for launching G-Cloud 7. Opening for early review.

### DIFF
--- a/scripts/make-g-cloud-7-live.py
+++ b/scripts/make-g-cloud-7-live.py
@@ -37,7 +37,7 @@ def find_submitted_draft_services(client, supplier, framework_slug):
     return (
         draft_service for draft_service in
         client.find_draft_services(supplier['supplierId'], framework=framework_slug)['services']
-        if draft_service['status'] == 'submitted'
+        if draft_service['status'] == 'submitted' and not supplier.get('serviceId')
     )
 
 
@@ -103,4 +103,4 @@ if __name__ == '__main__':
         draft_services = find_submitted_draft_services(client, supplier, FRAMEWORK_SLUG)
 
         for draft_service in draft_services:
-            make_draft_service_live(client, supplier, draft_service, framework_slug)
+            make_draft_service_live(client, supplier, draft_service, FRAMEWORK_SLUG)
--- a/scripts/make-g-cloud-7-live.py
+++ b/scripts/make-g-cloud-7-live.py
@@ -1,7 +1,7 @@
 """
 
 Usage:
-    scripts/make-g-cloud-7-live.py <data_api_url> <data_api_token> <draft_bucket> <documents_bucket> [--dry-run]
+    scripts/make-g-cloud-7-live.py <stage> <api_token> <draft_bucket> <documents_bucket> [--dry-run]
 """
 import sys
 sys.path.insert(0, '.')
@@ -11,11 +11,11 @@ try:
 except ImportError:
     import urllib.parse as urlparse
 
-from itertools import islice
 import random
 import re
 
 from docopt import docopt
+from dmscripts.env import get_api_endpoint_from_stage, get_assets_endpoint_from_stage
 from dmutils.apiclient import DataAPIClient
 from dmutils.s3 import S3
 
@@ -27,8 +27,7 @@ FRAMEWORK_SLUG = 'g-cloud-7'
 
 
 def assert_equal(one, two):
-    if one != two:
-        raise AssertionError("{} != {}".format(one, two))
+    assert one == two, "{} != {}".format(one, two)
 
 
 def find_suppliers_on_framework(client, framework_slug):
@@ -52,7 +51,6 @@ def parse_document_url(url, framework_slug):
     if not match:
         raise ValueError("Could not parse document URL {}".format(url))
     return {
-        'domain': urlparse.urlparse(url).netloc,
         'supplier_id': match.group(1),
         'draft_id': match.group(2),
         'document_name': match.group(3),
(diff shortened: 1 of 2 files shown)


scripts/make-g-cloud-7-live.py
```diff
@@ -0,0 +1,120 @@
...
+        supplier for supplier in client.find_framework_suppliers(FRAMEWORK_SLUG)['supplierFrameworks']
+        if supplier['onFramework']
+    )
+
+
+def find_submitted_draft_services(client, supplier, framework_slug):
+    return (
+        draft_service for draft_service in
+        client.find_draft_services(supplier['supplierId'], framework=framework_slug)['services']
+        if draft_service['status'] == 'submitted' and not supplier.get('serviceId')
```
TheDoubleK: Should this be `and not service.get('serviceId')` as suppliers don't have service IDs?

scripts/make-g-cloud-7-live.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def get_draft_document_path(parsed_document, framework_slug):
+    return '{framework_slug}/{supplier_id}/{draft_id}-{document_name}'.format(
+        framework_slug=framework_slug,
+        **parsed_document)
+
+
+def get_live_document_path(parsed_document, framework_slug, service_id):
+    return '{framework_slug}/documents/{supplier_id}/{service_id}-{document_name}'.format(
```
TheDoubleK: This is different from the path currently in production S3 - at the moment we have:
 - `documents/{supplier_id}/{service_id}-{document_name}` for G6 documents that have not been updated since G6 launch
 - `{framework_slug}/{supplier_id}/{service_id}-{document_name}-{timestamp}` for G5 and G6 documents that have been updated

So we probably want `{framework_slug}/{supplier_id}/{service_id}-{document_name}` without the extra `/documents/` in the middle?

scripts/make-g-cloud-7-live.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def make_draft_service_live(client, supplier, draft_service, framework_slug):
+    print("  > Migrating draft {} - {}".format(draft_service['id'], draft_service['serviceName']))
+    if DRY_RUN:
+        service_id = random.randint(1000, 10000)
+        print("    > generating random test service ID".format(service_id))
+    else:
+        service = client.publish_draft_service(draft_service['id'], 'make-g-cloud-7-live script')
+        service_id = service['id']
```
TheDoubleK: Looks like the key in the response here is `services`, so it should be `services['id']`: https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/drafts.py#L261

scripts/make-g-cloud-7-live.py
```diff
@@ -0,0 +1,120 @@
...
+                        draft_document_path, DRAFT_BUCKET.bucket_name))
+            message_suffix = '{}:{} to {}:{}'.format(
+                DRAFT_BUCKET.bucket_name, draft_document_path,
+                DOCUMENTS_BUCKET.bucket_name, live_document_path
+            )
+
+            if DRY_RUN:
+                print("    > not copying {}".format(message_suffix))
+            else:
+                DOCUMENTS_BUCKET.bucket.copy_key(live_document_path, DRAFT_BUCKET.bucket_name, draft_document_path)
```
TheDoubleK: Terrifying, but looks good.  Please can we make a copy of the entire uploaded draft documents bucket before we run this though, just for peace of mind?

scripts/make-g-cloud-7-live.py
```diff
@@ -0,0 +1,120 @@
...
+                print("    > copied {}".format(message_suffix))
+
```
TheDoubleK: We also need the next step here - to update the new published service with all the new document URLs.  We should probably build a dictionary of the 3/4 new URLs for the draft service and then update all document URLs with one call to `client.update_service` - what do you think?

scripts/make-g-cloud-7-live.py
robyoung: Done

scripts/make-g-cloud-7-live.py
robyoung: Ahh, yes, good spot! Thanks.

TheDoubleK:
OK let's do this... :+1:

scripts/make-g-cloud-7-live.py
```diff
@@ -0,0 +1,148 @@
+"""
+
+Usage:
+    scripts/make-g-cloud-7-live.py <data_api_url> <data_api_token> <draft_bucket> <documents_bucket> [--dry-run]
```
allait: Could we use `stage` here instead of the API URL, similar to https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/index-services.py#L117?

scripts/make-g-cloud-7-live.py
```diff
@@ -0,0 +1,148 @@
...
+
+DOCUMENT_KEYS = [
+    'pricingDocumentURL', 'serviceDefinitionDocumentURL',
+    'sfiaRateDocumentURL', 'termsAndConditionsDocumentURL',
+]
+FRAMEWORK_SLUG = 'g-cloud-7'
+
+
+def assert_equal(one, two):
+    if one != two:
```
allait: `assert one != two, "{} != {}".format(one, two)`

scripts/make-g-cloud-7-live.py
allait: @TheDoubleK `/documents/` in the middle is what we've agreed on during the URL discussions on that Google doc. This makes the paths more consistent between multiple buckets and makes the nginx dispatch rules simpler for the new URLs. We'll need to special case the old-style asset URLs though, but that's something we're aware of.

scripts/make-g-cloud-7-live.py
```diff
@@ -0,0 +1,148 @@
...
+        return 'assets.digitalmarketplace.service.gov.uk'
+
+
+def get_live_asset_url(parsed_document, live_document_path):
+    domain = get_live_asset_domain(parsed_document['domain'])
+    return 'https://{}/{}'.format(domain, live_document_path)
+
+
+def make_draft_service_live(client, supplier, draft_service, framework_slug):
+    print("  > Migrating draft {} - {}".format(draft_service['id'], draft_service['serviceName']))
```
allait: We could use the script logging helper instead of `print` https://github.com/alphagov/digitalmarketplace-scripts/blob/master/dmscripts/logging.py. This would also enable the API client logging, which could be useful

scripts/make-g-cloud-7-live.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def get_live_document_path(parsed_document, framework_slug, service_id):
+    return '{framework_slug}/documents/{supplier_id}/{service_id}-{document_name}'.format(
+        framework_slug=framework_slug,
+        service_id=service_id,
+        **parsed_document)
+
+
+def get_live_asset_domain(domain):
```
allait: This could also be part of the `stage` helper

scripts/make-g-cloud-7-live.py
allait: Better yet, can we use named arguments here, just so there's little doubt that there could be an error in the source/target order.

scripts/make-g-cloud-7-live.py
```diff
@@ -0,0 +1,148 @@
...
+            if DRY_RUN:
+                print("    > not copying {}".format(message_suffix))
+            else:
+                DOCUMENTS_BUCKET.bucket.copy_key(live_document_path, DRAFT_BUCKET.bucket_name, draft_document_path)
+                print("    > copied {}".format(message_suffix))
+
+    if DRY_RUN:
+        print("    > not updating document URLs {}".format(document_updates))
+    else:
+        client.update_service(service_id, document_updates, 'make-g-cloud-7-live script')
```
allait: I think we could have the actual "Moving documents to live bucket" update reason here

scripts/make-g-cloud-7-live.py
```diff
@@ -0,0 +1,148 @@
...
+        print("    > document URLs updated")
+
+
+if __name__ == '__main__':
+    arguments = docopt(__doc__)
+
+    client = DataAPIClient(arguments['<data_api_url>'], arguments['<data_api_token>'])
+    DRAFT_BUCKET = S3(arguments['<draft_bucket>'])
+    DOCUMENTS_BUCKET = S3(arguments['<documents_bucket>'])
```
allait: This is a minor point, but defining the global variables here sort of defeats the purpose of the `if __name__` condition. Functions rely on the variables being defined, so most of them won't work if the script is imported. Can we pass the buckets as arguments instead (similar to `client`)?

scripts/make-g-cloud-7-live.py
```diff
@@ -0,0 +1,148 @@
...
+    DRY_RUN = arguments['--dry-run']
+
+    suppliers = find_suppliers_on_framework(client, FRAMEWORK_SLUG)
+
+    for supplier in islice(suppliers, 50):
```
allait: Oops :fearful:

scripts/make-g-cloud-7-live.py
```diff
@@ -0,0 +1,148 @@
...
+            print("    > copied {}".format(message_suffix))
+
+    return copy_document
+
+
+def make_draft_service_live(client, copy_document, draft_service, framework_slug, dry_run):
+    print("  > Migrating draft {} - {}".format(draft_service['id'], draft_service['serviceName']))
+    if dry_run:
+        service_id = random.randint(1000, 10000)
+        print("    > generating random test service ID".format(service_id))
```
pcraig3: `{}`

PR #13, opened 2015-11-13
Add optional list of slugs of the frameworks to index as a parameter

TheDoubleK:
This is needed to prepare a search index in advance of G7 go-live + G5 expires.

### DIFF
--- a/scripts/index-services.py
+++ b/scripts/index-services.py
@@ -112,6 +112,9 @@ def do_index(search_api_url, search_api_access_token, data_api_url, data_api_acc
 
 if __name__ == "__main__":
     arguments = docopt(__doc__)
+    chosen_frameworks = arguments['--frameworks']
+    if not isinstance(chosen_frameworks, basestring):
+        chosen_frameworks = None
     ok = do_index(
         data_api_url=get_api_endpoint_from_stage(arguments['<stage>'], 'api'),
         data_api_access_token=arguments['--api-token'],
@@ -119,7 +122,7 @@ if __name__ == "__main__":
         search_api_access_token=arguments['--search-api-token'],
         serial=arguments['--serial'],
         index=arguments['--index'],
-        frameworks=arguments['--frameworks']
+        frameworks=chosen_frameworks
     )
 
     if not ok:
--- a/scripts/index-services.py
+++ b/scripts/index-services.py
@@ -6,12 +6,11 @@ Usage:
 
     --serial        Do not run in parallel (useful for debugging)
     --index=<index>  Search API index name [default: g-cloud]
-    --frameworks=<frameworks> Optional list of framework slugs that should be indexed
+    --frameworks=<frameworks> Optional comma-separated list of framework slugs that should be indexed
 
 Example:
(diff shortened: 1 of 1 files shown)


scripts/index-services.py
```diff
@@ -6,10 +6,12 @@
 
     --serial        Do not run in parallel (useful for debugging)
     --index=<index>  Search API index name [default: g-cloud]
+    --frameworks=<frameworks> Optional list of framework slugs that should be indexed
 
 Example:
-    ./index-services.py dev --api-token=myToken --search-api-token=myToken
+    ./index-services.py dev --api-token=myToken --search-api-token=myToken --frameworks="['g-cloud-6','g-cloud-7']"
```
allait: I think we could do a comma-separated list here, similar to other scripts

scripts/index-services.py
```diff
@@ -94,7 +98,7 @@ def do_index(search_api_url, search_api_access_token, data_api_url, data_api_acc
         pool = multiprocessing.Pool(10)
         mapper = pool.imap
 
-    indexer = ServiceIndexer(search_api_url, search_api_access_token, index)
+    indexer = ServiceIndexer(search_api_url, search_api_access_token, index, frameworks)
```
allait: I don't think this is right. API won't return non-live framework services by default, so the change should be in the find services part, not the indexer

scripts/index-services.py
```diff
@@ -64,7 +67,8 @@ def __call__(self, service):
 
     @backoff.on_exception(backoff.expo, apiclient.HTTPError, max_tries=5)
     def index_service(self, client, service):
-        if service['status'] == 'published':
+        if service['status'] == 'published' and \
```
allait: I don't think we need the additional check here. Once the `find_services` call is updated it should only return the services for the given frameworks.

scripts/index-services.py
TheDoubleK: The API now returns everything except expired frameworks.

scripts/index-services.py
TheDoubleK: Yes, don't know why I didn't do that in the first place.

scripts/index-services.py
allait: I don't think it does: https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/services.py#L47

TheDoubleK: Ah, sorry you're right - I've had it all working locally because I set G7 live.

scripts/index-services.py
```diff
@@ -27,14 +28,14 @@
...
+def request_services(api_url, api_access_token, frameworks):
 
     data_client = apiclient.DataAPIClient(
         api_url,
         api_access_token
     )
 
-    for service in data_client.find_services_iter():
+    for service in data_client.find_services_iter(framework=frameworks):
         yield service
```
robyoung: Can we just `return data_client.find_services_iter(framework=frameworks)` here?

robyoung:
I've left one minor, pedantic comment which shouldn't block this being merged :+1:

TheDoubleK:
No problem, Rob :)

PR #14, opened 2015-11-17
Add script to fix agreement uploads

robyoung:
Some signed agreements were uploaded as PDFs when the files were not
PDFs.

### DIFF
--- a/scripts/oneoff/fix-agreement-extensions.py
+++ b/scripts/oneoff/fix-agreement-extensions.py
@@ -1,0 +1,129 @@
+"""
+
+Usage:
+    scripts/oneoff/fix-agreement-extensions.py <stage> <api_token> <download_directory> [--dry-run]
+"""
+import sys
+sys.path.insert(0, '.')
+
+import os
+import re
+
+import magic
+from docopt import docopt
+
+from dmutils.apiclient import DataAPIClient
+from dmutils.s3 import S3
+from dmscripts.env import get_api_endpoint_from_stage
+
+
+def get_all_pdfs(download_directory):
+    for root, subfolder, files in os.walk(download_directory):
+        for filename in files:
+            if filename.endswith('.pdf') and not filename.startswith('2015-11-'):
+                yield os.path.join(root, filename)
+
+
+def get_filetype(path):
+    return magic.from_file(path, mime=True)
+
+
+def is_empty(path):
+    stat = os.stat(path)
+    return stat.st_size == 0
+
+
+def get_supplier_id_from_path(path):
+    match = re.search(r'/(\d+)/', path)
+    if not match:
+        raise ValueError("Could not find supplier ID in path {}".format(path))
+    return match.group(1)
+
+
+def handle_path(client, bucket, dry_run, path):
+    if is_empty(path):
+        show_contact_details(client, dry_run, path)
+    else:
+        filetype = get_filetype(path)
+        if filetype != b"application/pdf":
+            update_file_extension(client, bucket, dry_run, path, filetype)
+
+
+def show_contact_details(client, dry_run, path):
+    supplier_id = get_supplier_id_from_path(path)
+    if dry_run:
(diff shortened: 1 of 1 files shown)


scripts/oneoff/fix-agreement-extensions.py
```diff
@@ -0,0 +1,127 @@
...
+        b"image/jpeg": "jpg",
+        b'application/vnd.openxmlformats-officedocument.wordprocessingml.document': 'docx',
+    }.get(filetype)
+    if not extension:
+        raise ValueError("Unknown file type: {}".format(filetype))
+    return extension
+
+
+def get_path_in_s3(path):
+    return "{}/{}".format('g-cloud-7', path.split('/g-cloud-7/')[1])
```
allait: Isn't this just `path.lstrip('/')`? Or am I missing something?

scripts/oneoff/fix-agreement-extensions.py
```diff
@@ -0,0 +1,127 @@
...
+    else:
+        print(
+            "Copying {} to {} for supplier {}".format(
+                path_in_s3, new_path, supplier_id))
+        bucket.bucket.copy_key(
+            new_path,
+            src_bucket_name=bucket.bucket_name,
+            src_key_name=path
+        )
+        client.register_agreement_returned(
```
allait: Will this change the agreement return date? It'd be nice to keep the actual dates since we might use them to check who we've sent emails to etc.

scripts/oneoff/fix-agreement-extensions.py
```diff
@@ -0,0 +1,127 @@
...
+        show_contact_details(client, dry_run, path)
+    else:
+        filetype = get_filetype(path)
+        if filetype != b"application/pdf":
+            update_file_extension(client, bucket, dry_run, path, filetype)
+
+
+def show_contact_details(client, dry_run, path):
+    supplier_id = get_supplier_id_from_path(path)
+    if dry_run:
```
TheDoubleK: Not sure we need a dry run on this method as it isn't changing anything, but doesn't hurt I guess.

scripts/oneoff/fix-agreement-extensions.py
```diff
@@ -0,0 +1,127 @@
...
+        print("Empty file for {} - {}".format(supplier_id, os.path.basename(path)))
+    else:
+        supplier = client.get_supplier(supplier_id)['suppliers']
+        print(
+            "Empty file for {} {}".format(
+                supplier_id,
+                supplier['contactInformation'][0]['email']))
```
TheDoubleK: I don't think this is the best email address to use - this is the email address shown to buyers who might want to buy something.  I reckon the Digital Marketplace user account email addresses for the supplier would be a better bet.

Some suppliers also specified a different email in their declaration as the "Primary contact for the tender" from *any* of their marketplace account emails (I know!), so it might be worth pulling the primary email from the declaration (it's `SQ1-2b`) as well as the users.

scripts/oneoff/fix-agreement-extensions.py
TheDoubleK: I think the point is to force a date change, so that they appear back at the top of CCS's list of returned docs in the admin app.

scripts/oneoff/fix-agreement-extensions.py
robyoung: No, this is based off a synced local copy of the bucket so it's stripping off the local filesystem specific stuff as well.

scripts/oneoff/fix-agreement-extensions.py
robyoung: It was so I could hack on the train without a network connection.

scripts/oneoff/fix-agreement-extensions.py
robyoung: I'm wondering if it's best to just pull out the supplier id at this stage and then use whatever email we use when we're contacting interested suppliers.

TheDoubleK:
:+1:

PR #15, opened 2015-11-24
Add script to get DOS urls

carolinegreen:
Values are hard coded as this script is currently being used for very small specific task of getting
DOS urls only.

### DIFF
--- a/scripts/get-service-editing-urls.py
+++ b/scripts/get-service-editing-urls.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python
+"""
+Get service editing urls for each page in DOS service flow.
+
+Usage:
+    scripts/get-service-editing-urls.py <framework_location>
+
+"""
+from dmutils import content_loader
+from docopt import docopt
+
+
+def main(framework_location):
+    lots = ['digital-outcomes', 'digital-specialists', 'user-research-studios',
+            'user-research-participants']
+    loader = content_loader.ContentLoader(framework_location)
+    loader.load_manifest('digital-outcomes-and-specialists', 'services', 'edit_submission')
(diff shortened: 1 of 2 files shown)


scripts/get-service-editing-urls.py
```diff
@@ -0,0 +1,25 @@
...
+
+def main():
+    lots = ['digital-outcomes', 'digital-specialists', 'user-research-studios',
+            'user-research-participants']
+    loader = content_loader.ContentLoader('../digitalmarketplace-frameworks')
+    loader.load_manifest('digital-outcomes-and-specialists', 'services', 'edit_submission')
+    manifest = loader.get_manifest('digital-outcomes-and-specialists', 'edit_submission')
+
+    url = '/suppliers/frameworks/<framework_slug>/submissions/<lot_slug>/<service_id>'
+    url = '/suppliers/frameworks/{}/submissions/{}/{}'
```
robyoung: If you want keyword substitution this could be
`url = '/suppliers/frameworks/{framework_slug}/submissions/{lot_slug}/{service_id}'`
and then format's keyword substitution can be used
`url.format(framework_slug='whatever', lot_slug='something', service_id='5432')`

carolinegreen: Didn't mean to leave the first url line in.  Wasn't trying to do a keyword substitution as there don't seem to be any particular standards with these scripts?  But I can change it to that.

quis:
Need to bump dmutils to great than 14.0.0 in `requirements.txt` to bring in this change: https://github.com/alphagov/digitalmarketplace-utils/commit/045c6630096dcde9d64a42a5c8fb16523251b41b#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR16

robyoung:
:+1:

PR #18, opened 2015-12-10
Add script to move G7 communications

robyoung:
Move the G7 communications files to the new bucket while adding the `last_modified` date as a timestamp.

### DIFF
--- a/scripts/oneoff/move-communications.py
+++ b/scripts/oneoff/move-communications.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python
+"""Move communications files from g7 draft documents to communications bucket
+
+Usage:
+    scripts/oneoff/move-communications.py <stage>
+"""
+import sys
+sys.path.insert(0, '.')
+
+import re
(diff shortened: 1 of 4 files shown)


scripts/oneoff/move-communications.py
```diff
@@ -0,0 +1,55 @@
...
+sys.path.insert(0, '.')
+
+import re
+from docopt import docopt
+
+from dmutils.s3 import S3
+
+
+def make_coppier(src_bucket, target_bucket):
+    def copy_file(src_path, target_path):
```
pcraig3: This is super cool.  Didn't know you could do this.

<sub>p.s. copier</sub>

pcraig3:
:+1:

scripts/oneoff/move-communications.py
carolinegreen: Is the spelling intentional?

robyoung: No, I just can't words.

PR #19, opened 2015-12-10
Bump dmutils to 15.9.1

robyoung:
This fixes a bug around formatting the timestamp in S3 save and list.


TheDoubleK:
Looks like it should work once utils is in.
:+1: as long as Travis goes green.

PR #21, opened 2015-12-17
Bulk upload countersigned agreements

TheDoubleK:
This should be a one-time- (or few-times-at-most)-use script to upload the first batch of countersigned G7 framework agreements when we get them from CCS.
See: https://www.pivotaltracker.com/story/show/109718484

Depends on: 
 - [x] https://github.com/alphagov/digitalmarketplace-utils/pull/230

### DIFF
--- a/scripts/bulk-upload-countersigned-agreements.py
+++ b/scripts/bulk-upload-countersigned-agreements.py
@@ -30,10 +30,12 @@ def get_all_pdfs(local_directory):
 def upload_agreement(bucket, dry_run, file_path):
     supplier_id = get_supplier_id_from_framework_file_path(path)
     upload_path = get_countersigned_agreement_document_path("g-cloud-7", supplier_id)
-    print("UPLOADING: '{}' to '{}'".format(file_path, upload_path))
     if not dry_run:
         with open(file_path) as file_contents:
             bucket.save(upload_path, file_contents, acl='public-read', move_prefix=None)
+        print(supplier_id)
+    else:
+        print("[Dry-run] UPLOAD: '{}' to '{}'".format(file_path, upload_path))
 
 
 def get_bucket_name(stage):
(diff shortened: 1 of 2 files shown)


scripts/bulk-upload-countersigned-agreements.py
```diff
@@ -0,0 +1,73 @@
...
+
+def get_bucket_name(stage):
+    bucket_name = 'digitalmarketplace-agreements-{0}-{0}'.format(stage)
+    print("BUCKET: {}".format(bucket_name))
+    return bucket_name
+
+
+def get_supplier_id_from_framework_file_path(path):
+    # We have been told by CCS that filenames will be exactly the same as the unsigned agreement files
+    # e.g. Sanitised_Supplier_Name-123456_Framework_Agreement.pdf
```
robyoung: From the utils it looks like this should be `Sanitised_Supplier_Name-123456-Framework_Agreement.pdf` ie. hyphens either side of the supplier ID.
See: https://github.com/alphagov/digitalmarketplace-utils/blob/master/dmutils/documents.py#L185

TheDoubleK: The filenames sent to us by CCS are in the format shown in the comment here - i.e. with a single dash, underscores elsewhere and with capital letters for `Framework_Agreement`.

TheDoubleK:
Just had a conversation with @cathrooney - looks like the files will probably be coming in from CCS in the format `Sanitised_Supplier_Name-123456_Signed_Framework_Agreement.pdf` - but I think we might as well hold off updating the script until we have a bunch of files to update.

This is rebased up to date and hopefully ready to go, barring putting in the correct filename format when we get the files.

TheDoubleK:
This has been run to upload the first batch of agreements to prod now, so hopefully ready to go :)

PR #23, opened 2015-12-31
Use dmapiclient

allait:
### Use new dmapiclient package instead of dmutils.apiclient
Adds digitalmarketplace-apiclient to the list of requirements and
changes all `dmutils.apiclient` imports to `dmapiclient`.

Since apiclient is using a logger with the package name this also
requires changing the configured loggers in dmscripts/logging and
individual scripts to include `dmapiclient`.

API client `1.0.1` retries 504 responses.

### Fix iter_services not raising an exception when used with Pool.imap
Multiprocessing `Pool.imap` doesn't raise an exception if the iterator
fails to produce any results 👽👽👽 (ie when the first API request fails),
making the process exit with a 0 status. This will mark the CI job
as successful when the indexing failed to run.

Replacing `Pool` with `ThreadPool` seems to fix the problem, and other
scripts are already using `ThreadPool`. Also replacing `.imap` with
`.imap_unordered` since we don't really care about the response order.

### DIFF
--- a/dmscripts/generate_framework_agreement_data.py
+++ b/dmscripts/generate_framework_agreement_data.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 import os
-from dmutils.apiclient.errors import HTTPError
+from dmapiclient import HTTPError
 from dmutils.documents import sanitise_supplier_name
 
 import sys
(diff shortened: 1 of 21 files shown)


robyoung:
:+1:

PR #25, opened 2016-01-19
[#111599650] Evaluate DOS applications and mark as PASS/FAIL

TheDoubleK:
Well this has felt like a marathon to get together a reasonably short script... oh well, it's here now.
See: https://www.pivotaltracker.com/story/show/111599650

The script will check all interested DOS suppliers and mark as a FAIL (`on_framework=FALSE`):
 - those that didn't submit an application
 - those that got any mandatory declaration questions wrong
 - those who didn't submit any valid services

It will mark as a PASS (`on_framework=TRUE`):
 - those who completed the declaration properly and submitted a valid service in at least one lot

It will regard as DISCRETIONARY and not mark anything in the DB at all for:
 - those who answered a discretionary declaration question wrong and submitted a valid service in at least one lot

Where a "valid service" is one that has all the right answers to the "service essential" questions.


dmscripts/insert_dos_framework_results.py
```diff
@@ -0,0 +1,117 @@
...
+
+    for draft in submitted_drafts:
+        draft_result = check_service_essentials(draft)
+        if draft_result == PASS:
+            supplier_has_submitted_services = True
+        else:
+            # Update the draft to be 'failed'
+            client.update_draft_service(draft['id'], {"status": "failed"}, user)
+
+    return supplier_has_submitted_services
```
allait: I don't think we've decided what to do with suppliers who are passing except for one of the lots. It might be worth talking with Cath whether we should be marking them "discretionary" instead.

scripts/insert-dos-framework-results.py
```diff
@@ -0,0 +1,38 @@
...
+from dmapiclient import DataAPIClient
+from dmscripts.insert_dos_framework_results import process_dos_results
+from dmutils.content_loader import ContentLoader
+
+if __name__ == '__main__':
+    arguments = docopt(__doc__)
+
+    path_to_frameworks = arguments['<path_to_frameworks>']
+    content_loader = ContentLoader(path_to_frameworks)
+    client = DataAPIClient(arguments['<data_api_url>'], arguments['<data_api_token>'])
```
allait: We have a small helper to avoid typing full URLs https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/index-services.py#L120

TheDoubleK: Nice - now updated to use this.

dmscripts/insert_dos_framework_results.py
TheDoubleK: Discussed with Cath in real life - the behaviour here is correct and suppliers with no valid lots should be Failed.  The second script which generates the numbers for CCS then needs to show counts for `Draft`, `Complete-OK` and `Complete-Failed` services, which it can do by checking for the new `failed` status on drafts.

robyoung:
:+1:

PR #26, opened 2016-01-20
Bugfix to handle case where declaration has not been started

TheDoubleK:
There was a bug in the script where it dies if the declaration had not been started so came back as `None`; this fixes it.


robyoung:
:+1:

PR #31, opened 2016-02-03
Add a script to remove supplier name from agreement file paths

allait:
Finds agreement files that have the supplier name in the path and
renames them, so that links don't break if suppliers change their
name in the database.

Old file name is set as Content-Disposition header filename, so that
when file is downloaded it's saved with has supplier name in the
file name.


pcraig3:
:+1:

PR #32, opened 2016-02-04
Refactor agreements upload scripts to be more generic

carolinegreen:
Refactor agreements upload scripts to be more generic to allow different types of documents for different frameworks to be uploaded.  Also added tests for this.

Relies on:
https://github.com/alphagov/digitalmarketplace-utils/pull/239


scripts/bulk-upload-documents.py
```diff
@@ -0,0 +1,40 @@
+"""
+This will scan a directory for files with a filename matching -<supplier_id>_Framework_Agreement
```
TheDoubleK: The description and example usage needs to be updated to explain the new <document_type> parameter

scripts/bulk-upload-documents.py
```diff
@@ -0,0 +1,40 @@
...
+and upload them to the S3 agreements bucket for <stage> with the file path
+<framework_slug>/agreements/<supplier_id>/<supplier_id>-countersigned-framework-agreement.pdf
+
+Usage:
+    scripts/bulk-upload-countersigned-agreements.py <local_directory> <stage> <framework_slug>
+    <document_type> [--dry-run]
```
TheDoubleK: It would be good to list the allowed possible values for `document_type` here - i.e. the types of document we have at the moment.

dmscripts/bulk_upload_documents.py
```diff
@@ -0,0 +1,39 @@
...
+from dmutils.documents import get_countersigned_agreement_document_path
+
+
+def upload_file(bucket, dry_run, file_path, framework_slug):
+    supplier_id = get_supplier_id_from_framework_file_path(file_path)
+    upload_path = get_countersigned_agreement_document_path(framework_slug, supplier_id)
+    if not dry_run:
+        with open(file_path) as file_contents:
+            bucket.save(upload_path, file_contents, acl='public-read', move_prefix=None)
+        print(supplier_id)
```
robyoung: We should probably be using the [logging infrastructure](https://github.com/alphagov/digitalmarketplace-scripts/blob/e03a8c15276440e44a5c2104034f35523827e177/scripts/index-services.py#L28) for these print statements now that we have it.

tests/test_bulk_upload_countersigned_agreements.py
```diff
@@ -0,0 +1,69 @@
+import os
```
TheDoubleK: Should this file be `test_bulk_upload_documents` now?

dmscripts/bulk_upload_documents.py
```diff
@@ -0,0 +1,39 @@
+import os
+import re
+import sys
+sys.path.insert(0, '.')
+from dmutils.documents import get_countersigned_agreement_document_path
```
TheDoubleK: It looks like a recent change to utils has removed the `get_countersigned_agreement_document_path` method altogether :/ - does it need to be put back?  Or is there a new method that can be used here that will do the job?

TheDoubleK: (And as this script is no longer only for countersigned agreements we should probably have a more generic sounding method to use anyway)

dmscripts/bulk_upload_documents.py
```diff
@@ -0,0 +1,39 @@
...
+
+def get_all_files_of_type(local_directory, file_type):
+    for root, subfolder, files in os.walk(local_directory):
+        for filename in files:
+            if filename.endswith(file_type):
+                yield os.path.join(root, filename)
+
+
+def get_supplier_id_from_framework_file_path(path):
+    # Filenames from CCS are in the format:
```
TheDoubleK: Now this is more generic this comment is misleading.  With the new regex the only thing we can say about the file path is something like
```
# All files to be uploaded must have a dash followed by a supplier id
# The rest of the filename is disregarded 
```
This worries me slightly... it means there is a need to be extra-careful that only the types of file we want to upload are in the folder. 

I don't know if it would be possible to have another parameter that defines a bit more narrowly which files should be uploaded... it's probably too tricky to do in a robust way, but we'll just need to be careful about the folders we run this against.

scripts/bulk-upload-documents.py
carolinegreen: Couldn't that change though?

tests/test_bulk_upload_countersigned_agreements.py
```diff
@@ -0,0 +1,69 @@
...
+            upload_file(bucket, False, file_path, 'g-cloud-7')
+            assert bucket.save.call_count == 1
+
+
+@mock.patch('dmutils.s3.S3')
+def test_upload_file_dry_run_doesnt_upload(bucket):
+    file_path = 'The_Business_Software_Centre-92877-signed-framework-agreement.pdf'
+    with mock.patch.object(builtins, 'open', mock.mock_open(read_data='foo')):
+        with open(file_path):
+            upload_file(bucket, False, file_path, 'g-cloud-7')
```
TheDoubleK: The `False` here should be `True`, and then `bucket.save.call_count == 0`

dmscripts/bulk_upload_documents.py
```diff
@@ -0,0 +1,39 @@
...
+sys.path.insert(0, '.')
+from dmutils.documents import get_countersigned_agreement_document_path
+
+
+def upload_file(bucket, dry_run, file_path, framework_slug):
+    supplier_id = get_supplier_id_from_framework_file_path(file_path)
+    upload_path = get_countersigned_agreement_document_path(framework_slug, supplier_id)
+    if not dry_run:
+        with open(file_path) as file_contents:
+            bucket.save(upload_path, file_contents, acl='public-read', move_prefix=None)
```
allait: We shouldn't be setting `public-read` for agreement files, since they shouldn't have a public link, we're always generating signed URLs for them.

dmscripts/bulk_upload_documents.py
carolinegreen: Yeah managed to overlook that.

dmscripts/bulk_upload_documents.py
carolinegreen: Didn't know about that, will add it.

scripts/bulk-upload-documents.py
TheDoubleK: Yes I guess so. We only have agreements that we'd ever want to bulk-upload at the moment I think? In which case just an example usage with `agreements` should be fine.

dmscripts/bulk_upload_documents.py
carolinegreen: Oh right, that was taken from the original script.

dmscripts/bulk_upload_documents.py
carolinegreen: That was too specific for a regex.  Thought the formats they were using might be useful to know in the future if things changed.  Can take the comment out though.  I think the regex is self explanatory.

tests/test_bulk_upload_countersigned_agreements.py
carolinegreen: Yes

dmscripts/bulk_upload_documents.py
allait: No idea, but it might be worth checking the uploaded files to make sure they're not public if the old script was used before

TheDoubleK: They are.  I'll fix it.

TheDoubleK: But yes, should be `private` in the script here.

TheDoubleK: ACL of existing all fixed now.

dmscripts/bulk_upload_documents.py
```diff
@@ -0,0 +1,39 @@
...
+
+def upload_file(bucket, dry_run, file_path, framework_slug, bucket_category,
+                document_category, document_type):
+    supplier_id = get_supplier_id_from_framework_file_path(file_path)
+    document_name = '{0}.{1}'.format(document_category, document_type)
+    upload_path = get_document_path(framework_slug, supplier_id, bucket_category,
+                                    document_name)
+    if not dry_run:
+        with open(file_path) as file_contents:
+            bucket.save(upload_path, file_contents, acl='private', move_prefix=None)
```
allait: We should probably set `timestamp` here (like in https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/oneoff/set-agreements-content-disposition.py#L32, except using the current time), so that if files are modified we can keep the original upload date.

tests/test_bulk_upload_documents.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def test_get_supplier_id_from_framework_file_path_for_file_with_underscores():
+    path = '../../Downloads/Completed Frameworks BATCH2 SHARED WITH GDS/ICNT_Consulting_Ltd-584425_signed_framework_agreement.pdf'  # noqa
+    assert get_supplier_id_from_framework_file_path(path) == '584425'
+
+
+@mock.patch('dmutils.s3.S3')
+def test_upload_file(bucket):
+    file_path = 'The_Business_Software_Centre-92877-signed-framework-agreement.pdf'
```
allait: Uploaded file names should no longer include supplier names, so I think it's worth updating the tests to reflect the correct format: e.g. `<supplier_id>-signed-framework-agreement.pdf`

dmscripts/bulk_upload_documents.py
```diff
@@ -0,0 +1,39 @@
...
+
+def get_all_files_of_type(local_directory, file_type):
+    for root, subfolder, files in os.walk(local_directory):
+        for filename in files:
+            if filename.endswith(file_type):
+                yield os.path.join(root, filename)
+
+
+def get_supplier_id_from_framework_file_path(path):
+    match = re.search(r'-(\d{5,6})', path)
```
allait: With the new filename format, uploaded filenames should start with supplier_id, so if we're not modifying filenames generated by the pdf creation script then we should be looking for `(\d{5,6})-` instead.

dmscripts/bulk_upload_documents.py
```diff
@@ -0,0 +1,72 @@
...
+def generate_download_filename(supplier_id, document_name, supplier_name):
+    return '{}-{}-{}'.format(sanitise_supplier_name(supplier_name), supplier_id, document_name)
+
+
+def get_supplier_name_dict_from_tsv(tsv_path):
+    suppliers_name_dict = {}
+    with open(tsv_path, 'r') as csvfile:
+        tsv_path = csv.reader(csvfile, delimiter='\t')
+        for row in tsv_path:
+            print('>>>>row[0]', row[0])
```
allait: Looks like we forgot about this debug print

allait:
Other than the tsv debug print, :+1:

PR #33, opened 2016-02-12
Add script for framework agreements source file

tombye:
We're using Adobe Acrobat to generate result letters and framework agreements. This requires PDFs for the base pages and a data file, in TSV format, stored as a `.txt` file.

This script generates the data file from the CSV of successful suppliers produced by [this script](https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/export-dos-suppliers.py).

Part of this story: https://www.pivotaltracker.com/story/show/112422275


robyoung:
The name of the first commit is back to front. It removes an erroneous `]`.

dmscripts/generate_dos_import_data.py
```diff
@@ -0,0 +1,122 @@
...
+
+def create_import_data_file_for_documents(import_data, target_dir, document_type):
+    file_path = os.path.join(target_dir, '{}_import_data.txt'.format(document_type))
+    csv_writer = csv.writer(open(file_path, 'w+'), delimiter="\t", lineterminator="\n")
+    for row in import_data:
+        csv_writer.writerow(row)
+
+
+def generate_import_data(suppliers_file_path, import_data_dir_path, framework_slug, document_type):
+    suppliers_csv_data = get_list_from_csv_file(open(suppliers_file_path))
```
robyoung: It's good practice to use `open` in a `with` statement to avoid dangling resources.
```python
with open(filename) as opened_file:
    do_thing_with_file(opened_file)
```

dmscripts/generate_dos_import_data.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def get_indices_of_required_columns(supplier_info_headings, required_columns):
+    return [supplier_info_headings.index(heading) for heading in required_columns]
+
+
+def reformat_csv_data(csv_data, framework_slug, document_type):
+    config = IMPORT_DATA_CONFIG[framework_slug]
+    number_of_lots = len(config['lot_names'])
+    lot_results_start_idx = len(csv_data[1]) - (3 * number_of_lots)
```
robyoung: Why are we taking the length from row 1 rather than row 0? Is it possible that the column count for headers is different to that of the rows?

dmscripts/generate_dos_import_data.py
```diff
@@ -0,0 +1,122 @@
...
+    supplier_info_headings = csv_data[0][:lot_results_start_idx]
+    required_column_indicies = get_indices_of_required_columns(
```
robyoung: The slightly confusing slicing and indexing could be avoided with a [`csv.DictReader`](https://docs.python.org/3/library/csv.html#csv.DictReader).

tombye:
I've fixed the obvious issues raised and refactored to use `DictReader/Writer` (great advice there @robyoung). I also renamed the scripts to explain better that the files they produce are for importing into PDFs.

PR #35, opened 2016-02-15
Add a script to update DOS supplier onFramework flags

allait:
Reads supplier_ids from a file (one ID per line) and sets supplier
onFramework status to pass or fail, based on the flag used.


scripts/update-dos-framework-results.py
```diff
@@ -0,0 +1,43 @@
...
+
+logger = configure_logger()
+
+if __name__ == '__main__':
+    arguments = docopt(__doc__)
+
+    data_api_url = get_api_endpoint_from_stage(arguments['<stage>'], 'api')
+    client = DataAPIClient(data_api_url, arguments['--api-token'])
+    user = getpass.getuser()
+    result = True if arguments['--pass'] else False
```
TheDoubleK: Should we have a condition on the `else` here, to f guard against typos? E.g. `--paas` by mistake would fail everyone, but might be better to stop and warn `must be pass or fail`.

TheDoubleK:
I was going to suggest adding a `framework_slug` parameter to make it more general purpose, but it's only good for `pending` frameworks so not really worth it at this stage.

scripts/update-dos-framework-results.py
allait: I wasn't sure about this check either, but `docopt` won't accept the command line arguments unless one (and only one) of the `--pass` or `--fail` was set, so I thought there was no need to repeat those checks.

TheDoubleK: Ah sorry, yes I hadn't seen that in the usage definition.

TheDoubleK:
:+1:  after a rebase

PR #42, opened 2016-03-04
Rename dos locations

TheDoubleK:
Migrates outcomes and recruitment locations to plain old `locations`.

### DIFF
--- a/scripts/oneoff/rename-dos-locations.py
+++ b/scripts/oneoff/rename-dos-locations.py
@@ -8,6 +8,8 @@ Usage:
 import sys
 sys.path.insert(0, '.')
 
+from multiprocessing.pool import ThreadPool
+import itertools
 from docopt import docopt
 from dmapiclient import DataAPIClient
 from dmscripts.env import get_api_endpoint_from_stage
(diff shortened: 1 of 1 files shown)


scripts/oneoff/rename-dos-locations.py
```diff
@@ -0,0 +1,67 @@
...
+            update["recruitLocations"] = None
+
+        if update:
+            counter += 1
+
+            update_reason = "rename locations: https://www.pivotaltracker.com/story/show/113271913"
+            if dry_run:
+                print("{} {}".format(draft["id"], update))
+            else:
+                client.update_draft_service(draft["id"], update, update_reason)
```
allait: None of the apiclient methods accept `update_reason` anymore. While we could send anything instead of the user name the value ends up being stored in the audit event `.user` field, so I don't think we should be doing this.

robyoung: I think a dummy user name here is less useful than something that actually identifies why the change was made.

TheDoubleK: Yes, this was deliberate, so that the audit events would make clear why the change was made - who ran the script to make the update is irrelevant.

allait: I'm still not sure about this. I know the user field is not strictly machine readable since we put both user emails and `whoami` there, but it's still `user`, so I think you could expect to see a name or a reference there. Reason text would look like the fields got mixed up. Putting a script name instead of reason might be a good middle ground.

I've just remembered another problem with draft updates though – [we actually look up the last event user in the API](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/drafts.py#L171). This is then [used by the supplier frontend to show "Last edited by"](https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/app/templates/services/service_submission.html#L76). So if we do this without setting the user to an existing user email, it will show `Last edited <date> by `. I'm not sure what's the best approach here. We could use the admin account email address to make the message look right, but then we're not actually making any visible changes to the service data.

allait:
Was run in production :+1:

PR #46, opened 2016-04-12
Tweaks to make-dos-live script

TheDoubleK:
This has been run on production now - makes sense to save these tweaks to the script for future reference.

Print statements need to be unicode because some supplier names have unicode characters in them.
Catching and reporting errors is better than killing the whole run for a single HTTP error.

### DIFF
--- a/scripts/make-dos-live.py
+++ b/scripts/make-dos-live.py
@@ -6,10 +6,8 @@ Usage:
 import sys
 sys.path.insert(0, '.')
 
-import re
-
 from docopt import docopt
-from dmscripts.env import get_api_endpoint_from_stage, get_assets_endpoint_from_stage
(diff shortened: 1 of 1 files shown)


allait:
👍

Should we move this to scripts/oneoff though?

TheDoubleK:
Maybe - I was going to say the same thing to Rob when he first put the script up, but then noticed that our previous go-live scripts are also just in `/scripts`.

I thought maybe leave it here because it's OK to run it again, e.g. if some failed suppliers got on to the framework this script could be used to put their services live without breaking anything.

I'm easy either way really.

TheDoubleK:
Just having a look, there are quite a lot of scripts in `scripts` that are arguably one-offs.  I'm going to leave this here for now and we should have a clear out of stuff at some point in the future.

PR #47, opened 2016-04-27
Minor tweaks to question csv script

TheDoubleK:
The unicode changes here might not actually be required (I think it was LibreOffice choking rather than a problem with the output that prompted me to make them), but brings the script into line with how we do things in most other scripts with unicode/CSVs.


allait:
👍

PR #48, opened 2016-05-05
Add script to export buyer users and their briefs

TheDoubleK:
For this chore: https://www.pivotaltracker.com/story/show/118936027

Script has been run and I've sent the results to Sarah. Adding it here for future use, and for your suggestions for improvements, of course :)


allait:
👍

PR #49, opened 2016-05-12
Include sign-up date in buyer user list export

TheDoubleK:
This is useful for getting in touch with only users who are new since the last time you ~~spammed~~ wrote to everyone.


allait:
👍

PR #50, opened 2016-05-19
Allow agreements to be cleared individually

tombye:
The existing `clear-signed-agreements.py` script
will delete all the agreements for each supplier
in the given framework.

This adds the ability to specify certain suppliers
in that framework to target.

Related to the story to upload 2 new framework agreements: https://www.pivotaltracker.com/story/show/119371899

### DIFF
--- a/scripts/clear-signed-agreements.py
+++ b/scripts/clear-signed-agreements.py
@@ -38,9 +38,6 @@ logger = logging.configure_logger()
 
 
 def main(stage, framework_slug, api_token, user, supplier_ids=None):
-    if supplier_ids:
-        supplier_ids = supplier_ids.split(',')
-
     agreements_bucket_name = 'digitalmarketplace-agreements-{0}-{0}'.format(stage)
     agreements_bucket = S3(agreements_bucket_name)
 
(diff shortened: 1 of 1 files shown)


scripts/clear-signed-agreements.py
```diff
@@ -46,22 +48,41 @@ def main(stage, framework_slug, api_token, user):
...
+    return filter(
+        matcher,
+        agreements_bucket.list('{}/agreements/'.format(framework_slug))
+    )
+
+
+def get_supplier_ids_from_csv(csv_path):
+    with open(csv_path, 'r') as csvfile:
+        for row in csv.reader(csvfile):
+            # csv should only have one row
```
TheDoubleK: Should it check that there aren't more?  Maybe fail with a message if there's more than one row?  Deleting signed agreements feels pretty hardcore, so people need to be sure what they're doing.  If they have the wrong format file then they probably are deleting stuff they shouldn't be.

scripts/clear-signed-agreements.py
```diff
@@ -46,22 +48,41 @@ def main(stage, framework_slug, api_token, user):
...
+
 if __name__ == '__main__':
     arguments = docopt(__doc__)
     stage = arguments['<stage>']
     framework_slug = arguments['<framework_slug>']
     data_api_access_token = arguments['--api-token']
+    supplier_ids = None
 
     user = getpass.getuser()
+    if arguments['--csv_path']:
```
TheDoubleK: If it's just a single row of supplier IDs they could just be passed in on the command line rather than in a file?  Not sure... maybe that's more prone to user error?

TheDoubleK:
Personally I'd just have gone into S3 and deleted the files for two suppliers... but this looks like it will work.

tombye:
It does feel a bit like over-egging the custard but I do need to set the flag for those suppliers too and I was going to go through the commands manually so though I should just extend this script.

tombye:
This is missing work to only mark the flag for supplier on the list if its sent in. Closing until this is added.

scripts/clear-signed-agreements.py
```diff
@@ -41,13 +47,22 @@ def main(stage, framework_slug, api_token, user):
...
+    for supplier_id in supplier_ids:
         logger.info("Resetting agreement returned flag for supplier {supplier_id}",
-                    extra={'supplier_id': supplier['supplierId']})
-        api_client.unset_framework_agreement_returned(supplier['supplierId'], framework_slug, user)
+                    extra={'supplier_id': supplier_id})
+        api_client.unset_framework_agreement_returned(supplier_id, framework_slug, user)
 
     signed_agreements = filter(
-        lambda x: re.search(r'/(\d+)-signed-framework-agreement.pdf', x['path']),
+        lambda x: x['path'] in ['{}-signed-framework-agreement.pdf'.format(id) for id in supplier_ids],
```
allait: We should remove `.pdf`, since uploaded agreements could be scanned images of last page. I had to fix it manually last time, but forgot to commit.

scripts/clear-signed-agreements.py
```diff
@@ -41,13 +47,22 @@ def main(stage, framework_slug, api_token, user):
     )
 
     suppliers = api_client.find_framework_suppliers(framework_slug, agreement_returned=True)['supplierFrameworks']
-    for supplier in suppliers:
+
+    if supplier_ids is not None:
+        supplier_ids = [
+            str(supplier['supplierId']) for supplier in suppliers
```
allait: Supplier IDs should be integers, so I think we could convert to `int` here

tombye: I think they are only used in two places later in this script:

1. sent as an argument to `find_framework_suppliers`
2. in a comparison with the result of a regular expression match when filtering the contents of the bucket

I was assuming both of these can be strings and wanted `supplier_ids` to contain one type, whether they were sent in as options or not. Is there another reason to keep them as `int`?

allait: I think the main reason would be to check that they're actually numbers early, before sending them to the API

tombye:
I've rewritten it after chatting to @allait about the [previous comment](https://github.com/alphagov/digitalmarketplace-scripts/pull/50#discussion_r64191005) on the type conversion happening to the supplier_ids.

The script now includes validation of any ids sent in as options and keeps all ids in the `supplier_ids` list as `int`s.

scripts/clear-signed-agreements.py
```diff
@@ -41,13 +47,19 @@ def main(stage, framework_slug, api_token, user):
     )
 
     suppliers = api_client.find_framework_suppliers(framework_slug, agreement_returned=True)['supplierFrameworks']
-    for supplier in suppliers:
+
+    if supplier_ids is not None:
+        supplier_ids = validate_entered_ids(supplier_ids, suppliers)
```
allait:
```python

supplier_ids = [int(supplier_id.strip()) for supplier_id in supplier_ids.split(',')]

suppliers = api_client ....

missing_supplier_ids = set(supplier_ids) - set(supplier['supplierId'] for supplier in suppliers)
if missing_supplier_ids:
    raise Exception("Invalid supplier IDs: {}".format(missing_supplier_ids))
```

allait:
👍

PR #52, opened 2016-06-23
[#121428145] add generate-g-cloud-8-master-csv.py oneoff script

risicle:
Continued from https://github.com/alphagov/digitalmarketplace-api/pull/410


TheDoubleK:
Yep I think this is good to go - the proof will be in the pudding when we run the data extract tomorrow morning :)

:+1:

PR #53, opened 2016-07-11
Supplier "about you" data extract

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/121428561

See the story for details of the data extracted by this new script.  Output has been OK'd by Andrew on Preview and Staging, and Prod data passed to Cath.

The script takes over 2 minutes to run against production data, which means that if it was built in to the admin app (as in the story) then the request would time out.  Agreed with Cath and @allait that we won't embed in admin app right now, but instead create a Jenkins job to wrap this script so new versions of the data can be easily generated when required by CCS.


dmscripts/export_framework_applicant_details.py
```diff
@@ -0,0 +1,137 @@
...
+
+    return inner
+
+
+def add_submitted_draft_counts(client, framework_slug):
+    def inner(record):
+        counts = {lot: 0 for lot in LOTS[framework_slug]}
+
+        for draft in client.find_draft_services_iter(record['supplier']['id'], framework=framework_slug):
+            if draft['status'] == 'submitted':
```
allait: It's interesting that we don't have a `status` filter for `/drafts`. We should think about adding this at some point.

allait:
👍

PR #54, opened 2016-07-14
Generalise script for all frameworks, not just DOS

TheDoubleK:
~~The previous version of this script only updated results for DOS.  This can now be used for any framework.~~

Useful for completing these chores: 
https://www.pivotaltracker.com/story/show/126020263
https://www.pivotaltracker.com/story/show/126119097

This has grown a bit since initial PR- I've extracted some common methods into a  `framework_utils` file and adjusted existing scripts accordingly, and also tweaked the `make-g-cloud-live` script to work for the new document paths for G-Cloud 8 (this is legit because it should be the same for new frameworks going forward).

### DIFF
--- a/dmscripts/framework_utils.py
+++ b/dmscripts/framework_utils.py
@@ -1,9 +1,24 @@
 from dmapiclient import HTTPError
 
 
+def get_submitted_drafts(client, framework_slug, supplier_id):
+    services = client.find_draft_services(supplier_id, framework=framework_slug)
+    services = services["services"]
+    submitted_services = [service for service in services if service["status"] == "submitted"]
+    return submitted_services
+
+
 def set_framework_result(client, framework_slug, supplier_id, result, user):
     try:
         client.set_framework_result(supplier_id, framework_slug, result, user)
         return "  Result set OK: {} - {}".format(supplier_id, "PASS" if result else "FAIL")
     except HTTPError as e:
         return "  Error inserting result for {} ({}): {}".format(supplier_id, result, str(e))
+
+
+def has_supplier_submitted_services(client, framework_slug, supplier_id):
+    submitted_drafts = get_submitted_drafts(client, framework_slug, supplier_id)
+    if len(submitted_drafts) > 0:
+        return True
+    else:
+        return False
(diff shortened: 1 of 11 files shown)


scripts/update-framework-results.py
```diff
@@ -23,21 +23,23 @@
 
 from dmscripts.env import get_api_endpoint_from_stage
 from dmscripts.logging import configure_logger
-from dmscripts.insert_dos_framework_results import insert_result
```
allait: Is this file used by anything else or can we remove it?

dmscripts/framework_utils.py
```diff
@@ -0,0 +1,9 @@
+from dmapiclient import HTTPError
+
+
+def set_framework_result(client, framework_slug, supplier_id, result, user):
+    try:
+        client.set_framework_result(supplier_id, framework_slug, result, user)
+        return"  Result set OK: {} - {}".format(supplier_id, "PASS" if result else "FAIL")
```
allait: I'm a bit surprised the parser accepts this but I think the convention is to have a space after `return`

TheDoubleK: :eyes:
Yep I'd have expected PEP8 to choke on that.

scripts/update-framework-results.py
TheDoubleK: Good question - Ill check it.

TheDoubleK: A bunch of constants from that file are used in another script, but the code is unused.

It does, however, capture the more complex logic we had for DOS assessment - with the lot-specific "declaration"  questions that led to the logic of individual services/lots being able to fail, which we don't have for G-Cloud at all.

I don't know if this is something that we want to keep around for posterity in case we have that kind of thing again in future (it is a bit fiddly), or if we just want to clear it out in an effort to streamline our scripts a bit.

allait: I'd clear it out if we don't have a script using it anymore. There's always git history.

TheDoubleK: Turns out there is a oneoff script using it so I haven't killed it, but have refactored things a bit more instead.

idavidmcdonald:
Looks good 👍

PR #56, opened 2016-07-22
Only check free-text fields for banned words

TheDoubleK:
Previously this script checked all fields for bad words, including those with enum values.

One of our "service types" multiple choices is "Software development tools", and one of the current "bad words" is "software development", which means that previously all services which legitimately chose that option for a service type would be flagged as suspicious.


idavidmcdonald:
👍

PR #57, opened 2016-08-18
Update clear-signed-agreements script

allait:
This is the version we've run to remove G8 signed agreements after
fixing lot numbering. It adds <cutoff_time> support to only remove
agreements uploaded before given time.

### DIFF
--- a/scripts/clear-signed-agreements.py
+++ b/scripts/clear-signed-agreements.py
@@ -94,7 +94,7 @@ if __name__ == '__main__':
 
     if arguments.get('<supplier_id_file>'):
         with open(arguments['<supplier_id_file>'], 'r') as f:
-            supplier_ids = filter(None, [int(l.strip()) for l in f.readlines()])
+            supplier_ids = list(filter(None, [int(l.strip()) for l in f.readlines()]))
     else:
         supplier_ids = None
 
--- a/scripts/clear-signed-agreements.py
+++ b/scripts/clear-signed-agreements.py
@@ -12,12 +12,15 @@ This should be run before the new agreements are uploaded: this way there's no r
 new agreements signed by the suppliers.
 
 Usage:
-    scripts/oneoff/clear-signed-agreements.py <stage> <framework_slug> --api-token=<api_access_token>
-        [--supplier_ids=<supplier_ids>]
+    scripts/clear-signed-agreements.py <stage> <framework_slug> --api-token=<api_access_token>
+    [<supplier_id_file>] [<cutoff_time>]
+
+Arguments:
+    <cutoff_time>  if set, the script will only remove agreements uploaded before the given timestamp.
+                   This should be a valid ISO 8601 string (eg 2016-08-22T00:00:00.00000Z)
 
 Options:
     --api-token=<api_access_token>  API token
-    --supplier_ids=<supplier_ids>   Suppliers to target (comma-separated)
 
 """
 import sys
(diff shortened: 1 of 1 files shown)


TheDoubleK:
:+1:

pcraig3:
Thanks, Kev!

scripts/clear-signed-agreements.py
```diff
@@ -12,20 +12,18 @@
 new agreements signed by the suppliers.
 
 Usage:
-    scripts/oneoff/clear-signed-agreements.py <stage> <framework_slug> --api-token=<api_access_token>
-        [--supplier_ids=<supplier_ids>]
+    scripts/clear-signed-agreements.py <stage> <framework_slug> --api-token=<api_access_token>
+    [<supplier_id_file>] [<cutoff_time>]
```
pcraig3: I think it would be cool to have an example of a cutoff time that would work.

scripts/clear-signed-agreements.py
```diff
@@ -52,22 +50,29 @@ def main(stage, framework_slug, api_token, user, supplier_ids=None):
     suppliers = api_client.find_framework_suppliers(framework_slug, agreement_returned=True)['supplierFrameworks']
```
pcraig3: This line is fine, but line 48 keeps breaking for me.

```
# 48
supplier_ids = [int(supplier_id.strip()) for supplier_id in supplier_ids.split(',')]
```

It looks like it's looking for a comma-separated string of supplier ids, which isn't the format ids are stored as in the supplier_id file. 

On line 95, the supplier_id file is looking for a file with supplier ids separated by newlines.  This is also the format we use for bulk document uploading, so I think stick with that.

scripts/clear-signed-agreements.py
```diff
@@ -52,22 +50,29 @@ def main(stage, framework_slug, api_token, user, supplier_ids=None):
...
     signed_agreements = filter(
-        lambda x: match_signed_agreements(supplier_ids, x['path']),
+        lambda x: re.search(r'/(\d+)-signed-framework-agreement.', x['path']),
         agreements_bucket.list('{}/agreements/'.format(framework_slug))
     )
 
+    if supplier_ids is not None:
+        signed_agreements = [
+            agreement for agreement in signed_agreements
+            if int(re.search(r'/(\d+)-signed-framework-agreement', agreement['path']).group(1)) in supplier_ids
```
pcraig3: This doesn't work in Python3, but neither do most of our scripts 
`¯\_(ツ)_/¯`

scripts/clear-signed-agreements.py
```diff
@@ -87,4 +92,10 @@ def match_signed_agreements(supplier_ids, path):
```
pcraig3: Line 91 is still looking for the `supplier_ids` parameter we've gotten rid of (which, presumably was a comma-separated list).

```
# 91 
supplier_ids = arguments['--supplier_ids']
```

I think we just want it gone.

scripts/clear-signed-agreements.py
allait: Do you remember why this didn't work? `int(re.search(r'/(\d+)-signed-framework-agreement', '/2131-signed-framework-agreement').group(1))` seems to run fine on 3.5.2

allait:
@pcraig3 updated, should fix everything apart from the Python3 issue

scripts/clear-signed-agreements.py
pcraig3: No problems anymore! Not sure what I was getting originally, but it was probably a side effect of something else.

pcraig3:
I think this is a 👍 , but that's what I would say since I've done the last commit.

PR #58, opened 2016-08-25
Notify buyers when briefs close

allait:
* [x] Requires https://github.com/alphagov/digitalmarketplace-utils/pull/279

### Move render_html to dmscripts.email
Rendering html is required for emails, so moving the function from a script-specific dmscripts module.

### Update dmutils and dmapiclient
Includes briefs `with_users` request and `send_email` custom logger. Content loader has been split up since the last time utils were updated, so this adds dmcontent as a new dependency and updates all import paths.

Using 1.0.0 version of content-loader since the latest one breaks tests.

### Add a function to search sent Mandrill messages
Requests tagged messages from Mandrill history. Mandrill stores sent messages for 7 days and filtered by tag we don't expect to send more than 1000 messages so we can avoid adding pagination for now.

Note that messages don't appear in Mandrill search API response until they've been delivered, so rerunning the send script right away will probably end up sending duplicate emails.

### Add a script to notify buyers when requirement closes

screen shot 2016-08-25 at 17 19 28
screen shot 2016-08-25 at 17 18 59

Gets the list of briefs that closed on a given date (yesterday by default) from the API, checks brief IDs against Mandrill history of sent 'requirements-closed' emails and sends notification to brief users for briefs that don't have an email history in Mandrill.

Since briefs aren't ordered by close date we need to download all closed briefs from the API. There are only few pages at the moment, so this shouldn't be a problem for now.

Sent emails don't appear in mandrill history until the message has been delivered, so rerunning the script right away could result in duplicate emails.

Doesn't allow to send notifications for briefs that closed more than a week ago. We probably don't need to notify users about something that happened a long time ago and this helps to avoid typos.

Yesterday is determined from UTC time, so running the script at midnight UTC+1 could be a bit confusing.

### DIFF
--- a/dmscripts/email.py
+++ b/dmscripts/email.py
@@ -1,14 +1,4 @@
-import io
-from jinja2 import Template
-
 from mandrill import Mandrill
-
-
-def render_html(template_path, data=None):
-    with io.open(template_path, encoding='UTF-8') as htmlfile:
-        html = htmlfile.read()
-        template = Template(html)
-        return template.render(data or {})
 
 
 def get_sent_emails(mandrill_api_key, tags, date_from=None):
--- a/dmscripts/email.py
+++ b/dmscripts/email.py
@@ -1,0 +1,11 @@
+import io
+from jinja2 import Template
+
+from mandrill import Mandrill
+
+
+def render_html(template_path, data=None):
+    with io.open(template_path, encoding='UTF-8') as htmlfile:
+        html = htmlfile.read()
+        template = Template(html)
+        return template.render(data or {})
--- a/dmscripts/email.py
+++ b/dmscripts/email.py
@@ -9,3 +9,9 @@ def render_html(template_path, data=None):
         html = htmlfile.read()
         template = Template(html)
         return template.render(data or {})
+
+
+def get_sent_emails(mandrill_api_key, tags, date_from=None):
+    mandrill_client = Mandrill(mandrill_api_key)
+
+    return mandrill_client.messages.search(tags=tags, date_from=date_from, limit=1000)
--- a/dmscripts/notify_buyers_when_requirements_close.py
+++ b/dmscripts/notify_buyers_when_requirements_close.py
@@ -1,0 +1,103 @@
+# -*- coding: utf-8 -*-
+
+from datetime import datetime, timedelta
+
+import dmapiclient
+from dmutils.formats import DATE_FORMAT, DATETIME_FORMAT
+from dmutils.email import send_email, MandrillException
+
+from .email import render_html, get_sent_emails
+
+from . import logging
+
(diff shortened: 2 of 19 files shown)


dmscripts/email.py
```diff
@@ -0,0 +1,17 @@
+import io
+from jinja2 import Template
+
+from mandrill import Mandrill
+
+
+def render_html(template_path, data=None):
```
TheDoubleK: Does `render_html` belong in `email`?  It seems more general purpose, and e.g. looks odd to import something from `email` in the `generate_agreement_signature_pages` script, which doesn't do anything with emails at all.

dmscripts/notify_buyers_when_requirements_close.py
```diff
@@ -0,0 +1,103 @@
...
+
+from . import logging
+
+logger = logging.configure_logger({'dmapiclient': logging.INFO})
+
+
+def get_closed_briefs(data_api_client, closed_at):
+    return [
+        brief for brief in data_api_client.find_briefs_iter(status='closed', with_users=True)
+        if datetime.strptime(brief['applicationsClosedAt'], DATETIME_FORMAT).date() == closed_at
```
TheDoubleK: Maybe `closed_on` rather than `closed_at`? Or even `date_closed_on` to make it even clearer.

tests/test_notify_buyers_when_requirements_close.py
```diff
@@ -0,0 +1,186 @@
...
+        {"applicationsClosedAt": "2016-09-06T23:59:59.000000Z", "status": "closed"},
+        {"applicationsClosedAt": "2016-09-07T23:59:59.000000Z", "status": "closed"},
+        {"applicationsClosedAt": "2016-09-08T23:59:59.000000Z", "status": "closed"},
+        {"applicationsClosedAt": "2016-09-05T23:59:59.000000Z", "status": "closed"},
+        {"applicationsClosedAt": "2016-09-09T23:59:59.000000Z", "status": "closed"},
+        {"applicationsClosedAt": "2016-09-05T23:59:59.000000Z", "status": "closed"},
+    ])
+
+    assert get_closed_briefs(api_client, datetime.date(2016, 9, 5)) == [
+        {"applicationsClosedAt": "2016-09-05T23:59:59.000000Z", "status": "closed"},
```
TheDoubleK: Might be nice to mix up the time of day a bit?  Not that it makes any real difference.

TheDoubleK:
Left some really minor comments but essentially looks good.
:+1: with or without changes, your call if they're worth it.

tests/test_notify_buyers_when_requirements_close.py
allait: All of our requirements close at the same time at the moment

pcraig3:
Yup, also looks like it makes sense to me. 👍

PR #60, opened 2016-09-28
Add script to save signedAgreementPath

idavidmcdonald:
For suppliers who have returned their framework agreement, if there is no signedAgreementPath set in the FrameworkAgreement table then check s3 for their most recently uploaded signed framework agreement file and update the FrameworkAgreement record with this file path.


TheDoubleK:
This has now been run against production, is a one-off, no need for further deliberations. 
All good.  :+1:

PR #61, opened 2016-09-29
Save countersigned details for framework agreements

idavidmcdonald:
Save `countersignedAgreementPath` and `countersignedAgreementReturnedAt` for framework agreements. For suppliers who have returned their framework agreement and had it countersigned (indicated by a countersigned file in s3), we set their `countersignedAgreementPath` and
`countersignedAgreementReturnedAt` in the FrameworkAgreement table. The time for the `countersignedAgreementReturnedAt` is taken from s3 meta data for when the file was uploaded (this is likely the closest indicator we have for when a file was countersigned).


TheDoubleK:
Oneoff, now been run against prod. Too late to fix it now!
:+1:

PR #62, opened 2016-10-11
Make scripts executable

pcraig3:
Some of our `/oneoff` scripts weren't actually executable, so now they are.


TheDoubleK:
Sterling work :+1:

Issue #11, opened 2015-11-11
Requirement out of date for python-json-logger?


Newman101:
Hi all. 

Just for your information, it might be that one of your requirements is out of date. I tried to install dependencies and got this error message: 

```
Downloading/unpacking python-json-logger==v0.1.3 (from digitalmarketplace-utils==13.0.0)
  Could not find a version that satisfies the requirement python-json-logger==v0.1.3 (from digitalmarketplace-utils==13.0.0) (from versions: 
0.0.1, 0.0.2, 0.0.3, 0.0.4, 0.0.5, 0.0.6, 0.1.0, 0.1.1, 0.1.2, 0.1.4)
Cleaning up...
No distributions matching the version for python-json-logger==v0.1.3 (from digitalmarketplace-utils==13.0.0)
Storing debug log for failure in /root/.pip/pip.log
```
Looks like the latest version is v0.1.3, but the latest version is 0.1.4. The package installs if you execute ```pip install -r requirements.txt``` from the console straight away, but for example the installation aborts on PyCharm. 

For the latest version, please refer to: https://pypi.python.org/pypi/python-json-logger/0.1.4

Thanks.

PR #63, opened 2016-10-11
Update json-logger

pcraig3:
Looks like [a fairly small set of changes](https://github.com/madzak/python-json-logger/compare/948353f4bdeec41b5bf00c845c8eb04ff334e8ca...6425c16b9e0de05e9f886861151230d502f8c68a) which preserve seconds instead of losing them.

Will also resolve [Issue #11](https://github.com/alphagov/digitalmarketplace-scripts/issues/11).

Tests still pass.

***

Don't know what we do about updating things in our `requirements` files more generally.


TheDoubleK:
:+1:

PR #64, opened 2016-10-19
Get back model data and output to CSV

pcraig3:
Created a class that gets back all table rows from the database using the API and prints them to CSV.  In theory, works for anything with an "iter" method, and in practice works out of the box for `buyers`, `suppliers`, `briefs`, and `brief_responses`.

Ash is asking for weekly printouts of (DOS-related) data, and so this is the first step.  
👉  [This is the data he's looking for](https://docs.google.com/spreadsheets/d/1DlkUYNSDgK1vjNB4LQgKP6AQJD9g89kOAw5MPwTVPqs/edit?ts=57e513ba#gid=0). 👈 

Things left to do in subsequent stories are:
- [Join logic](https://www.pivotaltracker.com/story/show/132760553): (ie, how many applications do we get per published brief)
- [Set up a regular jenkins thing](https://www.pivotaltracker.com/story/show/131737693) that prints out stuff once a week

### DIFF
--- a/scripts/get-model-data.py
+++ b/scripts/get-model-data.py
@@ -32,10 +32,11 @@ from dmscripts import logging
 CONFIGS = {
     'buyers': {
         'base_model': 'users',
-        'keys': ('id', 'createdAt', 'role'),
+        'keys': ('id', 'emailAddress', 'createdAt', 'role'),
         'get_data_kwargs': {},
         'process_rules': {
-            'createdAt': format_datetime_string_as_date
+            'emailAddress': remove_username_from_email_address,
+            'createdAt': format_datetime_string_as_date,
         },
         'filter_rules': [
             ('role', '==', 'buyer'),
--- a/scripts/get-model-data.py
+++ b/scripts/get-model-data.py
@@ -1,0 +1,135 @@
+#!/usr/bin/env python3
+"""Export a CSV of all models or something
+
+Looks for a corresponding find_{model}_iter() method in the api client and returns
+all of the results in a CSV with only requested keys being used as columns
+Before generating the final csv, it is possible to:
+- filter the list, looking for only those entries with a particular value
+- do some additional processing on a value
+- sort the list
+
+Usage:
+    scripts/get-model-data.py [-h] <stage> <api_token> <model> <output_dir> [--limit=<limit>][-v]
+
+Options:
+    -h --help       Show this screen.
+    -v --verbose    Print apiclient INFO messages.
+"""
+import sys
+sys.path.insert(0, '.')
+
+from docopt import docopt
+from dmapiclient import DataAPIClient
+from dmscripts.env import get_api_endpoint_from_stage
+from dmscripts.models.modeltrawler import ModelTrawler
+from dmscripts.models.utils import process_collection, return_filtered_collection, return_sorted_collection
+from dmscripts.models.process_rules import format_datetime_string_as_date, remove_username_from_email_address
+from dmscripts.models.writecsv import csv_path, write_csv
+
+from dmscripts import logging
+
+
+CONFIGS = {
+    'buyers': {
+        'base_model': 'users',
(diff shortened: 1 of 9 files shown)


models/utils.py
```diff
@@ -0,0 +1,43 @@
...
+    ops = {
+        '>': operator.gt,
+        '<': operator.lt,
+        '>=': operator.ge,
+        '<=': operator.le,
+        '==': operator.eq
+    }
+
+    for key, op, to_compare in rules:
+        collection = list(filter(lambda x: ops[op](x[key], to_compare), collection))
```
pcraig3: Yeah, I think we're planning to use something like [pandas](http://pandas.pydata.org/) in the (near) future and at that point we can replace this logic.

Definitely agreed -- don't really want to go down this road any further.

models/utils.py
```diff
@@ -0,0 +1,43 @@
...
+        return collection
+
+    # `contains` would be trickier to include because it switches the argument order
+    ops = {
+        '>': operator.gt,
+        '<': operator.lt,
+        '>=': operator.ge,
+        '<=': operator.le,
+        '==': operator.eq
+    }
```
pcraig3: Yeah, I think this is stretching the bounds of what we want to be doing.

Works for our current use case though.

models/modeltrawler.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def func_wrapper(*args, **kwargs):
+        _self = args[0]
+        print_counts = kwargs.pop('print_counts', None)
+
+        for model in func(*args, **kwargs):
+            if print_counts:
+                func_wrapper.counter += 1
+                if func_wrapper.counter % print_counts == 0:
+                    print('{} {} returned in {}s'.format(func_wrapper.counter, _self.model, _self.get_time_running()))
```
allait: I'd prefer if we used `logger` methods instead of `print`. It adds the current timestamp to all messages and is generally easier to control verbosity etc.

There are examples of how to do this in the index services and buyer notification scripts.

models/modeltrawler.py
```diff
@@ -0,0 +1,117 @@
...
+
+        return _filter_keys_inner
+
+    def get_data(self, keys=None, limit=None, **kwargs):
+        models = self._model_iter(**kwargs)
+
+        if limit is not None:
+            models = itertools.islice(models, limit)
+
+        # not sure concurrent execution function is doing anything
```
allait: This is a weekly script, I think it should be ok to run it with a single thread and skip all the concurrency stuff altogether

models/modeltrawler.py
```diff
@@ -0,0 +1,117 @@
+import time
```
allait: Let's move this to `dmscripts/`

models/utils.py
```diff
@@ -0,0 +1,43 @@
+import operator
+from datetime import datetime
+from dmutils.formats import DATE_FORMAT, DATETIME_FORMAT
+
+
+format_datetime_string_as_date = lambda x: datetime.strptime(x, DATETIME_FORMAT).strftime(DATE_FORMAT)
```
allait: I'd put the functions we want to use in the CONFIG dictionaries in a separate file to make it easier to see what's available

scripts/get-model-data.py
```diff
@@ -0,0 +1,127 @@
...
+CONFIGS = {
+    'buyers': {
+        'base_model': 'users',
+        'keys': ('id', 'createdAt', 'role'),
+        'get_data_kwargs': {'print_counts': 100},
+        'process_rules': {
+            'createdAt': format_datetime_string_as_date
+        },
+        'filter_rules': [
+            ('role', '==', 'buyer'),
```
allait: We could use lambdas here directly, but if the plan is to transition to pandas query language soon, then we can keep this as is

allait:
👍

PR #66, opened 2016-11-15
Script to generate counterpart G-Cloud 8 PDFs

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/128842655

I started off trying to parameterise the `generate-agreement-sinature-pages` script, but it turns out that there are enough differences with generating the counterpart pages that it makes more sense to me as a separate (but very similar) thing.

The main differences are different conditions on when a file should be generated, an extra file that needs to be copied and later removed, (i.e. the signature file), and file names.

It could probably have been parameterised and just one script, but that would have involved an explosion of command line arguments and/or a bunch of conditional steps in generate_agreement_signatre_pages.py` that wouldn't be very pretty.

### DIFF
--- a/dmscripts/export_framework_applicant_details.py
+++ b/dmscripts/export_framework_applicant_details.py
@@ -5,6 +5,7 @@ else:
     import csv
 
 from multiprocessing.pool import ThreadPool
+from dmutils.formats import dateformat
 
 
 LOTS = {
@@ -62,10 +63,15 @@ def add_framework_info(client, framework_slug):
         supplier_framework = client.get_supplier_framework_info(record['supplier_id'], framework_slug)
         supplier_framework = supplier_framework['frameworkInterest']
         supplier_framework['declaration'] = supplier_framework['declaration'] or {}
+        supplier_framework['countersignedPath'] = supplier_framework['countersignedPath'] or ''
+        supplier_framework['countersignedAt'] = dateformat(supplier_framework['countersignedAt']) or ''
 
         return dict(record,
                     declaration=supplier_framework['declaration'],
-                    onFramework=supplier_framework['onFramework'])
+                    onFramework=supplier_framework['onFramework'],
+                    countersignedPath=supplier_framework['countersignedPath'],
+                    countersignedAt=supplier_framework['countersignedAt'],
+                    )
 
     return inner
 
@@ -93,6 +99,8 @@ def get_csv_rows(records, framework_slug):
         "supplier_id",
         "supplier_name",
         "on_framework",
+        "countersigned_at",
+        "countersigned_path",
         ] + LOTS[framework_slug] + DECLARATION_FIELDS[framework_slug]
 
     return headers, rows
@@ -103,6 +111,8 @@ def create_row(record, framework_slug):
         'supplier_id': record['supplier']['id'],
         'supplier_name': record['supplier']['name'],
         'on_framework': record['onFramework'],
+        'countersigned_at': record['countersignedAt'],
+        'countersigned_path': record['countersignedPath'],
     }
     row.update({(lot, record['counts'][lot]) for lot in LOTS[framework_slug]})
     row.update(((field, record['declaration'].get(field, "")) for field in DECLARATION_FIELDS[framework_slug]))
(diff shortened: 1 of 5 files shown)


scripts/generate-counterpart-signature-pages.py
```diff
@@ -0,0 +1,59 @@
...
+
+    client = DataAPIClient(get_api_endpoint_from_stage(STAGE), API_TOKEN)
+
+    headers, rows = find_suppliers_with_details(client, FRAMEWORK)
+
+    render_html_for_suppliers_awaiting_countersignature(rows, FRAMEWORK, TEMPLATE_FOLDER, html_dir)
+
+    html_pages = os.listdir(html_dir)
+    html_pages.remove('{}-signature-page.css'.format(FRAMEWORK))
+    html_pages.remove('{}-countersignature.png'.format(FRAMEWORK))
```
risicle: ...especially seeing as the lines to _remove_ the resource files created outside the `for` loop in `render_html_for_suppliers_awaiting_countersignature` are here, in the script.

dmscripts/export_framework_applicant_details.py
```diff
@@ -62,10 +63,15 @@ def inner(record):
         supplier_framework = client.get_supplier_framework_info(record['supplier_id'], framework_slug)
         supplier_framework = supplier_framework['frameworkInterest']
         supplier_framework['declaration'] = supplier_framework['declaration'] or {}
+        supplier_framework['countersignedPath'] = supplier_framework['countersignedPath'] or ''
```
TheDoubleK: Hmm, yeah the "blame" on this file is all me but the pattern was a straight-up copy and paste job from somewhere else that does a similar thing with multi-threaded building up of data. I'd rather leave this as is right now, or at least talk through the other approach in person.

dmscripts/generate_agreement_signature_pages.py
```diff
@@ -27,6 +27,26 @@ def render_html_for_successful_suppliers(rows, framework, template_dir, output_d
...
+    template_css_path = os.path.join(template_dir, '{}-signature-page.css'.format(framework))
+    countersignature_img_path = os.path.join(template_dir, '{}-countersignature.png'.format(framework))
+    for data in rows:
+        if data['on_framework'] is False or not data['countersigned_at'] or data['countersigned_path']:
+            print("SKIPPING {}: on_f={} at={} path={}".format(
+                data['supplier_id'],
+                data['on_framework'],
+                data['countersigned_at'],
+                data['countersigned_path'])
+            )
```
TheDoubleK: These aren't errors - just agreements that have already been generated, but definitely better to use a proper logger - I'll fix it.

dmscripts/generate_agreement_signature_pages.py
```diff
@@ -27,6 +27,26 @@ def render_html_for_successful_suppliers(rows, framework, template_dir, output_d
     shutil.copyfile(template_css_path, os.path.join(output_dir, '{}-signature-page.css'.format(framework)))
 
 
+def render_html_for_suppliers_awaiting_countersignature(rows, framework, template_dir, output_dir):
+    template_path = os.path.join(template_dir, '{}-counterpart-signature-page.html'.format(framework))
+    template_css_path = os.path.join(template_dir, '{}-signature-page.css'.format(framework))
+    countersignature_img_path = os.path.join(template_dir, '{}-countersignature.png'.format(framework))
+    for data in rows:
+        if data['on_framework'] is False or not data['countersigned_at'] or data['countersigned_path']:
```
TheDoubleK: Good point.

scripts/generate-counterpart-signature-pages.py
TheDoubleK: This doesn't remove the files themselves (they need to be there for the rendering of PDFs to work), but removes their names from the list of files to be converted to PDFs (because trying to convert a css or png file to PDF will error).

I agree it's weird though.

TheDoubleK:
I pushed a couple of commits to remove unused stuff, re-order the `not` to the end of the line and use tempfile for the HTML files.

I'm not going to change the import stuff right now, but yes it's something we can do as pat of a wider fix across all these scripts.

Think that's everything that's needed for now, holler if I missed something.

scripts/generate-counterpart-signature-pages.py
risicle: Ohhhhh right that makes more sense.

PR #67, opened 2016-11-25
Add script to upload counterpart signature pages

TheDoubleK:
This has already been run against production for the initial batch of countersigned agreements.

We will soon want to add it as part of a Jenkins job to run periodically, so would be good to get a proper review still!

Tests won't pass now until this is merged:
 - [x] https://github.com/alphagov/digitalmarketplace-utils/pull/289

### DIFF
--- a/dmscripts/upload_counterpart_agreements.py
+++ b/dmscripts/upload_counterpart_agreements.py
@@ -1,10 +1,9 @@
 import getpass
 
-from dmutils.documents import generate_timestamped_document_upload_path
+from dmutils.documents import generate_timestamped_document_upload_path, generate_download_filename, \
+    COUNTERPART_FILENAME
 
-from dmscripts.bulk_upload_documents import get_supplier_id_from_framework_file_path, generate_download_filename
-
-COUNTERPART_FILENAME = "agreement-countersignature.pdf"
+from dmscripts.bulk_upload_documents import get_supplier_id_from_framework_file_path
 
 
 def upload_counterpart_file(bucket, framework_slug, file_path, dry_run, client, logger):
--- a/dmscripts/upload_counterpart_agreements.py
+++ b/dmscripts/upload_counterpart_agreements.py
@@ -1,0 +1,47 @@
+import getpass
+
+from dmutils.documents import generate_timestamped_document_upload_path
+
+from dmscripts.bulk_upload_documents import get_supplier_id_from_framework_file_path, generate_download_filename
+
+COUNTERPART_FILENAME = "agreement-countersignature.pdf"
+
+
+def upload_counterpart_file(bucket, framework_slug, file_path, dry_run, client, logger):
+    supplier_id = get_supplier_id_from_framework_file_path(file_path)
+    supplier_framework = client.get_supplier_framework_info(supplier_id, framework_slug)
+    supplier_framework = supplier_framework['frameworkInterest']
+    supplier_name = supplier_framework['declaration']['nameOfOrganisation']
+    download_filename = generate_download_filename(supplier_id, COUNTERPART_FILENAME, supplier_name)
+
+    upload_path = generate_timestamped_document_upload_path(
+        framework_slug,
+        supplier_id,
+        "agreements",
+        COUNTERPART_FILENAME
+    )
+    try:
+        if not dry_run:
+            # Upload file
+            with open(file_path) as file_contents:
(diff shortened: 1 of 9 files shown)


dmscripts/upload_counterpart_agreements.py
```diff
@@ -0,0 +1,46 @@
...
+            with open(file_path) as file_contents:
+                bucket.save(upload_path, file_contents, acl='private', move_prefix=None,
+                            download_filename=download_filename)
+                logger.info("UPLOADED: '{}' to '{}'".format(file_path, upload_path))
+
+            # Save filepath to framework agreement
+            client.update_framework_agreement(
+                supplier_framework['agreementId'],
+                {"countersignedAgreementPath": upload_path},
+                'upload-counterpart-agreements script run by {}'.format(getpass.getuser())
```
TheDoubleK: I guess the assumption when we started using this pattern (yes, it is a pattern copied from earlier scripts) was that these scripts would be run by developers on the team using GDS-issued machines where the username will be the more meaningful GDS LDAP username (first name and surname).

The alternative is adding another command-line argument, which also seems a bit weird (when was the last time you came across a script where you have to tell it your name before it will run?)

TheDoubleK:
I'm not sure where you get the idea that the local path to the PDF file ends up being the path it's put under in the bucket?  The `generate_timestamped_document_upload_path` method from utils is used to determine where to save the file, and the local path isn't referenced here?

risicle:
Hmmmmmmmm I'll have to take a closer look then I got a bit lost.

dmscripts/upload_counterpart_agreements.py
```diff
@@ -0,0 +1,46 @@
...
+                upload_path, supplier_framework['agreementId'])
+            )
+        else:
+            logger.info("[Dry-run] UPLOAD: '{}' to '{}'".format(file_path, upload_path))
+            logger.info("[Dry-run] countersignedAgreementPath='{}' for agreement ID {}".format(
+                upload_path, supplier_framework['agreementId'])
+            )
+
+    except Exception as e:
+        logger.info("ERROR {}".format(e.message))
```
TheDoubleK: What's the "good behaviour" way to catch everything that might go wrong that I can't necessarily anticipate right now?

TheDoubleK: Ah OK found something about this... I'm going to go with 
```
except:
   e = sys.exc_info()[0]
   ...
```
Unless that's not what you meant?

dmscripts/upload_counterpart_agreements.py
```diff
@@ -0,0 +1,46 @@
+import getpass
+
+from dmutils.documents import generate_timestamped_document_upload_path, generate_download_filename, \
+    COUNTERPART_FILENAME
+
+from dmscripts.bulk_upload_documents import get_supplier_id_from_framework_file_path
+
+
+def upload_counterpart_file(bucket, framework_slug, file_path, dry_run, client, logger):
```
risicle: Oh dear I just started following the weird pattern with my script wanting to fit in... :)

dmscripts/upload_counterpart_agreements.py
risicle: :shrug: guess it's ok just wasn't aware of it

dmscripts/upload_counterpart_agreements.py
risicle: No, I just meant doing a "catch all" exception is thought to be a bit risky as it would catch & ignore all kinds of severe internal python errors that shouldn't be recovered from (memory errors etc). But lots of people do it anyway. I have on occasion.

TheDoubleK: Hmm, yeah there is that.  I guess the point here is to avoid the script dying altogether if one file fails to upload.  I think this is OK as is for running the script locally, but if we're planning to run it as a job on Jenkins I should probably do a bit of due diligence about catching just things we might expect to fail.

TheDoubleK:
OK I've made the suggested changes and added some explicit error handling (rather than a catch-all).  Do these seem like a reasonable list of errors to catch, log and recover from?  Did I miss anything?

risicle:
Bit more effort than I was expecting - frankly I often just dump the string representation of the exception to the log.

risicle:
Other than that I think it's ok :+1:

PR #68, opened 2016-12-06
CSV processing

galund:
Story: https://www.pivotaltracker.com/story/show/132760553

New `process-model-data.py` script processes CSV files downloaded using the existing `get-model-data.py` script.

It creates CSV files based on the queries in dmscripts.models.queries (and configured in the script).

### DIFF
--- a/scripts/get-model-data.py
+++ b/scripts/get-model-data.py
@@ -86,11 +86,14 @@ CONFIGS = {
         'keys': (
             'briefId',
             'supplierId',
-            'createdAt'
+            'createdAt',
+            'supplierName',
+            'submittedAt'
         ),
         'get_data_kwargs': {},
         'process_rules': {
             'createdAt': format_datetime_string_as_date,
+            'submittedAt': format_datetime_string_as_date,
         },
         'filter_rules': [],
         'sort_by': 'createdAt'
--- a/scripts/get-model-data.py
+++ b/scripts/get-model-data.py
@@ -44,7 +44,7 @@ CONFIGS = {
         ],
         'sort_by': 'createdAt'
     },
-    'suppliers': {
+    'supplier_users': {
         'base_model': 'users',
         'keys': ('id', ('supplier', 'supplierId'), 'createdAt', 'role'),
         'get_data_kwargs': {},
--- a/scripts/get-model-data.py
+++ b/scripts/get-model-data.py
@@ -88,12 +88,14 @@ CONFIGS = {
             'supplierId',
             'createdAt',
             'supplierName',
-            'submittedAt'
+            'submittedAt',
+            'essentialRequirements'
         ),
         'get_data_kwargs': {},
         'process_rules': {
             'createdAt': format_datetime_string_as_date,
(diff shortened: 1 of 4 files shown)


pcraig3:
Leaving in a minute, so I'll expand upon this tomorrow, but just a few observations:

**Good**
- bit funny, but using the argument names to find the files seems pretty good
- `essentialRequirements: all` is a good (if temporary) bit of code
- pandas looks really simple and powerful, which is nice

**Bad** 
- no tests 😱 
- pep8 problem
- some of the commits could be squashed

Looks good though! Ash will be thrilled 😉

idavidmcdonald:
So we are currently working on changing how a supplier applies for a brief. Whilst we currently receive data for `essentialRequirements` as a list of bools in a brief response, we will no longer be doing this and instead will receive `essentialRequirementsMet` which will be a single boolean (that must be answered as `True` to progress through the flow).

However, we will at some point have both flows running and therefore have brief responses with either key. So in the future this script may need adapting to support either type of format. 

This might not mean that you need to change anything related to your pull request but it may just mean we will at a later point need to make this script be able to handle both cases.

scripts/get-model-data.py
```diff
@@ -44,7 +44,7 @@
         ],
         'sort_by': 'createdAt'
     },
-    'suppliers': {
+    'supplier_users': {
```
pcraig3: Seems like a good change as well.

scripts/process-model-data.py
```diff
@@ -0,0 +1,49 @@
...
+sys.path.insert(0, '.')
+
+from docopt import docopt
+import pandas as pd
+import os.path
+
+from dmscripts.models import queries
+
+
+CONFIG = {
```
allait: I don't think this really achieves the idea of separating configuration from the code. Modifying this structure doesn't allow us to:
* add/remove output fields
* add new output files that are joining different CSVs
* change the count logic for one-to-many joins

So for example if we want to add a file that shows how many briefs each supplier has applied for we'd have to write a new separate function, even though all the logic behind it is more or less the same.

scripts/process-model-data.py
```diff
@@ -0,0 +1,49 @@
...
+if __name__ == '__main__':
+    arguments = docopt(__doc__)
+
+    CSV_DIR = arguments['<csv-dir>']
+    QUERY_NAME = arguments['<query-name>']
+
+    query = get_query(QUERY_NAME)
+    data_frames = dict()
+    # query params give us the table (data frame) names we need...
+    for data_frame_name in query.__code__.co_varnames[:query.__code__.co_argcount]:
```
allait: This feels backwards to me, but it's probably coming back to the same point I've made above.

I don't think function is a good format to *store* configuration, which is what this is essentially using it for. I think we need to generalise the logic and explicitly provide model/file names as parameters instead of introspecting them from argument names.

allait:
https://github.com/alphagov/digitalmarketplace-scripts/pull/78 got merged, so I think we can close this now

PR #69, opened 2016-12-06
[#134342161] add mark-definite-framework-results.py script

risicle:
~~This has no tests yet so caveat emptor.~~

Story https://www.pivotaltracker.com/story/show/134342161

This generalizes the action of marking SupplierFrameworks as passed/failed hopefully across frameworks. It does so by evaluating them against json schemas. json schemas for dos2 declarations and services are included.

I don't want to repeat the contents of the script docstring too much, but declaration schemas that are composed of an inner `definitions` entry `notDefiniteFail` sub-schema will also have that evaluated against the declaration to allow the script the differentiate between those it can mark as definite failures and those that are *discretionary*.

A supplier's submitted draft services will also be evaluated against a draft_service schema if supplied. Failing that schema would cause the draft service's `status` to be marked as `failed`.

Suppliers who end up with no non-`failed` services will be a definite fail for the framework.

~~As you can see I haven't even pep8'd this yet.~~

~~And my switch to `find_draft_services_iter` has caused a glitch in the mocks which I need to fix.~~


scripts/mark-definite-framework-results.py
```diff
@@ -0,0 +1,66 @@
...
+
+Usage: mark-definite-framework-results.py [options] <stage> <api_token> <framework_slug>
+            <declaration_definite_pass_schema_path> [<draft_service_schema_path>]
+
+--updated-by=<user_string>        Specify updated_by string to use in API requests
+--dry-run                         Don't actually perform any updates on API
+--reassess-passed-sf              Don't skip supplier_frameworks with onFramework already True
+--reassess-failed-sf              Don't skip supplier_frameworks with onFramework already False
+--reassess-failed-draft-services  Don't skip draft_services with "failed" status
+-v, --verbose                     Produce more detailed console output
```
TheDoubleK: Running locally with `-v` I get the following error for one of the suppliers, which prevents it from determining a result (runs fine without the `-v`):
```
2016-12-12 16:02:35,465 dmutils.logging ERROR failed to format log message: "u'status'" not found
Traceback (most recent call last):
  File "/Users/kevinkeenoy/gdsworkspace/digitalmarketplace-scripts/venv/lib/python2.7/site-packages/dmutils/logging.py", line 117, in format
    msg = msg.format(**record.__dict__)
KeyError: "u'status'"
2016-12-12 16:02:35,465 script DEBUG 		Failed service_schema @ : {u'status': u'submitted', u'frameworkFramework': u'dos', u'bespokeSystemInformation': True, u'supplierId': 700008, u'links': {u'self': u'http://localhost:5000/draft-services/194', u'copy': u'http://localhost:5000/draft-services/194/copy', u'complete': u'http://localhost:5000/draft-services/194/complete', u'publish': u'http://localhost:5000/draft-services/194/publish'}, u'lotSlug': u'digital-outcomes', u'helpGovernmentImproveServices': False, u'openStandardsPrinciples': True, u'updatedAt': u'2016-12-12T15:47:55.325566Z', u'frameworkSlug': u'digital-outcomes-and-specialists-2', u'locations': [u'Offsite', u'Scotland'], u'supplierName': u"Kev's Pies", u'frameworkStatus': u'open', u'dataProtocols': True, u'lotName': u'Digital outcomes', u'frameworkName': u'Digital Outcomes and Specialists 2', u'lot': u'digital-outcomes', u'performanceAnalysisTypes': [u'A/B and multivariate testing', u'Data analysis', u'Data cleansing', u'Data visualisation'], u'id': 194, u'createdAt': u'2016-12-12T15:47:35.110985Z'} is not valid under any of the given schemas
```
Unfortunately no more stack trace than that though.

schemas/digital-outcomes-and-services-2.declaration.json
```diff
@@ -0,0 +1,69 @@
...
+                "unfairCompetition": {"enum": [true]},
+                "readUnderstoodGuidance": {"enum": [true]},
+                "understandTool": {"enum": [true]},
+                "understandHowToAskQuestions": {"enum": [true]},
+                "proofOfClaims": {"enum": [true]},
+                "skillsAndResources": {"enum": [true]},
+                "accuratelyDescribed": {"enum": [true]},
+                "employersInsurance": {
+                    "type": "string",
+                    "pattern": "^(Yes |Not applicable )"
```
TheDoubleK: We probably don't need the trailing spaces after the values here?  Might make it a bit more readable I guess so fine either way.  I suppose once we auto-generate this stuff we'll probably end up with the whole long value here :/

schemas/digital-outcomes-and-services-2.declaration.json
```diff
@@ -0,0 +1,69 @@
...
+                "MI": {"enum": [true]},
+                "canProvideFromDayOne": {"enum": [true]},
+                "evidence": {"enum": [true]},
+                "environmentallyFriendly": {"enum": [true]},
+                "consistentDelivery": {"enum": [true]},
+                "skillsAndCapabilityAssessment": {"enum": [true]},
+                "continuousProfessionalDevelopment": {"enum": [true]},
+                "customerSatisfactionProcess": {"enum": [true]},
+                "equalityAndDiversity": {"enum": [true]},
+                "civilServiceValues": {"enum": [true]}
```
TheDoubleK: Looks like the schema is missing "TRUE" entries for the following declaration keys:
```
technologyCodesOfPractice
serviceStandard
ongoingEngagement
transparentContracting
publishContracts
safeguardOfficialInformation
safeguardPersonalData
```

schemas/digital-outcomes-and-services-2.declaration.json
```diff
@@ -0,0 +1,69 @@
...
+                "influencedContractingAuthority": {"enum": [false]},
+                "confidentialInformation": {"enum": [false]},
+                "misleadingInformation": {"enum": [false]},
+                "unspentTaxConvictions": {"enum": [false]},
+                "GAAR": {"enum": [false]}
+            }
+        }
+    ],
+    "definitions": {
+        "notDefiniteFail": {
```
TheDoubleK: I might be misunderstanding, but I think `notDefiniteFail` is the wrong name here altogether (although the script is doing the right thing with them) - if these ones aren't right then it *is* a definite fail.  It's the ones above (taxEvasion...GAAR) where it's not a definite fail if the schema isn't matched.

dmscripts/mark_definite_framework_results.py
```diff
@@ -0,0 +1,135 @@
+from collections import Counter
+import logging
+
+
+import jsonschema
+
+
+def _passes_validation(candidate, schema, logger, schema_name="schema", tablevel=0, loglevel=logging.INFO):
```
TheDoubleK: After initially  complaining about passing a logger around you've really gone to town here with the tablevel and everything :)

risicle:
Yeah I was really wedded to the `notDefiniteFail` term when I started out but the more I used it the more weird it seemed. I could re-assign the two as `minimum` and `strict`?

scripts/mark-definite-framework-results.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python
+"""
+Uses a number of supplied json schemas to determine whether suppliers on a particular framework should be marked as
```
risicle: Yup to all that.

scripts/mark-definite-framework-results.py
risicle: Yes I was getting this too - I disregarded it and assumed it was part of the weird way we set up our logger...?

scripts/mark-definite-framework-results.py
risicle: Oh except the subschema bit - all that happened there was that I designed in a graceful behaviour to follow if the subschema is not found by the script in the superschema. It's not a big deal, and no, none of our current frameworks do things this way. It would correspond to a framework where all of the questions were mandatory.

dmscripts/mark_definite_framework_results.py
risicle: Yup, I followed the existing precedent, the tablevel is because the same function is called both under the supplier_framework and draft service contexts, which work at different tablevels.

scripts/mark-definite-framework-results.py
```diff
@@ -0,0 +1,66 @@
...
+if __name__ == "__main__":
+    from docopt import docopt
+    args = docopt(__doc__)
+
+    client = DataAPIClient(get_api_endpoint_from_stage(args["<stage>"], "api"), args["<api_token>"])
+    updated_by = args["--updated-by"] or getpass.getuser()
+
+    declaration_definite_pass_schema = json.load(open(args["<declaration_definite_pass_schema_path>"], "r"))
+
+    declaration_not_definite_fail_schema = \
```
risicle: `discretionary` doesn't describe which border of the discretionary classification it defines - the line between `fail` and `discretionary` or the line between `discretionary` and `fail`.

I'm thinking maybe `minimum`, or `bareMinimum` (`minimumRequirements` except "requirements" is already a confusingly used term elsewhere)

`baseline`?

schemas/digital-outcomes-and-services-2.declaration.json
risicle: Yes I appear to have missed a section. I want to get cath's input on this because the way it's been worded in the story makes it a little unclear whether these ones should be considered minimal requirements or auto-pass requirements.

scripts/mark-definite-framework-results.py
TheDoubleK: How about `mustHave`?  In order to be a possible pass the declaration must have all the things listed in that section.

TheDoubleK:
OK let's give this a go after a rebase :+1:

I wonder if we should remove the `-v` option altogether if it's broken though?

risicle:
I don't think `-v` is broken, I think this is the bug @galund was fixing in utils. He's about to merge that fix so once it's in I'll try running it again and see how it fairs...

PR #70, opened 2016-12-09
134342371-DOS2-CSV-file-showing-application-status

benvand:
https://www.pivotaltracker.com/story/show/135769263
> Probably a bit overkill for one script but I needed to have a look around.

* Create a master CSV builder from @risicle s previous script.
* Create a once off script using the new builder: `generate-dos-2-master-csv.py`
* Refactor old script to use new builder: `generate-g-cloud-8-master-csv.py` (slightly pointless but acts as a test sort of)
* Tidy up the multiple `mock_data_client` declarations.
* Reorder imports.
* Heavily test.


risicle:
I... think you muddled up your rebase at some point, and now you're including a bunch of other peoples unrelated commits (kev's, pauls...) in this PR. Probably just need to do a `rebase -i` and strip out their commits, then rebase on top of master.

(My guess is at some point you performed the single-arg variant of `rebase` where you should have used the three-arg variant with `--onto`)

PR #71, opened 2016-12-09
NO-TICKET_add_venv3

benvand:
####This should pass once the intermittent test failure fix is merged  https://github.com/alphagov/digitalmarketplace-scripts/pull/73

As per https://github.com/alphagov/digitalmarketplace-api/blob/master/setup.cfg#L2
* Add venv3* to pytest `norecursedirs`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,4 +3,4 @@ exclude = ./venv,./venv3
 max-line-length = 120
 
 [pytest]
-norecursedirs= venv*
+norecursedirs= venv venv3


risicle:
huh... why is the glob not working then?

benvand:
Good point. I think I misnamed my virtualenv to something like `env3` then didn't think about the glob. 
Would be good to standardise this across branches. I also think a case could be made for 'explicit is better than implicit' in the context of excluding files. I mean it would be a horrible idea to call anything `venv.txt` and actually want to commit it to a branch but even so. 

ALso I had a look into excluding this with some sort of `venv[3 ]` regex but py.test doesn't have support for optional suffixes.
http://doc.pytest.org/en/latest/customize.html#confval-norecursedirs

TheDoubleK:
I've kicked the tests off again - think they'd failed with the intermittent failure you fixed elsewhere.

Might be worth squashing the two commits here into a single one.

TheDoubleK:
ohmyglob

benvand:
@TheDoubleK squashed

TheDoubleK:
Cheers! :+1:

PR #73, opened 2016-12-09
NO-TICKET-cannot_rely_on_order_here (intermittently failing test)

benvand:
* Can't rely on order here. Just check the pairs are present and correct


risicle:
Quite right. Uses `ThreadPool.imap_unordered` if you dig down deep enough. :+1:

PR #75, opened 2016-12-13
[#134342643]export_framework_applicant_details dos2 additions

risicle:
https://www.pivotaltracker.com/story/show/134342643

Couldn't help a few tidyups but all I've really done is add the dos2 details. This is another thing that feels like it should be sourced from the frameworks. e.g. by a question having an `is_details` flag or something better named.


TheDoubleK:
The docstring in the main driver script (`scripts/export-framework-applicant-details.py`) could do with updating too.

TheDoubleK:
This isn't working for me locally - `find_suppliers` gets a list of supplier IDs OK: `[{'supplier_id': 579009}, {'supplier_id': 700007}, {'supplier_id': 700008}]` but then the call to `records = pool.imap(partial(add_supplier_info, client), records)` wipes the `records` list empty rather than augmenting the items as it should.

TheDoubleK:
So I just end up with an empty file.

TheDoubleK:
Digging in a bit more, the API is never called after the initial fetch of supplier IDs - i.e. `add_supplier_info`, `add_framework_info` and `add_submitted_draft_counts` are never actually called.

risicle:
I'll deal with this - I may have pushed the wrong thing.

risicle:
Fixed - though this really needs a test that tests the whole operation. Unit tests aren't enough.

TheDoubleK:
The `registeredAddress` field no longer exists - it needs to be three fields `registeredAddressBuilding`, `registeredAddressTown`, and `registeredAddressPostcode`.

The Pivotal story doesn't have submitted draft counts - it instead has 
```
Lot statuses:
* Lot 1 - pass or no_application
* Lot 2 - pass or no_application
* Lot 3 - pass or no_application
* Lot 4 - pass or no_application
```
Which *after running the script to flag `failed` lots* will be equivalent to zero-submitted=>no_application and >0-submitted => pass.  But it feels a bit weird to have a pass/no_application listed for a supplier who has failed overall.  So maybe stick with the 1 and 0 counts of submitted.  Or grab Tobi for a quick chat to see what she thinks is better.  Actually, I think stick with the counts of "submitted" - it kind-of hides the fact that people might have submitted but failed the lot questions, but i think it's the right information.

Also, the docstring in the main script still needs a quick update. (i.e. don't say it will only work for G8)

risicle:
- [ ] do that stuff ^

TheDoubleK:
We should show some staging data to people to check over, but all looks :+1: to me.

risicle:
Yeah I'll do a staging run before I merge actually.

risicle:
@andrewchong make a thumbs up signal in my direction. Merging.

PR #76, opened 2016-12-15
[#134342161]dos2 declaration schema: require "status" field

risicle:
doesn't look like anything else is going to check it


TheDoubleK:
:+1:

PR #78, opened 2016-12-16
Join CSV with group counts

allait:
Adds additional operations support to the get-model-data scripti config.

CONFIG is now an ordered list of items. Each item has a name and can be requested or skipped when the script is invoked.

Since items are processed in order, each one can depend on CSVs for the previous ones existing in the OUTPUT_DIR.

Each CONFIG item can describe configuration for several steps, which are invoked by the script in order. First step is loading the data from one of the sources:

* Data API, if `base_model` key is set
* Existing CSV file if `model` key is set
* A join of two existing CSVs if `join` key is set

Once the data is loaded it is transformed into pandas `DataFrame` object, which is then used by the other processing steps.

Currently supported config keys include:
1. `process_fields` defines individual field transformations to be applied to the data
2. `add_counts` adds count columns for groups of related data (eg counts of brief responses grouped by status for each brief)
3. `filter_query` is a pandas query expression to filter output CSV rows
4. `keys` is setting the set and order of output CSV columns
5. `sort_by` sets the list of columns to use for ordering the CSV


galund:
FWIW (not a lot, this rather re-runs last week's argument, but you did ask me to take a look ;) ):

 - I can see where you're going here, and it does what you'd said it would do
 - I still think YAGNI, there are some good (fairly historic) blog posts on code vs configuration, like http://blog.urth.org/2011/01/06/config-versus-code/ 
- it's going to be tricky to make this support rollup/pivot tables, but I'm sure it'll be possible to come up with some config for that

allait:
@galund what do you mean rollup/pivot tables? It already does count + group by to add successful/unsuccessful brief application counts columns, which is the only use for them we've had so far.

galund:
Ahh, I didn't get the 'this-dash-group-value' thing, but I see it now :)

pcraig3:
It 's short, it works and it's not much more than what was here before.

Getting rid of all the handmade logic in `utils` seems like quite a good move, and pretty much all of the methods in `queries` are short and totally generic.

If we do need to change/add stuff, this looks like a good base on which to do it, and even if we don't it still seems simple enough what it's doing.  

👍

PR #79, opened 2016-12-16
Add export_framework_results_reasons script

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/134342559

This is a version of the existing `export_g8_suppliers.py` script that should hopefully be general purpose for new frameworks going forward.

I have duplicated some methods from these earlier scripts rather than reusing them, because I think we should delete the framework-specific versions altogether at some point in the near future.

Rather than explicitly encoding the "right" answers to declaration questions in the
script this version uses the same JSON schema originally used to determine PASS/FAIL (supplied in the arguments) to determine which questions caused a "fail" or "discretionary" result.

**NOTE**: This has *NO TESTS* as yet.  But seeing as we need to run it first thing on Monday it'd be good to get an eye over it first.  Tests will be forthcoming.  Honest.

Also rolled in to this PR is the option to only generate agreement files for a select list of supplier IDs - I needed this to do: https://www.pivotaltracker.com/story/show/135769263


risicle:
I wonder if we could rename this script. `export-definite-framework-results-reasons.py`? Something to more vividly denote the purpose of the script...

dmscripts/export_framework_suppliers.py
```diff
@@ -0,0 +1,390 @@
...
+
+from multiprocessing.pool import ThreadPool
+
+
+DRAFT_STATUSES = [
+    'completed',
+    'failed',
+    'draft',
+]
+LOTS = [
```
risicle: Lookout - hardcoding lots...

dmscripts/export_framework_suppliers.py
```diff
@@ -0,0 +1,390 @@
...
+    return sum(1
+               for service in record["services"]
+               if field_label in service.get(field_id, []))
+
+
+def make_field_title(field_id, field_label):
+    return "{} {}".format(field_id, field_label)
+
+
+def make_fields_from_content_question(question, record):
```
risicle: I think actually `fields_from_content_question` would be a better name for this - `make_` implies almost that it has a side effect. In fact in the past, often when writing a generator I've been explicit that it's a generator - `fields_from_content_question_gen` or something...

dmscripts/export_framework_suppliers.py
```diff
@@ -0,0 +1,390 @@
...
+                "|".join(service.get(field_id, "") for service in record["services"])
+            )
+    else:
+        yield (
+            question["id"],
+            "|".join(str(service.get(question["id"], "")) for service in record["services"])
+        )
+
+
+def make_fields_from_content_questions(questions, record):
```
risicle: Is this called from anywhere?

dmscripts/export_framework_suppliers.py
```diff
@@ -0,0 +1,390 @@
...
+            fieldnames = [key for key, _ in row]
+            self._csv_writers[handler.NAME] = csv.DictWriter(self._csv_files[handler.NAME], fieldnames=fieldnames)
+            self._csv_writers[handler.NAME].writeheader()
+
+        return self._csv_writers[handler.NAME]
+
+    def csv_path(self, handler):
+        return os.path.join(self.output_dir, handler.NAME + '.csv')
+
+    def __enter__(self):
```
risicle: Ooh fancy!

dmscripts/export_framework_suppliers.py
```diff
@@ -0,0 +1,390 @@
...
+        declaration_baseline_schema=None
+):
+    records = find_suppliers_with_details(client,
+                                          content_loader,
+                                          framework_slug,
+                                          declaration_definite_pass_schema,
+                                          declaration_baseline_schema
+                                          )
+
+    handlers = [SuccessfulHandler(), FailedHandler(), DiscretionaryHandler()]
```
risicle: *I* see what you're doing. Neato.

dmscripts/export_framework_suppliers.py
```diff
@@ -0,0 +1,390 @@
...
+            handler.NAME: 0 for handler in handlers
+        }
+
+    def write_row(self, record):
+        for handler in self.handlers:
+            if handler.matches(record):
+                self._counters[handler.NAME] += 1
+                row = handler.create_row(record)
+                return self.csv_writer(handler, row).writerow(dict(row))
+        raise Exception("record not handled by any handler")
```
risicle: I would use a `ValueError` or maybe even a custom subclass - the problem with emitting bare `Exception`s is they are only catchable using `except Exception` which of course is troublesome.

dmscripts/export_framework_suppliers.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+class MultiCSVWriter(object):
+    """Manage writing to multiple CSV files"""
+    def __init__(self, output_dir, handlers):
+        self.output_dir = output_dir
+        self.handlers = handlers
+        self._csv_writers = dict()
+        self._csv_files = dict()
+        self._counters = {
```
risicle: You can use `collections.Counter` for things like this and not have to do initialization routines (unmentioned keys are considered `0`)

dmscripts/export_framework_suppliers.py
TheDoubleK: yes - let's do that.

dmscripts/export_framework_suppliers.py
TheDoubleK: Yep, makes sense.

dmscripts/export_framework_suppliers.py
TheDoubleK: Nope - deleting it right now.

dmscripts/export_framework_suppliers.py
TheDoubleK: Yep.

dmscripts/export_framework_suppliers.py
TheDoubleK: I'll look at that now.

TheDoubleK:
@ris - still no tests :sob: but have taken your previous comments into consideration here, and also have the script only outputting the right things and not a bunch of extra superfluous stuff.

dmscripts/export_framework_results_reasons.py
```diff
@@ -0,0 +1,376 @@
...
+
+
+def add_draft_services(client, framework_slug, lot=None, status=None):
+    def inner(record):
+        drafts = client.find_draft_services(record["supplier_id"], framework=framework_slug)
+        drafts = drafts["services"]
+
+        drafts = [
+            draft for draft in drafts
+            if (not lot or draft["lotSlug"] == lot) and (not status or draft["status"] == status)
```
risicle: I can never remember the order of precedence when leading statements with `not`

risicle:
Cool looking better - though we should probably get some tests behind this before it's thumbs-up time...

risicle:
Tobi just mentioned that it would be useful to include the `failed_discretionary` column in the `failed` output csv so that the csv consumer is able to see whether each definite fail also happens to have declared any discretionary data (presumably because the "definite fail" could be overturned on appeal and they would have another piece of discretionary data to assess). I'll leave you to hack that out with her on the story though.

risicle:
Tobi also mentioned that they're going to need reasons for individual service failure, and I guess this script is the place that that belongs as much as anywhere. It's a little tricky on two fronts:

 - extracting errors from `anyOf` rules is a little more complex to do. Especially if we're to make this flexible to arbitrary json schemas, not just json schemas which us `anyOf` to do a switch across `lotSlug` - we can talk about this when you get back - I have a few ideas...
 - of course, you can't do this as a row-per-supplier - it's per-service data. But you could of course divert this data off into its own file which is row-per-service and allow the user to do the join themselves.

TheDoubleK:
Agreed with @risicle that we should merge this a a record of what has been already run for DOS2.

PR #81, opened 2016-12-20
mark-definite-framework-results: reformat description in docstring for docopt's sake

risicle:
docopt was interpreting a description line (mid-paragraph) beginning
with `--reassess-failed-draft-services` as a declaration of that option, and
so, it being declared twice, it was refusing to allow it to be used, claiming
`--reassess-failed-draft-services is not a unique prefix:...`


allait:
👍

PR #82, opened 2016-12-20
Question csv download fix

galund:
Remove hard-coded lots [#136021199]
 - default behaviour is now to output all questions, with no
   context;
 - to choose a specific lot, pass it in using new 'context' CLI
   option;
 - old behaviour of repeating questions if they appeared in
   multiple lots is gone, you can now specify only one lot (or
   none, for all questions).

### DIFF
--- a/scripts/generate-questions-csv.py
+++ b/scripts/generate-questions-csv.py
@@ -30,8 +30,8 @@ import yaml
 if __name__ == '__main__':
     arguments = docopt(__doc__)
 
-    path_to_manifest = arguments['<DM-content-root-dir>']
-    content_loader = ContentLoader(path_to_manifest)
+    content_path = arguments['<DM-content-root-dir>']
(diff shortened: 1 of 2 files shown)


scripts/generate-questions-csv.py
```diff
@@ -8,22 +12,31 @@
...
 from docopt import docopt
 from dmcontent.content_loader import ContentLoader
 from dmscripts.generate_questions_csv import generate_csv
+import yaml
 
 if __name__ == '__main__':
     arguments = docopt(__doc__)
 
-    path_to_manifest = arguments['<path_to_manifest>']
+    path_to_manifest = arguments['<DM-content-root-dir>']
```
allait: We should probably rename the variable too then

galund:
I think this is good to go now?

pcraig3:
Yup, it works and it's easy!

👍

PR #84, opened 2017-01-16
137170159-framework-success-email-script

benvand:
#### Script to send supplier success email

*  `dmscripts/supplier_data.py`
  * `SupplierFrameworkData` class to supply data from api. Attached `users` and `services` to a `SupplierFramework` object
  * `SuccessfulSupplierContextForNotify` gets the personalisation for the 'Your application result - if successful email'

* Add script `scripts/notify_successful_suppliers_for_framework.py` to send mails

https://www.pivotaltracker.com/story/show/137170159


dmscripts/supplier_data.py
```diff
@@ -0,0 +1,103 @@
...
+    def __init__(self, client, target_framework_slug):
+        self.client = client
+        self.target_framework_slug = target_framework_slug
+
+    def get_supplier_frameworks(self):
+        """Return supplier frameworks."""
+        return self.client.find_framework_suppliers(self.target_framework_slug)['supplierFrameworks']
+
+    def get_supplier_users(self):
+        """Return a dict, supplier id: application status."""
```
TheDoubleK: don't think this docstring is accurate?  Isn't it `supplier_id: [list of supplier users]`?

dmscripts/supplier_data.py
```diff
@@ -0,0 +1,103 @@
...
+        """Return a dict of lot status for each lot on the framework.
+        """
+        any_lots = False
+
+        lot_dict = OrderedDict((lot_name, 'No application') for lot_name in self.framework_lots)
+        for service in supplier_framework['services']:
+            status = service['status']
+            if status == 'submitted':
+                any_lots = True
+            lot_dict[service['lotName']] = self.STATUS_MAP[status]
```
TheDoubleK: This works for the "one-service-per-lot" lots, but for the `user-research-studios` lot it's possible to have multiple services for the lot, and so maybe some submitted and some not-submitted services in the lot.  If there's a single `submitted` user-research-studio then the application will be `Successful` overall.

(And in general - e.g. all G-Cloud lots - we should expect that there could be multiple draft services per lot, some submitted and some not.)

dmscripts/supplier_data.py
```diff
@@ -0,0 +1,103 @@
...
+            return {}
+        return lot_dict
+
+    def get_user_personalisation(self, user, supplier_framework):
+        """Get dict of all info required by template given a user and framework."""
+        lot_dict = self.get_lot_dict(supplier_framework)
+        if not lot_dict:
+            return {}
+        lot_dict = dict(zip(
+            ['lot_%d' % i for i in range(1, len(self.framework_lots) + 1)],
```
TheDoubleK: I'm not sure the ordering of the lots as they come out of the API is guaranteed, so numbering by the order in the dict seems like it might be inconsistent.  If we need the lots to be numbered then we should maybe just hard code the mapping of lot names here to be on the safe side.

TheDoubleK:
Left a few comments, and the commits could do with a squash.

TheDoubleK:
Couple of things:
1. I confirmed with @allait that the way to get around the pip dependency issue for github requirements is to include the requirement in the app that needs it - i.e. in this case include the notify client import in scripts `requirements.txt`
2. The script naming convention is to hyphenate `scripts` filenames, and use underscores in the accompanying `dmscripts` file that has the helper methods.

benvand:
@TheDoubleK 
Renamed file, squashed commits, brought in dependency.
Ordering satisfied by api as per face-to-face convo

TheDoubleK:
Looks good, let's get this merged :+1:

PR #85, opened 2017-01-27
Document reusable approach in README

pcraig3:
We want to raise awareness of how to generate large data dumps in future so that we don't end up with ever more slightly divergent script files.

Documented some of our new code in the README as a recommended best practice.


TheDoubleK:
Seems good.  I'm in the middle of a massive refactor of existing scripts at the moment to make them more usable for framework close (which is when the majority of them get run). 
This will be just one more thing to rebase against.

TheDoubleK:
:+1:

PR #87, opened 2017-02-06
Question csv export tweak

galund:
Changes required as a result of the 'one question per page' thing, and a better understanding of the fields we will expect to be using for G9. Mostly changes around the interpretation of multiquestions and sections.

### DIFF
--- a/dmscripts/generate_questions_csv.py
+++ b/dmscripts/generate_questions_csv.py
@@ -25,13 +25,16 @@ def get_questions(questions):
         return question
 
     def get_question(question):
-        if question.questions:
-            return [
-                augment_question_data(nested_question, question.get('name'), question.get('hint'))
-                for nested_question in question.questions
-            ]
+        # only multiquestions have nested (child) questions that we want to flatten into our CSV
+        try:
+            nested_questions = question.questions
+        except AttributeError:
+            return [question]
 
-        return [question]
+        return [
+            augment_question_data(nested_question, question.get('name'), question.get('hint'))
+            for nested_question in nested_questions
+        ]
 
     questions_list = []
     for question in questions:
--- a/dmscripts/generate_questions_csv.py
+++ b/dmscripts/generate_questions_csv.py
@@ -47,21 +47,27 @@ def return_rows_for_sections(sections):
 
     for section in sections:
         for question in get_questions(section.questions):
-            page_title = question.get('parent', {}).get('question')
+            # Using get_source because we want the original markdown, not the rendered HTML
+            # this means that no context can be reflected in the question content, i.e. our passed-in
+            # context can only influence whether a question is displayed or not.
+            # TODO if this turned out to matter, instead of get_source, we could convert back to markdown(!)
+
+            parent = question.get('parent')
+            page_title = parent.get_source('question') if parent else None
             section_and_page_title = ' / '.join(filter(None, [section.name, page_title]))
 
             # section.description is no longer expected to be used
-            multiquestion_description = question.get('parent', {}).get('description')
-            multiquestion_hint = question.get('parent', {}).get('hint')
+            multiquestion_description = parent.get_source('description') if parent else None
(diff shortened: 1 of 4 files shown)


dmscripts/generate_questions_csv.py
```diff
@@ -45,11 +47,31 @@ def return_rows_for_sections(sections):
...
+            # Using get_source because we want the original markdown, not the rendered HTML
+            # this means that no context can be reflected in the question content, i.e. our passed-in
+            # context can only influence whether a question is displayed or not.
+            # TODO if this turned out to matter, instead of get_source, we could convert back to markdown(!)
+
+            parent = question.get('parent')
+            page_title = parent.get_source('question') if parent else None
+            section_and_page_title = ' / '.join(filter(None, [section.name, page_title]))
+
+            # section.description is no longer expected to be used
```
TheDoubleK: When you say "no longer expected to be used" - you mean just for the purposes of this CSV file?

scripts/generate-questions-csv.py
```diff
@@ -12,12 +12,15 @@
...
+    scripts/generate-questions-csv.py <DM-content-root-dir> <output-file> [options]
 
+--framework=<slug>
+--question-set=<question-set-directory-name>
+--context=<yaml>
 -h --help    show this message
 
 Example:
-    scripts/generate-questions-csv.py /path/to/dm-frameworks/ ~ --framework=g-cloud-9 --context="lot: SaaS"
+    scripts/generate-questions-csv.py /path/to/dm-frameworks/ ~/mystuff.csv --framework=g-cloud-9 --context="lot: SaaS"
```
TheDoubleK: Ah, OK, I see, we can only exclude things that have a `depends: on: lot...` entry in the content using the context.

TheDoubleK: OK OK, so we do that with `question-set` rather than `context`.

I think the upshot of my comments here is that the docstring for the script should explicitly say that  there are two possible values for `question-set` and that they are `declaration` and `services`.

dmscripts/generate_questions_csv.py
```diff
@@ -95,9 +118,16 @@ def generate_csv(output_directory, framework_slug, content_loader, context):
     for index, row in enumerate(rows):
         row.extend([''] * (max_length - len(row)))
 
-    with open('{}/{}-questions.csv'.format(output_directory, framework_slug), 'wb') as csvfile:
-        writer = csv.writer(csvfile, delimiter=b',', quotechar=b'"')
-        header = ["Page title", "Page title hint", "Question", "Hint"]
+    filename = framework_slug
```
TheDoubleK: None of this `filename` stuff gets used at all now as the `output_file` is passed in, so this section can probably be deleted. 

(Although I like the idea of giving the script a directory where it will generate sensibly-named files for you depending on the params you pass in, so would be happy to see that fleshed out too as an alternative to deleting this `filename` stuff.)

dmscripts/generate_questions_csv.py
galund: Hmm, so this is what I understood on the day I wrote it (from @allait possibly - or not, the mistake is most likely mine).

The issue is, that with 'one question per page', there is no place for the section description to appear?

TheDoubleK: On the summary page maybe?  But anyway, for the purposes of this script it's fine not to include it I think.

dmscripts/generate_questions_csv.py
galund: The problem with generating the name, was that it didn't take into account the context you passed in on the command line. So I realised the task was then to generate a filename from a (potentially, if not in practice) arbitrarily complex YAML expression, which was a bit eww. It also makes for less surprises in terms of trashing a file you didn't expect. I'll remove the obsolete code.

TheDoubleK:
:+1:

PR #90, opened 2017-03-08
Add brief status to successful brief response CSV

pcraig3:
Ash was getting confused that all successful brief_responses
were included in the `successful_brief_responses` CSV, including
responses for briefs that were later withdrawn (or drafts :/).

Thus he wants one extra column in the `successful_brief_responses`
CSV: the current status of the brief that was applied to.

He would like it to be the final column in the CSV.

### DIFF
--- a/scripts/get-model-data.py
+++ b/scripts/get-model-data.py
@@ -123,6 +123,7 @@ CONFIGS = [
             'supplierId',
             'supplierName',
             'submittedAt',
+            'status',
         ),
         'sort_by': ['briefId', 'submittedAt']
     },


scripts/get-model-data.py
```diff
@@ -123,6 +123,7 @@
             'supplierId',
             'supplierName',
             'submittedAt',
+            'status',
```
idavidmcdonald: This looks good but I haven't seen any of this script csv stuff before. My assumption is that this is using the brief status rather than the brief response status because is defined in line 98 what properties are being retrieved for the brief response.

Assuming my assumption is correct, then this should be good 👍

TheDoubleK: Nice work with the meta assuming.

pcraig3: Yup, exactly.

PR #91, opened 2017-03-10
Add script to send stats to Performance Platform

TheDoubleK:
This new script hits our API statistics endpoint and reformats the data into the form required by our performance platform datasets before posting it to the relevant place.

The stats reformatting methods (`_format_statistics`, `_label_and_count`, `_sum_counts`, `_find`) are based on those used in the admin app, but there are enough differences here that I don't think it's worth extracting these methods into shared utils right now.

I guess in future if the performance platform works out for us then the admin app stats views will become obsolete and be removed anyway.

### DIFF
--- a/dmscripts/send_stats_to_performance_platform.py
+++ b/dmscripts/send_stats_to_performance_platform.py
@@ -1,5 +1,3 @@
-import json
-
 import backoff
 import base64
 import requests
--- a/dmscripts/send_stats_to_performance_platform.py
+++ b/dmscripts/send_stats_to_performance_platform.py
@@ -1,0 +1,163 @@
+import json
+
+import backoff
+import base64
+import requests
+
+from datetime import datetime, timedelta
+import dmapiclient
+
+from dmscripts.helpers import logging_helpers
+from dmscripts.helpers.logging_helpers import logging
+
+HOURLY_TIME_FORMAT = '%Y-%m-%dT%H:00:00+00:00'  # On the hour exactly
+DAILY_TIME_FORMAT = '%Y-%m-%dT00:00:00+00:00'  # Midnight
+PERFORMANCE_PLATFORM_URLS = {
+    "day": {
+        "stage": "https://www.performance.service.gov.uk/data/gcloud/applications-by-stage",
+        "lot": "https://www.performance.service.gov.uk/data/gcloud/applications-by-lot",
+    },
+    "hour": {
+        "stage": "https://www.performance.service.gov.uk/data/gcloud/applications-by-stage-realtime",
+        "lot": "https://www.performance.service.gov.uk/data/gcloud/applications-by-lot-realtime",
+    },
+}
+
+logger = logging_helpers.configure_logger({'dmapiclient': logging.WARNING})
+
+
+def _format_statistics(stats, category, groupings):
+    return _label_and_count(stats[category], groupings)
+
+
+def _label_and_count(stats, groupings):
+    data = {
+        label: _sum_counts(stats, filters)
+        for label, filters in groupings.items()
+    }
+    return data
+
+
+def _sum_counts(stats, filter_by=None, sum_by='count'):
+    return sum(
+        statistic[sum_by] for statistic in stats
+        if not filter_by or all(
+            _find(statistic.get(key), value)
+            for key, value in filter_by.items()
+        )
+    )
+
+
+def _find(statistic, filter_value):
+    if isinstance(filter_value, list):
+        return statistic in filter_value
+    else:
+        return statistic == filter_value
+
+
+def _generate_id(timestamp, period, data_type, data_item):
+    # Instructions from Performance Platform:
+    # _id should be a unique url-friendly, base64-encoded, UTF8 encoded concatenation identifier, formed from:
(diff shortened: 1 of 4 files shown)


dmscripts/send_stats_to_performance_platform.py
```diff
@@ -0,0 +1,163 @@
...
+        )
+    return res.status_code
+
+
+def applications_by_stage(stats):
+    return _format_statistics(
+        stats,
+        'interested_suppliers',
+        {
+            'interested': {
```
risicle: It might be nice to document how all these `_format_statistics` et al. functions are supposed to work - what the meaning of these blocks here is - i.e. that lists cause an `in` test, other types lead to an equality test...

dmscripts/send_stats_to_performance_platform.py
```diff
@@ -0,0 +1,163 @@
...
+        {
+            lot['slug']: {
+                'lot': lot['slug'],
+                'status': 'submitted',
+            } for lot in framework['lots']
+        }
+    )
+
+
+def send_by_stage_stats(stats, timestamp_string, period, pp_bearer):
```
risicle: `send_by_stage_stats`

wat?

Oh - `send` `by_stage` `stats`...

dmscripts/send_stats_to_performance_platform.py
```diff
@@ -0,0 +1,163 @@
...
+        return statistic in filter_value
+    else:
+        return statistic == filter_value
+
+
+def _generate_id(timestamp, period, data_type, data_item):
+    # Instructions from Performance Platform:
+    # _id should be a unique url-friendly, base64-encoded, UTF8 encoded concatenation identifier, formed from:
+    # _timestamp, service (= gcloud), period (= day or hour), dataType (= applications-by-stage/lot), stage/lot
+    id_bytes = '{}-gcloud-{}-{}-{}'.format(timestamp, period, data_type, data_item).encode('utf-8')
```
risicle: are we just using a single `g-cloud` for everything? What happens when we want to do this for dos3? Or are we going to break things down by framework in the `data_type`? Either way, we should perhaps take this as an argument...

dmscripts/send_stats_to_performance_platform.py
```diff
@@ -0,0 +1,163 @@
...
+    return send_data(data, PERFORMANCE_PLATFORM_URLS[period]['lot'], pp_bearer)
+
+
+@backoff.on_exception(backoff.expo, dmapiclient.HTTPError, max_tries=5)
+def send_framework_stats(data_api_client, framework_slug, period, pp_bearer):
+    stats = data_api_client.get_framework_stats(framework_slug)
+    framework = data_api_client.get_framework(framework_slug)['frameworks']
+    now = datetime.utcnow()
+    # _timestamp is the *start* of the period to which the data relates but the "reading" here is at the *end*
+    # of the period, so need to subtract the period from the current time
```
risicle: are you sure this is the right approach to take? it seems that both approaches are as wrong as each other (it's not marking a time period at all, it's marking an infinitesimal point in time. I'm not sure doing this juggle is totally necessary and is possibly a little confusing/surprising...?

TheDoubleK: It's a requirement from the performance platform to send the timstamp for the beginning of the period the data applies to as part of the payload.  "I don't make the rules", etc...

dmscripts/send_stats_to_performance_platform.py
TheDoubleK: We only have one dataset that exists on performance platform at the moment. If we want new non- `gcloud` datasets they will need to set them up for us, and so it felt like using the name  in here is OK.   If it's parameterised then people might get the idea they could send a different value and expect it to work.

tests/test_send_stats_to_performance_platform.py
```diff
@@ -0,0 +1,359 @@
+import json
```
risicle: Correct me if I'm wrong but `send_data` isn't itself tested at any point. Oversight or did you think it's too simplistic to bother?

TheDoubleK: I was losing the will to live :/
I can add a test to check it makes the right call if you reckon it's worth it.  I didn't feel like there would be a lot of value in it, but I tend to be slacker than I should be about these things.

risicle: Yeah, don't do it on my behalf. I would have just done the tests as end-to-end tests using requests mock to make sure the whole thing works.

TheDoubleK: Hmm, yep maybe I should have thought of that - you mean just set up the data responses in a mock client, hit `send_framework_stats` and then check the right calls got made to performance platform?  That would have been nice.  I can change it if you reckon it's worth it now, but I'm not sure it's worth re-doing.

dmscripts/send_stats_to_performance_platform.py
TheDoubleK: I've added a comment to explain these a bit.  Anything else that needs changing or can we live with this now?

risicle:
Yeah I think we're good :+1:

PR #92, opened 2017-03-13
Create CSV of all DOS services

idavidmcdonald:
For this pivotal story - https://www.pivotaltracker.com/story/show/140921671

I suggest reading through the 4 commit messages to give you an overview of what this PR contains.

### DIFF
--- a/dmscripts/models/queries.py
+++ b/dmscripts/models/queries.py
@@ -42,6 +42,11 @@ def process_fields(rules, data):
     return data
 
 
+def rename_fields(columns, data):
+    """ Replace any column names using a dict of 'old column name': 'new column name' """
+    return data.rename(columns=columns)
+
+
 def sort_by(columns, data):
     return data.sort_values(by=columns)
 
--- a/scripts/get-model-data.py
+++ b/scripts/get-model-data.py
@@ -22,6 +22,7 @@ Arguments:
     models:
         buyer_users
         supplier_users
+        dos_services
         briefs
         brief_responses
         successful_brief_responses
@@ -42,6 +43,26 @@ from dmscripts.models import queries
 
 from dmscripts.helpers.logging_helpers import logging, configure_logger
 
+DOS_SPECIALIST_ROLES = [
+    "agileCoach",
+    "businessAnalyst",
+    "communicationsManager",
+    "contentDesigner",
+    "securityConsultant",
+    "deliveryManager",
+    "designer",
+    "developer",
+    "performanceAnalyst",
+    "portfolioManager",
+    "productManager",
+    "programmeManager",
+    "qualityAssurance",
+    "serviceManager",
+    "technicalArchitect",
+    "userResearcher",
+    "webOperations"
+]
+DOS_SPECIALIST_ROLES_PRICE_MAX = [s + 'PriceMax' for s in DOS_SPECIALIST_ROLES]
 
 CONFIGS = [
     {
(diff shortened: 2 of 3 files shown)


dmscripts/models/queries.py
```diff
@@ -42,6 +42,15 @@ def process_fields(rules, data):
...
 
 
+def rename_fields(rules, data):
+    """ Return data after renaming any keys using the renaming rules """
+    for key in data.keys():
+        if key in rules.keys():
+            data[rules[key]] = data.pop(key)
+
+    return data
+
```
benvand: So the above function does a couple of things that I'd be sure to note, if not be wary of.

First things first I'm not sure `rules` is the best name possible for the `rules` arg. When I read it I sort of expected a dict like:
`{'string_field_name': some_method_to_transform_said_field}`
maybe something like:
`rules = {
'name': str.capitalize
'age': lambda i: 'You are %d years old' % i
}`
What the function is doing is applying a mapping to the existing field names which is a rule in itself i.e. _The rule_ is if the field is in the mapping dict then swap it with the value in that mapping. Or _apply this mapping to these fields_.

Second thing, you might already know this, with dicts you don't need to call keys when iterating over them. Dicts automatically iterate over their keys. 
`[i for i in d.keys()] == [i for i in d] == [i for i in d.__iter__()]`
It's a bit of a weird one tbh because it's a part of Python that doesn't feel very intuitive.

Finally, dicts are mutable. It always feels a bit weird modifying them in place. You're relying on other people realising what you're up to.
So although you're returning, and reassigning to, `data` when you use this function below you can just as easily not. 
That's why `list.extend` and `dict.update` return None

Something that _appears_ to negate the above is:
```
def rename_fields(mapping, data):
    """If the key has an entry in the mapping replace the key with the value in mapping, otherwise skip."""
    return {mapping.get(key, None) or key: value for key, value in data.iteritems()}
```

scripts/get-model-data.py
```diff
@@ -68,6 +89,28 @@
...
+                'lotSlug',
+                'status',
+                'supplierId',
+                'supplierName'
+            ] + DOS_SPECIALIST_ROLES_PRICE_MAX  # We use price max as a proxy for a service having a role
+        ),
+        'process_fields': {
+            role: bool for role in DOS_SPECIALIST_ROLES_PRICE_MAX
+        },
+        'rename_fields': {
```
benvand: This can be 
`'rename_fields': dict(zip(DOS_SPECIALIST_ROLES_PRICE_MAX, DOS_SPECIALIST_ROLES))`
I think.
Saves you constructing the DOS_SPECIALIST_ROLES_PRICE_MAX list twice.

dmscripts/models/queries.py
benvand: So the above function does a couple of things that I'd be sure to note, if not be wary of.

First things first I'm not sure about the name `rules` for the `rules` arg. When I read it I sort of expected a dict like:
`{'string_field_name': some_method_to_transform_said_field}`
maybe something like:
`rules = {
'name': str.capitalize
'age': lambda i: 'You are %d years old' % i
}`
The function is applying a mapping to the existing field names which is a rule in itself i.e. _The rule_ is if the field is in the mapping dict then swap it with the value in that mapping. Or _apply this mapping to these fields_.

Second thing, you might already know this, with dicts/ DataFrames you don't need to call keys when iterating over them. Dicts/ DataFrames automatically iterate over their keys. 
`assert [i for i in data.keys()] == [i for i in data] == [i for i in data.__iter__()]`
It's a bit of a weird one tbh because it's a part of Python that doesn't feel very intuitive.

Watch out for the whole dict mutability thing. It always feels a bit weird modifying them in place. Although you're returning, and reassigning to, `data` when you use this function below you can just as easily not bother.
`rename_fields(rules, data)`
Is going to do the same as
`data = rename_fields(rules, data)`

That's why `list.extend` and `dict.update` return None

I have a funny feeling that changing the key names/ rebuilding the dataframe from a new dict might mess with the ordering.

In fact scrap that, I believe this is what you want :)
http://stackoverflow.com/a/11354850

benvand: Just found this:
http://stackoverflow.com/a/11354850

scripts/get-model-data.py
idavidmcdonald: Nice suggestion :)

dmscripts/models/queries.py
idavidmcdonald: I see what you mean about the `rules` naming, it does feel a little ambigious. Can you think of a potentially better name? We could use `columns` like pandas do?(http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.rename.html)

I didn't realise that about not needing to call `.keys()`. That's good to know! It does seem a little weird but at the same time makes sense (although feels a little less readable/explicit if you are a python noob like me)

idavidmcdonald:
Great shout on using panda's built in renaming functionality Ben. I think I've addressed most of your points.

The one I didn't address was returning and resetting `data` rather than just editing it. Although there is no reason for it, the rest of our query functions are doing this so I went for consistency, especially if you look at the `get-model-data.py` script where data is reset all of the time (although you could argue it's worth changing this for all the query functions...)
image

benvand:
Looks great.
The data reassignment isn't an issue now. You actually _have_ to reassign it now because you're don't passing in `inplace=True` to `rename`.  That's the same as what the other functions do in this file so all good.

👍 

[LGTM](https://lgtm.in/i/Y8nqju1Tp)

PR #94, opened 2017-03-20
Update agreement-generation scripts to use details from API

TheDoubleK:
Previously we needed to pass in a JSON file but now this JSON will come from the frameworkAgreementDetails an the API response for the framework.

### DIFF
--- a/scripts/generate-framework-agreement-counterpart-signature-pages.py
+++ b/scripts/generate-framework-agreement-counterpart-signature-pages.py
@@ -12,23 +12,27 @@ in the API response for the given framework.
 
 The 'frameworkAgreementDetails' JSON object from the API should look something like this:
 {
-    "frameworkName": "Digital Outcomes and Specialists 2",
+    "contractNoticeNumber": "2016/S 217-395765",
     "frameworkAgreementVersion": "RM1043iv",
-    "frameworkRefDate": "16-01-2017",
-    "frameworkStartDate": "30/01/2017",
     "frameworkEndDate": "29/01/2018",
     "frameworkExtensionLength": "12 months",
+    "frameworkRefDate": "16-01-2017",
+    "frameworkStartDate": "30/01/2017",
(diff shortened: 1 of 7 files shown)


risicle:
Note I used different names for the key `pageNumber` (->`signaturePageNumber`)  and made it and `pageTotal` integers, not strings.

risicle:
oh and also I got rid of the `frameworkName` because I thought we agreed that the whole framework dictionary should be the context passed to the template, hence it would have access to the `name`...?

scripts/generate-framework-agreement-counterpart-signature-pages.py
```diff
@@ -58,8 +61,10 @@
     args = docopt(__doc__)
 
     framework_slug = args['<framework_slug>']
-    framework_kwargs = declaration_definite_pass_schema = json.load(open(args["<framework_json>"], "r"))
     client = DataAPIClient(get_api_endpoint_from_stage(args['<stage>']), args['<api_token>'])
+    framework = client.get_framework(framework_slug)['frameworks']
+    framework_kwargs = framework['frameworkAgreementDetails']
+    framework_kwargs['frameworkName'] = framework['name']
```
risicle: Ok. I thought we were going to just pass the whole framework in to the template, but :man_shrugging:

TheDoubleK: I thought about it but I like the idea of passing in a completely flat set of kwargs to `.render()` rather than have to have the templates know about the structure of our API responses.

Also, it was more typing to add `frameworkAgreementDetails.xxx` to everything in the template.

TheDoubleK: Happy to go that way if you reckon it would be nicer.

risicle: My thinking was just that using a regular framework object would be the most no-surprises route. Any developer coming to the template just gets this *"I know this - this is UNIX"* feel. Agree that it's more typing though.

PR #95, opened 2017-03-22
[#127983495]upload_counterpart_agreement notifications

risicle:
Story https://www.pivotaltracker.com/story/show/127983495

One thing I'm considering is whether it's the right thing to swallow `EmailError` and continue to the next supplier - if we're not able to send an email we maybe don't want to blaze ahead and upload a bunch of files if there's a likelihood that *all* the emails are going to fail and we've just lost the chance to send the users emails...?

### DIFF
--- a/dmscripts/upload_counterpart_agreements.py
+++ b/dmscripts/upload_counterpart_agreements.py
@@ -34,7 +34,7 @@ def upload_counterpart_file(
     supplier_name = supplier_framework['declaration']['nameOfOrganisation']
     download_filename = generate_download_filename(supplier_id, COUNTERPART_FILENAME, supplier_name)
 
-    notify_emails = dm_notify_client and frozenset(chain(
+    email_addresses_to_notify = dm_notify_client and frozenset(chain(
         (supplier_framework["declaration"]["primaryContactEmail"],),
         (
             user["emailAddress"]
@@ -71,7 +71,7 @@ def upload_counterpart_file(
                 upload_path, supplier_framework['agreementId'])
             )
 
-        for notify_email in (notify_emails or ()):
+        for notify_email in (email_addresses_to_notify or ()):
             if not dry_run:
                 dm_notify_client.send_email(notify_email, notify_template_id, {
                     "framework_slug": framework["slug"],
--- a/dmscripts/upload_counterpart_agreements.py
+++ b/dmscripts/upload_counterpart_agreements.py
@@ -5,8 +5,8 @@ from boto.exception import S3ResponseError
 from dmapiclient import APIError
 from dmutils.documents import generate_timestamped_document_upload_path, generate_download_filename, \
     COUNTERPART_FILENAME
-from dmutils.email.exceptions import EmailError
 from dmutils.email.helpers import hash_string
+from dmutils.email.exceptions import EmailError
 
 from dmscripts.bulk_upload_documents import get_supplier_id_from_framework_file_path
 from dmscripts.helpers import logging_helpers
@@ -21,6 +21,7 @@ def upload_counterpart_file(
         data_api_client,
         dm_notify_client=None,
         notify_template_id=None,
+        notify_fail_early=True,
         logger=None,
         ):
     if bool(dm_notify_client) != bool(notify_template_id):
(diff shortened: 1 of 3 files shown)


risicle:
(I also suspect I should add some tests that verify this behaviour)

(Though the closer I look at it the more it looks like the exception swallowing should happen in the script and not `upload_counterpart_agreements` to make it more useful as a utility function :man_shrugging:)

scripts/upload-counterpart-agreements.py
```diff
@@ -33,25 +40,48 @@
...
+        raise ValueError("Either specify both --notify-key and --notify-template-id or neither")
+
     stage = arguments['<stage>']
-    client = DataAPIClient(get_api_endpoint_from_stage(stage), arguments['<api_token>'])
-    framework_slug = arguments['<framework_slug>']
+    data_api_client = DataAPIClient(get_api_endpoint_from_stage(stage), arguments['<api_token>'])
+    framework = data_api_client.get_framework(arguments['<framework_slug>'])["frameworks"]
     document_directory = arguments['<documents_directory>']
     dry_run = arguments['--dry-run']
+    dm_notify_client = arguments.get("--notify-key") and DMNotifyClient(arguments["--notify-key"], logger=logger)
```
TheDoubleK: Is it worth making the dm_notify_client in the `if dry_run... else`, along with bucket initialisation?  That way no chance of sending emails if it's a dry run. (Not that there is anyway, but might make it a bit more obvious).

dmscripts/upload_counterpart_agreements.py
```diff
@@ -1,26 +1,49 @@
...
 
-def upload_counterpart_file(bucket, framework_slug, file_path, dry_run, client):
     supplier_id = get_supplier_id_from_framework_file_path(file_path)
-    supplier_framework = client.get_supplier_framework_info(supplier_id, framework_slug)
+    supplier_framework = data_api_client.get_supplier_framework_info(supplier_id, framework["slug"])
     supplier_framework = supplier_framework['frameworkInterest']
     supplier_name = supplier_framework['declaration']['nameOfOrganisation']
     download_filename = generate_download_filename(supplier_id, COUNTERPART_FILENAME, supplier_name)
 
+    notify_emails = dm_notify_client and frozenset(chain(
```
TheDoubleK: `email_addresses_to_notify`?

dmscripts/upload_counterpart_agreements.py
```diff
@@ -47,10 +70,23 @@ def upload_counterpart_file(bucket, framework_slug, file_path, dry_run, client):
             logger.info("[Dry-run] countersignedAgreementPath='{}' for agreement ID {}".format(
                 upload_path, supplier_framework['agreementId'])
             )
+
+        for notify_email in (notify_emails or ()):
+            if not dry_run:
```
TheDoubleK: If we moved the notify client generation inside the `if dry_run` bit we wouldn't need this, but would probably want to leave it here for clarity anyway I guess.

tests/test_upload_counterpart_agreements.py
```diff
@@ -33,27 +45,174 @@ def test_upload_counterpart_file_uploads_and_calls_api_if_not_dry_run(getuser):
...
+        },
+        {
+            "id": 111321,
+            "emailAddress": "user.111321@example.com",
+            "supplierId": 123456,
+            "active": True,
+        },
+    ))
+    dm_notify_client = mock.Mock()
+    dm_notify_client.send_email.side_effect = AssertionError("This shouldn't be called")
```
TheDoubleK: Oh I like this more than `assert .called is False` at the end.

scripts/upload-counterpart-agreements.py
risicle: I was in two minds over this - on one hand, what you say - on the other hand I'd like to somehow give it a chance for the API key to fail if it's going to. But I don't know if the notify client does any sort of handshaking on client creation to verify that - I imagine it doesn't. Maybe it can verify the key's formatting...?

tests/test_upload_counterpart_agreements.py
risicle: Main niceness is it flagposts exactly where it gets called when it fails.

dmscripts/upload_counterpart_agreements.py
risicle: Yup.

scripts/upload-counterpart-agreements.py
risicle: Oh, if we didn't create a notify client to pass to the inner function, it wouldn't bother calculating the correct email addresses and you couldn't see from the logs what the run would do.

tests/test_upload_counterpart_agreements.py
risicle: I have an intention to try and bring `assert_args_and_raise` and `assert_args_and_return` (as used in e.g. https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/9c5e42c6a227dd4bd287eacc231e88e93fdc2439/tests/app/main/test_frameworks.py#L49) in to dmutils for further use elsewhere.

risicle: And to make them support e.g. `mock.ANY`

scripts/upload-counterpart-agreements.py
TheDoubleK: Ah yep, good point.

TheDoubleK:
Had a new thought about this - I think we should count all errors (IO/API/Notify/whatever) and return the error count from the script.

That way Jenkins can make the job as failed if there were any errors, we'll see it red on the board and can go in to check what went wrong.

risicle:
Iiiiiiiiinteresting...

TheDoubleK:
Think this is good to go now :+1:

PR #96, opened 2017-03-23
Script to generate fake (G9) services to enable testing and iteration of search and summaries.

samuelhwilliams:
## Summary
A new script which can be used to generate fake (G9) services from JSON Validation schemas.

## Changes
* Add a minimalist JsonSchemaDataFaker class and JsonSchemaG9ServiceFaker subclass which accept a given JSON Validation Schema and reverse-engineers compatible data. These classes are implemented to the bare-minimum level required for generating services from G9 schemas, and are not guaranteed to generate data from any given generic JSON Validation Schema.
* Add the JSON Validation Schemas for cloud-hosting, cloud-software, cloud-support.
* Add a JSON Validation Schema for G9 supplier declarations.
* Add a '--ignore-documents' parameter to the `scripts/make-fake-g-cloud-services-live.py`, to bypass S3/document operations.
* Fix logging import in `scripts/mark-definite-framework-results.py`
*  Adds `scripts/oneoff/generate-g-cloud-services.py` to generate fake G9 services
   * This script changes the framework state (open -> standstill -> live).

## To Test
* (Optional) Wipe your dev DB and re-generate: https://github.gds/pages/gds/digitalmarketplace-manual/data-migrations.html#importing-data-for-developers
* Add g-cloud-9 framework (see previous e-mail from Kev Keenoy)
* Pull down the buyer-frontend dependency PR and get it running with the frameworks dependency PR.
* Delete your local g-cloud index:
`curl -XDELETE -H 'Authorization: Bearer myToken' http://localhost:5001/g-cloud`
* Run the script: `./scripts/oneoff/generate-g-cloud-services.py <# of services to generate>`
   * Recommend generating 100+ services (by default, randomly distributes across all three lots).
* Visit `localhost/g-cloud/search` and browse services

## Dependencies
~~https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/463~~ (MERGED):
* Requires an update to the buyer-frontend to support G9 services, which have some new/changed question keys

~~https://github.com/alphagov/digitalmarketplace-frameworks/pull/410~~ (MERGED):
* Requires a new G9 display_service manifest.

### DIFF
--- a/dmscripts/helpers/service_faker_helpers.py
+++ b/dmscripts/helpers/service_faker_helpers.py
@@ -1,0 +1,71 @@
+# coding=utf-8
+
+import random
+
+
+def eligible_g9_declaration_base():
+    return {
+        "unfairCompetition": True, "offerServicesYourselves": True,
+        "fullAccountability": True, "termsOfParticipation": True, "termsAndConditions": True,
+        "canProvideFromDayOne": True, "10WorkingDays": True, "MI": True, "conspiracy": False,
+        "corruptionBribery": False, "fraudAndTheft": False, "terrorism": False, "organisedCrime": False,
+        "taxEvasion": False, "environmentalSocialLabourLaw": False, "bankrupt": False,
+        "graveProfessionalMisconduct": False, "distortingCompetition": False, "conflictOfInterest": False,
+        "distortedCompetition": False, "significantOrPersistentDeficiencies": False, "seriousMisrepresentation": False,
+        "witheldSupportingDocuments": False, "influencedContractingAuthority": False, "confidentialInformation": False,
+        "misleadingInformation": False, "mitigatingFactors": "Money is no object", "unspentTaxConvictions": False,
+        "GAAR": False, "mitigatingFactors2": "Project favourably entertained by auditors",
+        "environmentallyFriendly": True, "equalityAndDiversity": True,
+        "employersInsurance": u"Not applicable - your organisation does not need employer’s liability insurance because"
+                              u" your organisation employs only the owner or close family members. ",
+        "publishContracts": True, "readUnderstoodGuidance": True, "understandTool": True,
+        "understandHowToAskQuestions": True, "accurateInformation": True, "informationChanges": True,
+        "accuratelyDescribed": True, "proofOfClaims": True,
+        "nameOfOrganisation": "Mr Malachi Mulligan. Fertiliser and Incubator.",
+        "tradingNames": "Omphalos dutiful yeoman services", "registeredAddressBuilding": "Omphalos",
+        "registeredAddressTown": "Lambay Island", "registeredAddressPostcode": "N/A", "firstRegistered": "5/6/1904",
+        "currentRegisteredCountry": u"Éire", "companyRegistrationNumber": "00000014", "dunsNumber": "987654321",
+        "registeredVATNumber": "123456789", "establishedInTheUK": False, "appropriateTradeRegisters": True,
+        "appropriateTradeRegistersNumber": "242#353", "licenceOrMemberRequired": "none of the above",
+        "licenceOrMemberRequiredDetails": "",
+        "subcontracting": ["yourself without the use of third parties (subcontractors)", ], "organisationSize": "small",
+        "tradingStatus": "other (please specify)", "tradingStatusOther": "Proposed", "primaryContact": "B. Mulligan",
+        "primaryContactEmail": "buck@example.com", "contactNameContractNotice": "Malachi Mulligan",
+        "contactEmailContractNotice": "malachi@example.com", "servicesHaveOrSupport": True,
+        "servicesDoNotInclude": True, "payForWhatUse": True, "helpBuyersComplyTechnologyCodesOfPractice": True,
+        "status": "complete"
+    }
+
+
+# Just some word lists to generate random cloudy service names.
+SERVICE_QUESTIONABLE_ADJECTIVES = ['shiny', 'fractal', 'golden', 'sublime', 'self-replicating', 'superlative',
+                                   'impenetrable', 'fully-managed', 'wonderful', 'cost-effective', 'secure',
+                                   'certified', 'accredited', 'cybernetic', 'quantum', 'cutting-edge', 'AI-powered']
+SERVICE_ADJECTIVES = ['new', 'modern', 'high', 'advanced', 'medical', 'current', 'appropriate', 'military', 'digital',
+                      'nuclear', 'available', 'latest', 'western', 'foreign', 'industrial', 'agricultural',
+                      'educational', 'sophisticated', 'electronic', 'improved', 'wireless', 'innovative', 'assistive',
+                      'instructional', 'superior', 'artisanal', 'conventional', 'complex', 'optical', 'intensive',
+                      'scientific', 'proprietary', 'underlying', 'mechanical', 'newer', 'contemporary', 'integrated',
+                      'indigenous', 'laser', 'efficient', 'genetic', 'mobile', 'intermediate', 'alternative',
+                      'solar', 'interactive', 'clean', 'mature', 'primitive', 'modem', 'oriented', 'electrical',
+                      'developed', 'promising', 'ceramic', 'multimedia', 'containerised']
+SERVICE_TYPES = ['text', 'online', 'e-service', 'insight', 'intelligence', 'mainframe', 'marketing', 'quality',
+                 'digital', 'dynamic', 'engagement', 'discovery', 'solutions', 'strategy', 'transformation',
+                 'evaluation', 'advisory', 'physical', 'threat', 'fraud', 'improvement', 'detection', 'waterfall',
+                 'managed', 'delivery']
+SERVICE_NOUNS = ['security', 'workspace', 'portal', 'platform', 'solution', 'module', 'form', 'strategy', 'development',
+                 'search', 'engine', 'manager', 'engineer', 'developer', 'stack', 'environment', 'cloud', 'store',
+                 'database', 'healthcheck', 'optimiser', 'doctor', 'intelligence', 'support', 'analyst', 'tools',
+                 'course', 'omni-channel hub', 'assistant', 'gateway', 'processor', 'container', 'neural net', 'node',
+                 'dashboard', 'visualiser', 'system', 'network', 'suite']
+
+
+def gen_gcloud_name():
+    service_name = '{} {} {}'.format(random.choice(SERVICE_ADJECTIVES),
+                                     random.choice(SERVICE_TYPES),
+                                     random.choice(SERVICE_NOUNS))
+    if random.randint(1, 5) == 1:
+        service_name = '{} {}'.format(random.choice(SERVICE_QUESTIONABLE_ADJECTIVES),
+                                      service_name)
+
+    return service_name.title()
--- a/dmscripts/json_schema_service_faker.py
+++ b/dmscripts/json_schema_service_faker.py
@@ -1,0 +1,183 @@
+import copy
+import json
+import logging
+import lorem
+import random
+import re
+import xeger
+from jsonschema import validate, ValidationError
+
+logger = logging.configure_logger()
+
+
+class JsonSchemaDataFaker(object):
+
+    """
+    JSON Schema Data Faker takes a generated JSON schema and returns an object containing fake data that matches
+    the constraints put forward by that schema.
+    Only a subset of the JSON Schema syntax - enough to create fake G9 services - has been implemented.
+    """
+
+    def __init__(self, max_recalcs=100):
+        """Instantiate the class.
+        :param limit_calculations: Set the upper bounds on how many recalculations to run when generating data
+            with max/min constraints, as this data is currently generated dumbly and then constrained afterwards."""
+        self._max_recalcs = max_recalcs
+
+    def _null(self, schema):
+        return None
+
+    def _boolean(self, schema):
+        return random.choice([True, False])
+
+    def _enum(self, schema):
+        return random.choice(schema['enum'])
+
+    def _array(self, schema):
+        """Handles JSON Schema 'array' type: essentially a Python list"""
+        min_items = schema.get('minItems', 1)
+        max_items = schema.get('maxItems', 10)
+        num_items = random.randint(min_items, max_items)
+
+        if schema.get('uniqueItems', False):
+            items = set()
+            add_to_items = items.add
+        else:
+            items = []
+            add_to_items = items.append
+
+        attempts = 0
+        while len(items) < num_items:
+            len_items = len(items)
+
+            add_to_items(self._parse_definition(schema['items']))
+
+            # Prevent infinite loops
+            if len_items == len(items):
+                attempts += 1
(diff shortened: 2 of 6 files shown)


scripts/oneoff/generate-g-cloud-services.py
```diff
@@ -0,0 +1,163 @@
...
+        sys.exit(1)
+
+    data_api_url = get_api_endpoint_from_stage(args.env, 'api')
+    data_api_client = DataAPIClient(data_api_url, args.data_api_token)
+
+    search_api_url = get_api_endpoint_from_stage(args.env, 'search-api')
+    search_api_client = SearchAPIClient(search_api_url, args.search_api_token)
+
+    services_generated = 0
+    jsfd = JsonSchemaServiceFaker()
```
samuelhwilliams: Oh. Yes. That probably is a bit confusing. Think I called it Json Schema Fake Data for a while... Will fix it.

dmscripts/json_schema_service_faker.py
```diff
@@ -0,0 +1,183 @@
...
+            attempts += 1
+
+        logger.warning("Unable to generate within length constraints ({} < len < {}) for pattern: '{}'"
+                       .format(min_length, max_length, pattern))
+
+        return random_string
+
+    def __uri(self, min_length, max_length):
+        """Just returns a single URL; g-9 services' compulsory fields only use URI for uploaded documents."""
+        return 'https://www.digitalmarketplace.service.gov.uk/uploaded_document.pdf'
```
samuelhwilliams: It does... But also don't really expect the class to ever get re-used, so it's a good way to prompt a proper review before trying to jerry-rig it to do something else.

But I'm just trying to justify my bad decision here...

schemas/g-cloud-9.declaration.json
```diff
@@ -0,0 +1,76 @@
+{
```
samuelhwilliams: Yeah. :(

dmscripts/json_schema_service_faker.py
```diff
@@ -0,0 +1,183 @@
+#!/usr/bin/env python
```
samuelhwilliams: Good spot. I always drop this thing in without thinking about it.

dmscripts/json_schema_service_faker.py
samuelhwilliams: Removed the bad juju. 👍

schemas/g-cloud-9.declaration.json
TheDoubleK: What do you mean when you say "we already did it for all the other repos"?

Only API has schemas in it, surely?

scripts/oneoff/generate-g-cloud-services.py
```diff
@@ -0,0 +1,163 @@
...
+    # a) Runs the `mark-definite-framework-results.py` script to mark x suppliers as on the new framework.
+    subprocess.call(['python', 'scripts/mark-definite-framework-results.py', '--reassess-failed-sf', '--updated-by',
+                     identity, args.env, args.data_api_token, args.new_slug, filepath_declaration_validator])
+
+    # b) Runs the `make-g-cloud-live.py` script (ignoring all documents)
+    subprocess.call(['python', 'scripts/make-g-cloud-live.py', '--ignore-documents', args.new_slug, args.env,
+                     args.data_api_token, 'null', 'null'])
+
+    # c) Mark framework agreements as returned for all selected suppliers
+    # TODO: Maybe do this. Not doing it breaks some supplier summary pages, but shouldn't affect browsing the catalogue.
```
TheDoubleK: "breaks" in what sense?  All pages should work just fine whether or not an agreement has been returned.

scripts/oneoff/generate-g-cloud-services.py
```diff
@@ -0,0 +1,163 @@
...
+    identity = 'generate-g-cloud-services script ({})'.format(getpass.getuser())
+    filepath_declaration_validator = 'schemas/{}.declaration.json'.format(args.new_slug)
+
+    # 1) Set a given (new) framework to open
+    data_api_client._post_with_updated_by(url='{}/frameworks/{}'.format(data_api_url, args.new_slug),
+                                          data={"frameworks": {"status": "open"}},
+                                          user=identity)
+
+    # 2) Find suppliers with services on a given (old) framework (e.g. g-cloud-8)
+    suppliers = data_api_client.find_framework_suppliers(framework_slug=args.old_slug)['supplierFrameworks']
```
TheDoubleK: I worry slightly about having such a wide choice of suppliers for this - can we restrict the suppliers with fake applications somehow?  At least check with Andrew what he has set up in terms of test data and definitely don't mess with those accounts. 

But the way this is the chances are almost all suppliers will probably end up with a fake application - we could further restrict to only suppliers with a supplier ID beginning with a '5' (signed up for G7) , '7' (signed up for DOS, G8 or DOS2) or '9' (signed up pre-G7) maybe?

scripts/oneoff/generate-g-cloud-services.py
```diff
@@ -0,0 +1,163 @@
...
+                     identity, args.env, args.data_api_token, args.new_slug, filepath_declaration_validator])
+
+    # b) Runs the `make-g-cloud-live.py` script (ignoring all documents)
+    subprocess.call(['python', 'scripts/make-g-cloud-live.py', '--ignore-documents', args.new_slug, args.env,
+                     args.data_api_token, 'null', 'null'])
+
+    # c) Mark framework agreements as returned for all selected suppliers
+    # TODO: Maybe do this. Not doing it breaks some supplier summary pages, but shouldn't affect browsing the catalogue.
+
+    # 5) Set a given (new) framework to live
```
TheDoubleK: Is this because framework needs to be live for the indexing job to be able to retrieve services? I don't think it should need to be?  I'm sure we have created indexes before services go live in the past.

If that works then I don't think flipping the framework status should be part of this script.

schemas/g-cloud-9.declaration.json
galund: Yeah you're right. I guess I was thinking of the framework content getting linked in via bower, that's different though. Just the API does this so far.

scripts/oneoff/generate-g-cloud-services.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python
+# coding=utf-8
+
+"""
+Script to generate, submit, approve and index fake services from the validation JSON schemas
+Only tested to work against g-cloud-9
+
+Pre-requisites:
```
samuelhwilliams: Thought about that and wondered if you'd bring it up. In some respects it seems nicer, and it does prevent having duplicated schemas... So I guess let's do it.

dmscripts/helpers/service_faker_helpers.py
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python
```
samuelhwilliams: Yep.

dmscripts/helpers/service_faker_helpers.py
```diff
@@ -0,0 +1,72 @@
...
+                      'nuclear', 'available', 'latest', 'western', 'foreign', 'industrial', 'agricultural',
+                      'educational', 'sophisticated', 'electronic', 'improved', 'wireless', 'innovative', 'assistive',
+                      'instructional', 'superior', 'artisanal', 'conventional', 'complex', 'optical', 'intensive',
+                      'scientific', 'proprietary', 'underlying', 'mechanical', 'newer', 'contemporary', 'integrated',
+                      'indigenous', 'laser', 'efficient', 'genetic', 'mobile', 'intermediate', 'alternative',
+                      'solar', 'interactive', 'clean', 'mature', 'primitive', 'modem', 'oriented', 'electrical',
+                      'developed', 'promising', 'ceramic', 'multimedia', 'containerised']
+SERVICE_TYPES = ['text', 'online', 'e-service', 'insight', 'intelligence', 'mainframe', 'marketing', 'quality',
+                 'digital', 'dynamic', 'engagement', 'discovery', 'solutions', 'strategy', 'transformation',
+                 'evaluation', 'advisory', 'physical', 'threat', 'fraud', 'improvement', 'detection', 'agile',
```
samuelhwilliams: Ok.

scripts/oneoff/generate-g-cloud-services.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python
+# coding=utf-8
+
```
samuelhwilliams: Ok.

scripts/oneoff/generate-g-cloud-services.py
```diff
@@ -0,0 +1,163 @@
...
+    # 1) Set a given (new) framework to open
+    data_api_client._post_with_updated_by(url='{}/frameworks/{}'.format(data_api_url, args.new_slug),
+                                          data={"frameworks": {"status": "open"}},
+                                          user=identity)
+
+    # 2) Find suppliers with services on a given (old) framework (e.g. g-cloud-8)
+    suppliers = data_api_client.find_framework_suppliers(framework_slug=args.old_slug)['supplierFrameworks']
+    suppliers_prepared = set()
+
+    # 3) Submit y draft services through randomly selected suppliers.
```
samuelhwilliams: x/y - some number of.

scripts/oneoff/generate-g-cloud-services.py
```diff
@@ -0,0 +1,163 @@
...
+    while services_generated < args.quantity:
+        random_lot = random.choice(args.lots)
+        random_supplier = random.choice(suppliers)
+
+        supplier_id = random_supplier['supplierId']
+        if supplier_id not in suppliers_prepared:
```
samuelhwilliams: Yeah it should have added it at the end of the statement, good catch.

scripts/oneoff/generate-g-cloud-services.py
samuelhwilliams: You can only view the service summary page if the framework is live or expired.

scripts/make-g-cloud-live.py
```diff
@@ -11,6 +11,7 @@
     and updating document URLs in the migrated version of the services
 Usage:
     scripts/make-g-cloud-live.py <framework_slug> <stage> <api_token> <draft_bucket> <documents_bucket> [--dry-run]
+    [--ignore-documents]
```
samuelhwilliams: It does. Moved to use that script instead.

scripts/oneoff/generate-g-cloud-services.py
```diff
@@ -0,0 +1,163 @@
...
+2) Find suppliers with services on a given (old) framework (e.g. g-cloud-8)
+    a) For each supplier, register interest in a given (new) framework
+    b) Submit and mark as complete a valid supplier declaration for the new framework
+3) Submit y draft services through randomly selected suppliers.
+4) Set a given (new) framework to standstill
+    a) Runs the `mark-definite-framework-results.py` script to mark x suppliers as on the new framework.
+    b) Runs the `make-g-cloud-live.py` script (ignoring all documents)
+5) Set a given (new) framework to live
+6) Runs the `index-services.py` script to index the new services
+"""
```
samuelhwilliams: Ok.

scripts/oneoff/generate-g-cloud-services.py
samuelhwilliams: You can restrict the supplier pool by using the --old-slug parameter, currently defaulting to g-cloud-8. Only suppliers with valid services against that framework will be returned.

I could change the default to g-cloud-7?

TheDoubleK:
Where does the G9 declaration schema come from? And why do we need it if we have a hard-coded valid declaration to post up anyway?

TheDoubleK:
Ah OK, sorry - it's the schema used to assess pass/fail/discretionary results, I was thinking about validation schemas.

samuelhwilliams:
mark-definite-framework-results needs a declaration validator; in theory I could just use an almost empty one, but it wasn't hard to adapt from the dos declaration validator.

TheDoubleK:
In fact, this repo is the *only* place that the declaration schemas live - they aren't used in any of the apps, only here for the pass/fail script.

So there's no duplication for the declaration schemas (@ris is working on auto-generating the declaration schemas from the content now though).

For the service validation schemas I reckon avoid copying them in here - better to point the script to a checked-out API with the latest schemas I reckon.

samuelhwilliams:
Have already removed the service schemas, yep.

scripts/oneoff/generate-g-cloud-services.py
TheDoubleK: Yep, but I'm not convinced the script needs to flip the framework status - the script is about setting up and indexing the data, not changing the state of things in environments.

scripts/oneoff/generate-fake-g-cloud-services.py
```diff
@@ -0,0 +1,197 @@
...
+    # 4) Set a given (new) framework to standstill
+    data_api_client._post_with_updated_by(url='{}/frameworks/{}'.format(data_api_url, args.new_slug),
+                                          data={"frameworks": {"status": "standstill"}},
+                                          user=identity)
+
+    # a) Runs the `mark-definite-framework-results.py` script to mark x suppliers as on the new framework.
+    subprocess.call(['python', 'scripts/mark-definite-framework-results.py', '--reassess-failed-sf', '--updated-by',
+                     identity, args.env, args.data_api_token, args.new_slug, filepath_declaration_validator])
+
+    # b) Runs the `make-dos-live.py` script (which simply migrats draft services to 'real' services)
```
TheDoubleK: Typo `migrats` but I can live with it.

samuelhwilliams: I don't think I can live with that. Let me start all over again!

scripts/oneoff/generate-g-cloud-services.py
TheDoubleK: At the very least I think check with Andrew to get a "blacklist" list of supplier IDs that should be excluded from having any stuff added for them.  They could potentially be on any old framework I think.

scripts/oneoff/generate-g-cloud-services.py
galund: Is there a scenario where we would want to generate the data, and then not be able to see it? Arguably the more of the process that can be automated, the more of it we should automate.

TheDoubleK: To be honest I think it's a bit weird having a script flipping framework statuses at all.  This one needs to do it so we can submit drafts and publish services.  But the state of frameworks should generally depend on what we as a team need them to be for testing things (and in some cases what the functional tests need them to be for testing things).

For example, if this was any previous framework application window we would have functional tests running and submitting services in each lot every 30 minutes.  Flipping the framework status to 'live' would then cause those tests to all fail.  We would need to agree as a team what status the live environment should currently be in - just because a script has run isn't really a reason for moving framework status on.

Thinking about it, there's an argument that the script should probably re-set the framework status to whatever it was to begin with.

TheDoubleK:
Think this can go with or without the flip to "live" at the end :+1:

PR #97, opened 2017-03-28
Add `frameworkSlug` and `specialistRole` to final csv

pcraig3:
For each brief, we know how many people applied to it, as well as how many people applied successfully/unsuccessfully.

Ash also wants to know how many people _could have_ applied to each brief, so we're putting some more stuff in the information about the applied-to briefs so that he can use a lookup table.


idavidmcdonald:
👍

PR #98, opened 2017-03-28
[#141943157]mark_definite_framework_results: make declaration "status" handling "special case"

risicle:
Story https://www.pivotaltracker.com/story/show/141943157

It was going to be ugly to get the frameworks to automagically generate
an assessment schema that also included the `status` criteria, and what's
more we already do "special case" handling of the service "status" field.
also change included dos2 schema, no longer including `status`, to match
this.


TheDoubleK:
Seems reasonable to me.
:+1:

risicle:
Famous last words

PR #101, opened 2017-04-04
Return error if errors generating PDFs and unobfuscate dry-run emails

TheDoubleK:
For this story: https://www.pivotaltracker.com/story/show/128842967

We need the Jenkins job to know if the PDF generation step hasn't worked as expected, so the script needs to return an error code.

Its also useful in dry-run mode to see what on earth the script would have done, so I'm unobfuscating email addresses in dry-run output.

### DIFF
--- a/dmscripts/generate_framework_agreement_signature_pages.py
+++ b/dmscripts/generate_framework_agreement_signature_pages.py
@@ -56,6 +56,7 @@ def render_html_for_suppliers_awaiting_countersignature(rows, framework_kwargs, 
 def render_pdf_for_each_html_page(html_pages, html_dir, pdf_dir):
     html_dir = os.path.abspath(html_dir)
     pdf_dir = os.path.abspath(pdf_dir)
+    ok = True
     if not os.path.exists(pdf_dir):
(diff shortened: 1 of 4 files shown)


dmscripts/generate_framework_agreement_signature_pages.py
```diff
@@ -65,3 +66,5 @@ def render_pdf_for_each_html_page(html_pages, html_dir, pdf_dir):
         exit_code = subprocess.call(['wkhtmltopdf', 'file://{}'.format(html_path), pdf_path])
         if exit_code > 0:
             logger.error("ERROR {} on {}".format(exit_code, html_page))
+            ok = False
```
galund: This is fine, although unless failure here is really common, it might just be simpler to throw. If rendering one of the docs didn't work, it's not completely clear to me that trying to do the rest is at all likely to succeed?

TheDoubleK: Yeah I'm not sure.  Feels like it's worth trying the rest and then it'll be more obvious looking at the output whether it's an issue with just one file or a general issue with everything.

PR #104, opened 2017-04-05
[#141943669]add tests for export_framework_results_reasons

risicle:
Story https://www.pivotaltracker.com/story/show/141943669

Firstly the test data for `mark_definite_framework_results` was generalized into `assertion_helpers.py` so that it could be reused for new `export_framework_results_reasons` tests. Retrofitted some more data into it that would be used by these new tests. Created some mixin classes that can be used to mess up the mocked values of `onFramework` etc when that is required.

Also had to perform a small refactor of `export_framework_results_reasons` itself. Paraphrased from the commit message:

```
minor refactor using content loader only at the top level

this is used to supply a dict of {question_id: question_number} which
is really all this script needs. extracting this in a separate function 
(get_questions_numbers_from_framework) allows it to be neatly
mocked out in tests
```

This was required because the test data didn't use a "real" framework (and we wouldn't want to make it have to).

I guess what should be checked here is that the "expected results" should be what we want. Mostly only relevant for the first parametrization of the first test (the only non-degenerate case).


dmscripts/export_framework_results_reasons.py
```diff
@@ -194,6 +161,14 @@ def create_row(self, record):
             contact_details(record)
 
 
+def get_questions_numbers_from_framework(framework_slug, content_loader):
```
TheDoubleK: :scream: I tried to look at this so many times up to now, sorry it's taken so long.

Trying to run this it breaks here for me, not 100% sure why without digging in a bit more:

```
Traceback (most recent call last):
  File "./scripts/export-framework-results-reasons.py", line 57, in <module>
    supplier_ids
  File "./dmscripts/export_framework_results_reasons.py", line 184, in export_suppliers
    questions_numbers = get_questions_numbers_from_framework(framework_slug, content_loader)
  File "./dmscripts/export_framework_results_reasons.py", line 168, in get_questions_numbers_from_framework
    for question in chain.from_iterable(content_loader.get_manifest(framework_slug, 'declaration'))
  File "./dmscripts/export_framework_results_reasons.py", line 167, in <dictcomp>
    question.id: question.number
  File "/Users/kevinkeenoy/gdsworkspace/digitalmarketplace-scripts/venv/lib/python2.7/site-packages/dmcontent/content_loader.py", line 176, in __getitem__
    return getattr(self, key)
TypeError: getattr(): attribute name must be string
```

TheDoubleK: Same error running Python 3.
```
Traceback (most recent call last):
  File "./scripts/export-framework-results-reasons.py", line 57, in <module>
    supplier_ids
  File "./dmscripts/export_framework_results_reasons.py", line 185, in export_suppliers
    questions_numbers = get_questions_numbers_from_framework(framework_slug, content_loader)
  File "./dmscripts/export_framework_results_reasons.py", line 166, in get_questions_numbers_from_framework
    print(list(chain.from_iterable(content_loader.get_manifest(framework_slug, 'declaration'))))
  File "/Users/kevinkeenoy/gdsworkspace/digitalmarketplace-scripts/venv3/lib/python3.4/site-packages/dmcontent/content_loader.py", line 176, in __getitem__
    return getattr(self, key)
TypeError: getattr(): attribute name must be string
```

risicle: Oh fanny I know what this is - it's usually due to trying to use the non-mutating update `dict(basedict, **updatedict)` trick when it turns out that one of `updatedict`'s keys isn't a string. Not entirely sure where in this case but I'll have to find it.

risicle: Not sure why you're getting it and I wasn't. Content loader version?

risicle: Run it under `py.test --pdb` and find out what the value of `key` is?

TheDoubleK: Not running the tests - they're OK.  It's running the script itself for G9 that breaks.

risicle: arg. Run with `python -m pdb the_script.py` ?

risicle: Oh! I think I know what it's doing... talk to you in slack

risicle: Another wtf here is that I didn't remove the original load_manifest call after moving it

dmscripts/export_framework_results_reasons.py
```diff
@@ -194,6 +161,13 @@ def create_row(self, record):
...
+    return {
```
TheDoubleK: Yep this needs  `content_loader.load_manifest(framework_slug, 'declaration', 'declaration')` before we do the `get_manifest`, but then we get `AttributeError: 'ContentManifest' object has no attribute 'questions'`

risicle:
Ok I've actually *tried it* this time so it should work.

risicle:
Need mock frameworks so bad.

TheDoubleK:
So the output is right, but in the old version the "failed" questions were output in increasing numerical order, e.g. `Q12 - conspiracy,Q13 - corruptionBribery,Q14 - fraudAndTheft,Q15 - terrorism,Q16 - organisedCrime` but running this version gives an apparently random ordering of questions, e.g. `Q16 - organisedCrime,Q14 - fraudAndTheft,Q12 - conspiracy,Q13 - corruptionBribery,Q15 - terrorism`.

It's not a deal-breaker, but it feels like it was nicer before with them in order.

TheDoubleK:
Oh you fixed it already :+1:

tests/assessment_helpers.py
```diff
@@ -0,0 +1,389 @@
...
+                        "irrelevantQuestion": "Irrelevant answer",
+                    },
+                },
+                2345: {
+                    "onFramework": False,
+                    "declaration": {
+                        "status": "complete",
+                        "shouldBeTrueLax": True,
+                        "shouldMatchPatternLax": "Good pattern",
+                        "shouldBeFalseStrict": None,  # <- subtle but important test here
```
risicle: unless the content loader misbehaves (or the view using the content loader uses it wrong)

tests/assessment_helpers.py
```diff
@@ -0,0 +1,389 @@
...
+                        "shouldBeTrueLax": True,
+                        "shouldMatchPatternLax": "Good   Pattern",
+                        "shouldMatchPatternStrict": "H.E. L.  Y",
+                        "irrelevantStupidQuestion": "Irrelevant stupid answer",
+                    },
+                },
+                4567: {
+                    "onFramework": False,
+                    "declaration": {
+                        "status": "bored-gave-up",
```
risicle: yeah i think this was originally that but i think i noticed that nothing in this area knows/cares about `started` so i thought i'd throw some chaos into the mix

tests/test_export_framework_results_reasons.py
```diff
@@ -0,0 +1,370 @@
...
+from mock import Mock, patch
+
+from dmscripts.export_framework_results_reasons import export_suppliers
+
+from .assessment_helpers import BaseAssessmentTest, BaseAssessmentOnFrameworksAsThoughNoBaselineTestMixin
+
+
+class _BaseExportFrameworkResultsReasonsTest(BaseAssessmentTest):
+    def _mock_get_questions_numbers_from_framework_impl(self, framework_slug, content_loader):
+        assert content_loader is self.mock_content_loader
```
risicle: :thinking: mmmmmmmmmm

risicle: Checks the code under test is sending the args in the right order. Yes, I did screw this up during development.

It's also a symmetry thing, routinely check all args are what you expect them to be - the "pre-canned" reply the code should expect *given* these specific args.

TheDoubleK:
Can we get this rebased and merged now?

PR #105, opened 2017-04-10
Notify suppliers when framework closes

TheDoubleK:
Script for this story: https://www.pivotaltracker.com/story/show/141695173

It reuses the pattern from the existing `notify-successful-suppliers` script.  This one needs a slightly less complicated set of personalisation data so the new `AppliedToFrameworkSupplierContextForNotify` is pretty simple.  I discussed with Paul whether it should instead use the `modeltrawler` pattern but we decided that wasn't a good way to go right now (time constraints as much as anything else, but this also isn't quite the typical use-case for modeltrawler).

I hard-coded the Notify template IDs in the script as (1) these shouldn't change from framework to framework and (2) it feels less error-prone than people having to pass multiple template IDs as command line params.

The fetch of user data here is relying on the currently-broken `export_users` method on the API.  This is a shame - it wasn't broken when I wrote this - but hopefully we'll find some way to fix it before tomorrow rather than having to move all the logic for "applied/didn't apply" out to this script.

### DIFF
--- a/dmscripts/helpers/supplier_data_helpers.py
+++ b/dmscripts/helpers/supplier_data_helpers.py
@@ -111,3 +111,37 @@ class SuccessfulSupplierContextForNotify(SupplierFrameworkData):
         }
         personalisation.update(lot_dict)
         return {user['email address']: personalisation}
+
+
+class AppliedToFrameworkSupplierContextForNotify(SupplierFrameworkData):
+    """Get the personalisation/ context for 'You application result - if successful email'"""
+
+    def __init__(self, client, target_framework_slug, successful_notification_date):
+        """Get the target framework to operate on and list the lots.
+
+        :param client: Instantiated api client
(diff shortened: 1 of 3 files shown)


dmscripts/helpers/supplier_data_helpers.py
```diff
@@ -111,3 +111,37 @@ def get_user_personalisation(self, user, supplier_framework):
...
+
+        :param client: Instantiated api client
+        :param target_framework_slug: Framework to fetch data for
+        """
+        super(AppliedToFrameworkSupplierContextForNotify, self).__init__(client, target_framework_slug)
+
+        self.successful_notification_date = successful_notification_date
+        self.framework = client.get_framework(self.target_framework_slug)['frameworks']
+
+    def get_users_peronalisations(self):
```
pcraig3: speliling error

pcraig3:
Looks okay to me 👍

PR #106, opened 2017-04-11
Implement script to send emails for each DOS lot using Mailchimp client

CrystalPea:
Trello card: https://trello.com/c/RDQcaDGo/252-write-and-implement-the-script-to-send-the-emails-for-each-lot

The script sends emails with new briefs to suppliers.
The script will be run by a jenkins job every weekday.
The script will look for briefs published on a previous day or three previous days, depending on instructions from the jenkins job. E.g. if email is sent on Monday, the script will search for briefs published on Fri, Sat, Sun. If the email is sent between Tue-Fri, the script will look for briefs published on a previous day to sending.

For the moment, we haven't touched he content of the email and it's setup with some dummy text. There is a separate story to implement this:
https://trello.com/c/rajSbay6/251-create-the-email-template-for-supplier-notifications

### DIFF
--- a/dmscripts/send_dos_opportunities_email.py
+++ b/dmscripts/send_dos_opportunities_email.py
@@ -15,20 +15,23 @@ from dmscripts.helpers.logging_helpers import logging
 logger = logging_helpers.configure_logger({'dmapiclient': logging.INFO})
 
 
-lots = {
-    "digital-specialists": {
+lots = [
+    {
+        "lot_slug": "digital-specialists",
         "lot_name": "Digital specialists",
         "list_id": "096e52cebb"
     },
-    "digital-outcomes": {
+    {
+        "lot_slug": "digital-outcomes",
         "lot_name": "Digital outcomes",
         "list_id": "096e52cebb"
     },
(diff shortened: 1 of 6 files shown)


dmscripts/send_dos_opportunities_email.py
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+
+def get_html_content():
+    # function not yet implemented
+    html_content = render_html("email_templates/dos_opportunities.html", data={})
+    return {"html": html_content}
+
+
+def create_campaign(mailchimp_client, campaign_data):
```
benvand: Just a note: `create_campaign`, `set_campaign_content` and `send_campaign` will implicitly return `None` if no return line is specified. This can be used in a truthey/ falsey check as None is Falsey.
_However_ 'explicit is better than implicit' and I think in this case explicitly defining the returns makes the code more readable so I'm happy for them to stay in.

idavidmcdonald:
Closed as replaced with https://github.com/alphagov/digitalmarketplace-scripts/pull/110

PR #107, opened 2017-04-11
Final script tweaks for g9 closing

TheDoubleK:
Adds G9 lots and declaration fields we want to extract.

Also a small bugfix.

Tests are now dependent on this being merged: 
 - [x] https://github.com/alphagov/digitalmarketplace-utils/pull/305

### DIFF
--- a/scripts/export-framework-applicant-details.py
+++ b/scripts/export-framework-applicant-details.py
@@ -37,7 +37,7 @@ if __name__ == '__main__':
     client = DataAPIClient(get_api_endpoint_from_stage(STAGE), API_TOKEN)
     now = datetime.datetime.now()
 
-    filename = OUTPUT_DIR + os.sep if OUTPUT_DIR else ""
+    filename = (OUTPUT_DIR if OUTPUT_DIR else ".") + os.sep
     filename = filename + FRAMEWORK + "-applicant_details-" + now.strftime("%Y-%m-%d_%H.%M-") + STAGE + ".csv"
 
     # Create output directory if it doesn't already exist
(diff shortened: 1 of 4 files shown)


scripts/export-framework-applicant-details.py
```diff
@@ -37,7 +37,7 @@
     client = DataAPIClient(get_api_endpoint_from_stage(STAGE), API_TOKEN)
     now = datetime.datetime.now()
 
-    filename = OUTPUT_DIR + os.sep if OUTPUT_DIR else ""
+    filename = (OUTPUT_DIR if OUTPUT_DIR else ".") + os.sep
```
risicle: It might be neater to just do `getcwd` in the `else` case but whatever

risicle:
:+1:

scripts/export-framework-applicant-details.py
galund: I've been using a slightly difference pattern for other scripts, which is that if OUTPUT_DIR is not specified, then write to stdout instead of writing a file at all.

I think this is preferable because probably you either want to be explicit about the target location ('.' could be anything), or you are probably debugging stuff in which case stdout is preferable anyway.

For an example see https://github.com/alphagov/digitalmarketplace-frameworks/pull/417/files#diff-9e8880f84bd814f2dc1eeef5d4a11361

risicle: Doesn't work here as it needs to output a number of separate files.

galund: I think I'd still rather be explicit about the target directory, and otherwise just see everything concatenated on stdout, myself :) Otherwise one can overwrite things one did not expect to..

risicle: I can't see a concatenation of the files being useful, but idunno maybe it should require a directory specified. :man_shrugging:

TheDoubleK: I've made it require a directory - makes sense.  I was just thoughtlessly fixing bad behaviour when I made that bugfix for current directory.

PR #108, opened 2017-04-13
Add more keys to briefs data that gets exported to csv

CrystalPea:
Trello card: https://trello.com/c/sdwj5wKt/275-add-more-details-to-automatic-data-export-script

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,4 @@
 __pycache__
 output/
 venv
-venv3
+venv*
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 *.pyc
 __pycache__
 output/
+venv
+venv3
--- a/scripts/get-model-data.py
+++ b/scripts/get-model-data.py
@@ -124,7 +124,10 @@ CONFIGS = [
             'requirementsLength',
             'specialistRole',
             'clarificationQuestions',
-            'frameworkSlug'
+            'frameworkSlug',
+            'budgetRange',
+            'startDate',
+            'contractLength'
         ),
         'get_data_kwargs': {'with_users': 'true'},
         'process_fields': {


benvand:
LGTM
This method https://github.com/alphagov/digitalmarketplace-scripts/blob/master/dmscripts/models/modeltrawler.py#L31-L31
makes sure that if the key doesn't exist then `''` is returned instead of the value.

Running this:
```
client = DataAPIClient('http://localhost:5000', 'myToken')
briefs = client.find_briefs()['briefs']
counter = dict(zip(['frameworkSlug', 'budgetRange', 'startDate', 'contractLength'], [0 for _ in range(4)]))
for brief in briefs:                                                                                       
     for v in ['frameworkSlug', 'budgetRange', 'startDate', 'contractLength']:
         try:
             print brief[
             counter[v] += 1
         except:
             pass
print counter

Out: 
{'budgetRange': 55,
 'contractLength': 85,
 'frameworkSlug': 100,
 'startDate': 95}
```
Shows that the fields are legit.

.gitignore
```diff
@@ -1,3 +1,5 @@
 *.pyc
 __pycache__
 output/
+venv
+venv3
```
idavidmcdonald: Either way works but same outcome could be achieved by `venv*`

idavidmcdonald:
👍 

I left a optional comment about venv things but apart from that it looks good.

idavidmcdonald:
Ah, sorry misunderstanding.
If you have 
`venv*` in your .gitignore then you don't also need to have `venv`.
If you pop a quick pull request in to fix that and then I'll thumbs it.

You might also be interested in reading a bit more about clever ways of gitignoring stuff - http://365git.tumblr.com/post/519016351/three-ways-of-excluding-files

PR #109, opened 2017-04-13
Remove redundant venv from gitignore

CrystalPea:


idavidmcdonald:
👍 good after tests are green

PR #110, opened 2017-04-18
Script to send mailchimp email notifications to suppliers for new briefs

idavidmcdonald:
### Logic around sending emails:
https://trello.com/c/RDQcaDGo/252-write-and-implement-the-script-to-send-the-emails-for-each-lot

The script sends emails with new briefs to suppliers.
The script will be run by a jenkins job every weekday.
The script will look for briefs published on a previous day or three previous days, depending on instructions from the jenkins job. E.g. if email is sent on Monday, the script will search for briefs published on Fri, Sat, Sun. If the email is sent between Tue-Fri, the script will look for briefs published on a previous day to sending.

### Email content
https://trello.com/c/rajSbay6/251-create-the-email-template-for-supplier-notifications
Content reacts dynamically to whether one brief or multiple briefs are included in the email
Content reacts dynamically to whether we are sending an email for the past day of briefs or for multiple days (i.e. all briefs from yesterday or briefs since Friday)
Content dynamically driven by the briefs, including a `utm` tracking link for Ash (https://trello.com/c/XpBzpA4S/255-implement-analytics-for-supplier-opportunities-notification)

### Questions
We have no dry run functionality, users will be emailed regardless of which environment is being used. However you can override and send test emails to a testing list. Does this feel sufficient and also safe enough to prevent user error? Ideas could be taking all the list_id's as command line input or instead of having a 'dry run' mode have a 'real run' mode.

### Screenshots
image

image

image


dmscripts/send_dos_opportunities_email.py
```diff
@@ -0,0 +1,143 @@
+from datetime import datetime, date, timedelta
```
benvand: It's possible to avoid importing `date` here by accessing today's date object through `datetime.today().date()` but that's just an FYI not objectively better or worse IMO

email_templates/dos_opportunities.html
```diff
@@ -0,0 +1,711 @@
...
+                            <p><span style="color:#000000"><span style="font-size:19px">Dear supplier,</span></span><br>
+<br>
+<span id="opening-paragraph" style="color:#000000"><span style="font-size:19px">{% if number_of_days == 1 %}In the last day{% else %}Since {{ start_date.strftime("%A") }}{% endif %}, {{ briefs|length }} new {{ briefs[0].lotName|lower }} {% if briefs|length == 1 %}opportunity was{% else %}opportunities were{% endif %} published on the Digital Marketplace. View and apply for {% if briefs|length == 1 %}this opportunity{% else %}these opportunities{% endif %}:</span></span><br>
+<br>
+
+{% for brief in briefs %}
+	<span class="opportunity-title" style="font-size:16px"><font color="#000000"><strong>{{ brief.title }}</strong></font></span><br>
+	<span class="opportunity-organisation" style="color:#000000">{{ brief.organisation }}<br></span>
+	<span class="opportunity-location" style="color:#000000">{{ brief.location }}</span><br>
+	<span class="opportunity-closing" style="color:#000000">Closing {{ (datetime.strptime(brief.applicationsClosedAt, datetime_format)).strftime(display_date_format) }}</span><br>
```
benvand: Would rather see this date formatting done in `get_html_content` and passed in to the template as template variables; passing the whole library into the template renderer doesn't feel right. The smarter templates are the harder they are to debug/ maintain and it creates a barrier for entry meaning that devs that only know HTML can't maintain your templates for you.

scripts/send_dos_opportunities_email.py
```diff
@@ -0,0 +1,82 @@
...
+from dmapiclient import DataAPIClient
+
+sys.path.insert(0, '.')
+from dmscripts.helpers.env_helpers import get_api_endpoint_from_stage
+from dmscripts.send_dos_opportunities_email import main
+
+
+lots = [
+    {
+        "lot_slug": "digital-specialists",
```
benvand: I'd probably propose a slightly different route here
1.  Not updating these dicts with the list ID: The list ID doesn't bear any reference to the lots as far as I can see. Therefore calling the list `lots` is a bit misleading.
2. Not having a placeholder: Theres no need for the placeholder as both keys and values can be dynamically updated at any point. It doesn't serve any purpose really.

Maybe something like:
```
lots = (
    ("digital-specialists", "Digital specialists"),
    ("digital-outcomes", "Digital outcomes"),
    ("user-research-participants", "User research participants")
)
...
...
for lot_slug, lot_name in lots:
    ok = main(
        data_api_client=data_api_client,
        mailchimp_client=mailchimp_client,
        lot_slug=lot_slug,
        lot_name=lot_name,
        list_id=list_id,
        number_of_days=int(arguments['<number_of_days>'])
    )
```

idavidmcdonald: List id will be different for every lot (one list of recipients for specialists, one for outcomes and so on...)

benvand: Also since the data now never needs to change it's possible to use a tuple. There's a few good reasons for this outlines in the various answers here (especially the most upvoted one):
http://stackoverflow.com/questions/1708510/python-list-vs-tuple-when-to-use-each

dmscripts/send_dos_opportunities_email.py
```diff
@@ -0,0 +1,143 @@
...
+    return {
+        "type": "regular",
+        "recipients": {
+            "list_id": list_id,
+        },
+        "settings": {
+            "subject_line": "New opportunities for {0}: Digital Outcomes and Specialists 2".format(lot_name),
+            "title": "DOS Suppliers: {0} [{1}]".format(lot_name, datetime.now().strftime("%d %B")),
+            "from_name": "Digital Marketplace Team",
+            "reply_to": "do-not-reply@digitalmarketplace.service.gov.uk",
```
benvand: Looks like this value is hard coded in the supplier fe :/
https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/master/config.py#L47-L47

Maybe we could have it in a config file in the dmutils/email repo. Might be worth making a tech debt ticket for that. I'm happy to consider it as slightly out of scope for this ticket.

scripts/send_dos_opportunities_email.py
benvand: @idavidmcdonald 
> List id will be different for every lot (one list of recipients for specialists, one for outcomes and so on...)
Is passing in `--list_id` to the script just an override?
https://github.com/alphagov/digitalmarketplace-scripts/pull/110/files#diff-953d9b342588be3223e805b10a6119adR64

idavidmcdonald: Yep, will override all list id's so one of us devs can pass in our testing list ID (which currently is a list with my email on)

benvand: Ah OK. I thought list id might be passed in every time and that `"list_id": "TO BE FILLED IN"` was to be filled in by the script. That will come in a later commit then? Would be cool to release with this filled in but if it's absolutely impossible then fair enough.

benvand:
Cool so I'm happy with the changes here as long as someone gets back to me on:
https://github.com/alphagov/digitalmarketplace-scripts/pull/110#discussion_r111974238

What I would say is that I feel like this code could have benefited from being a class, maybe something like
```
class DOSLotNotificationHelper(object):

    def __init__(self, lot_name, lot_slug, number_of_days, dataclient, mail_client etc.):
         self.lot_name = lot_name
         self.lot_slug = lot_slug
        self.dataclient = dataclient
        ....
         self.start_date=datetime.today()
         self.earliest_time = self.start_date - datetimedelta(days=number_of_days)
    def get_data(self):
        ...
    def render_mail(self):
        ...
    def send_email(self):
        ...
```

It would have saved a lot of passing around arguments and allowed for creating all the variables in the `__init__` method and accessing them via `self`. 

It effectively already has this structure, decent splitting out into functions etc. but it currently just doesn't have the ability to access variables used in other functions.

If you imagine the object we create as an _instance_ of a script run there are a few things we need, start_date, lot_name, list_id etc. and then there's not that much dynamic stuff going on. We'd just require a few interface methods (very similar to the existing functions), one between the apiclient and the script (to get the data) one between the template renderer and the script (to pass the data to the template), one between the email client and the script (to send the actual email) and we're home free.


But like I say, that's more of a stylistic thing. 
Looks good, is highly readable and well tested :100:

email_templates/dos_opportunities.html
galund: Agree with the above: it'd be much better if `brief.applicationsClosedAt` were already a datetime, then the template could just call its `strftime` method?

What would be _really_ nice would be if our API client knew which fields were datetimes, and automatically deserialized them from the JSON as such :)

idavidmcdonald:
Thanks @benvand. Pea and I agree with your point that this might be nicer in a class but probably won't spend the time to implement that.

We believe we've addressed your comment 110 https://github.com/alphagov/digitalmarketplace-scripts/pull/110#discussion_r111974238 and have also added 2 minor bugfixes since you last looked so we are ready for a final review :)

The final step will then be to add the real list ID's in although I don't want to do this until the very last minute just to reduce the risk of sending emails to real users.

benvand:
LGTM!
[LGTM](https://lgtm.in/i/mtXcg6Spq)

[:+1:](https://lgtm.in/u/mtXcg6Spq) [:-1:](https://lgtm.in/r/mtXcg6Spq)

benvand:
LGTM!
[LGTM](https://lgtm.in/i/uxrIydXNp)

[:+1:](https://lgtm.in/u/uxrIydXNp) [:-1:](https://lgtm.in/r/uxrIydXNp)

benvand:
LGTM!
[LGTM](https://lgtm.in/i/atOSUsdZp)

[:+1:](https://lgtm.in/u/atOSUsdZp) [:-1:](https://lgtm.in/r/atOSUsdZp)

benvand:
LGTM!
[LGTM](https://lgtm.in/i/AQD2Mzm5q)

[:+1:](https://lgtm.in/u/AQD2Mzm5q) [:-1:](https://lgtm.in/r/AQD2Mzm5q)

benvand:
LGTM :+1:

scripts/send_dos_opportunities_email.py
```diff
@@ -61,6 +66,17 @@
 if __name__ == "__main__":
     arguments = docopt(__doc__)
 
+    # Override number of days
+    if arguments.get("--number_of_days"):
+        number_of_days = int(arguments['--number_of_days'])
+    else:
+        day_of_week = date.today().weekday()
```
benvand: :100:

benvand:
:+1: All looks good
For posterity:
We had a quick chat about whether our list ids can be stored in a public repo. David checking into it

idavidmcdonald:
Seems Ok to store list ID's as they are already exposed in mailchimp url's anyway

PR #112, opened 2017-04-21
Abort script if an individual lot fails to send

idavidmcdonald:
Before it was only aborting if the last lot failed which was incorrect. The script should fail as soon as it comes across an error otherwise failures to previous lots will go undetected.


benvand:
lgtm 👍

PR #115, opened 2017-05-02
Make framework required as indexing a mix of DOS and G-Cloud is bad.

galund:


scripts/index-services.py
```diff
@@ -2,11 +2,11 @@
...
 Usage:
-    index-services.py <stage> --api-token=<api_access_token> --search-api-token=<search_api_access_token> [options]
+    index-services.py <stage> --frameworks=<frameworks> --api-token=<api_access_token> --search-api-token=<search_api_access_token> [options]
 
-    --serial        Do not run in parallel (useful for debugging)
-    --index=<index>  Search API index name [default: g-cloud]
-    --frameworks=<frameworks> Optional comma-separated list of framework slugs that should be indexed
+    --serial                  Do not run in parallel (useful for debugging)
+    --index=<index>           Search API index name [default: g-cloud]
+    --frameworks=<frameworks> Comma-separated list of framework slugs that should be indexed
```
TheDoubleK: I *think* that this is a list of the possible `[options]`, so if frameworks is required it should probably be removed from this list of possible options. (But I might be wrong... I don't really know the ins and outs of docopt).

Would be nice to keep in the comment somewhere that `<frameworks>` is a comma-separated list though.

TheDoubleK:
Also, if we're making frameworks required then we can drop the check at line 117, i.e. `if not isinstance(chosen_frameworks, six.string_types):`.

scripts/index-services.py
```diff
@@ -2,11 +2,11 @@
 """Read services from the API endpoint and write to search-api for indexing.
 
 Usage:
-    index-services.py <stage> --api-token=<api_access_token> --search-api-token=<search_api_access_token> [options]
+    index-services.py <stage> --frameworks=<frameworks> --api-token=<api_access_token> --search-api-token=<search_api_access_token> [options]
```
TheDoubleK: To appease PEP8 you can drop the later command line args in the "usage" on to the next line and it works OK, see e.g. https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/notify-suppliers-whether-application-made-for-framework.py#L6

scripts/index-services.py
galund: I checked this, and you can put them in both places. And there is no other place to document a non-optional option, so there is must stay.

scripts/index-services.py
galund: hmm but then you get the newline appearing when you type the command, which is weird.

Looks like you can have line-continuation inside a comment, so let's see what that looks like (!!)

TheDoubleK:
:+1: after a rebase.

PR #116, opened 2017-05-03
copy-my-brief

benvand:
Add new `Brief.parent` field to `get-model-data` export


idavidmcdonald:
See comment in API regarding serialization and camel case. Also the PR description refers to the old naming of 'parent'

idavidmcdonald:
👍

PR #120, opened 2017-05-16
Improve make-g-cloud-live script

TheDoubleK:
When running on Jenkins we saw over 300 services fail to publish but no useful error messages were shown.  On a subsequent run of the script locally all the failed serviced published just fine.

This adds:
 * Per-service backoff so that 5 attempts will be made to publish a service that fails as the result of an API or S3 error
 * Better reporting at the end of the run, showing number of published and failed services
 * Exit with a non-zero status code if there were failures, so Jenkins knows to mark the build red

### DIFF
--- a/scripts/make-g-cloud-live.py
+++ b/scripts/make-g-cloud-live.py
@@ -13,6 +13,13 @@ Usage:
     scripts/make-g-cloud-live.py <framework_slug> <stage> <api_token> <draft_bucket> <documents_bucket> [--dry-run]
 """
 import sys
+from datetime import datetime
+
+import backoff
+import boto
+import collections
+import dmapiclient
+
 sys.path.insert(0, '.')
 
 try:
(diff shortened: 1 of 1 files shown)


scripts/make-g-cloud-live.py
```diff
@@ -138,10 +141,20 @@ def make_draft_service_live(client, copy_document, draft_service, framework_slug
...
 
     for supplier in suppliers:
         print(u"Migrating drafts for supplier {} - {}".format(supplier['supplierId'], supplier['supplierName']))
         draft_services = get_submitted_drafts(client, FRAMEWORK_SLUG, supplier['supplierId'])
-
         for draft_service in draft_services:
-            make_draft_service_live(client, copy_document, draft_service, FRAMEWORK_SLUG, DRY_RUN)
+            try:
+                make_draft_service_live(client, copy_document, draft_service, FRAMEWORK_SLUG, DRY_RUN)
+                results.update({'success': 1})
```
galund: Will this not overwrite a later fail? (i.e. fail, followed by success, will here = success. But I think any fail should give an exit code of 1.)

TheDoubleK: This will update the counter, incrementing by 1 each time, so we get a total count of successes and fails.

e.g. by the end of the run the counter will look something like `{'success': 12345, 'fail': 6}`

TheDoubleK: And so the exit code will be the total number of fails.

galund: oooh nice

scripts/make-g-cloud-live.py
```diff
@@ -89,39 +96,35 @@ def copy_document(draft_document_path, live_document_path):
...
+            assert_equal(str(parsed_document['draft_id']), str(draft_service['id']))
+
+            draft_document_path = get_draft_document_path(parsed_document, framework_slug)
+            live_document_path = get_live_document_path(parsed_document, framework_slug, service_id)
+
+            copy_document(draft_document_path, live_document_path)
+            document_updates[document_key] = get_live_asset_url(live_document_path)
+
+    if dry_run:
+        print(u"    > not updating document URLs {}".format(document_updates))
```
galund: Whilst we're here can we change this to "> dry run: skipped updating document URLs {}" as this seems a bit confusing to me?

TheDoubleK: Yep - done.

scripts/make-g-cloud-live.py
```diff
@@ -138,10 +141,20 @@ def make_draft_service_live(client, copy_document, draft_service, framework_slug
...
+            try:
+                make_draft_service_live(client, copy_document, draft_service, FRAMEWORK_SLUG, DRY_RUN)
+                results.update({'success': 1})
+            except Exception as e:
+                print(u"{} ERROR MIGRATING DRAFT {} TO LIVE: {}".format(datetime.now(), draft_service['id'], e))
+                results.update({'fail': 1})
+
+    print("Successfully published {} G-Cloud services".format(results.get('success')))
+    if(results.get('fail')):
+        print("Failed to publish {} services because of errors".format(results.get('fail')))
```
galund: Some of these prints should probably be to stderr but eh :)

PR #122, opened 2017-05-16
Add users to mailchimp lists using the mailchimp API

idavidmcdonald:
### Changes to existing script that sends mailchimp campaigns for DOS opportunities
- Prevents existing script (`send_dos_opportunities_email`) from calling production mailchimp lists if you call the script to run against a non production environment (instead it will use a test list).
- Removes mailchimp api client functions as these have been moved into the utils library (see https://github.com/alphagov/digitalmarketplace-utils/pull/310)
- Pulls in the new DMMailChimpClient from utils and uses it instead

### New script to add users to mailchimp lists for DOS opportunities
Trello story: https://trello.com/c/XOCeuOYW/254-update-mailchimp-user-list-everyday

- Tests will currently fail until [utils PR](https://github.com/alphagov/digitalmarketplace-utils/pull/310) has been merged
- We have not yet coded in production list ID's to our code. This will be done as the final step just so we definitely avoid running anything against production when we shouldn't do
- Running this on preview will currently fail due to `example.com` email domains being rejected from the mailchimp API. To solve this we intend to change the domains we are using for dummy users in our preview and staging environments (https://github.com/alphagov/digitalmarketplace-aws/pull/281)
- It has been decided that if a user removes themselves as a contributer from a supplier account but their email has already been added to a mailchimp list, we will not go to the effort of also removing them from the mailchimp list. This is because we do not currently do this manually and this process has been ongoing for several months with no user queries/complaints. Users are also able to unsubscribe themselves manually if they so wish at the bottom of any email from mailchimp.

### DIFF
--- a/dmscripts/send_dos_opportunities_email.py
+++ b/dmscripts/send_dos_opportunities_email.py
@@ -1,18 +1,16 @@
 from datetime import datetime, date, timedelta
 
-from requests.exceptions import RequestException
-
 from dmscripts.helpers.html_helpers import render_html
 from dmscripts.helpers import logging_helpers
 from dmscripts.helpers.logging_helpers import logging
 from dmutils.formats import DATETIME_FORMAT, DISPLAY_DATE_FORMAT
 
-import dmapiclient
 
 logger = logging_helpers.configure_logger({'dmapiclient': logging.INFO})
 
 
 def get_live_briefs_between_two_dates(data_api_client, lot_slug, start_date, end_date):
+    """Get all briefs for a lot which were published between 2 dates."""
     return [
         brief for brief in data_api_client.find_briefs_iter(status="live", lot=lot_slug, human=True)
         if datetime.strptime(brief['publishedAt'], DATETIME_FORMAT).date() >= start_date
@@ -66,43 +64,6 @@ def get_html_content(briefs, number_of_days):
     return {"html": html_content}
 
 
-def create_campaign(mailchimp_client, campaign_data):
-    try:
-        campaign = mailchimp_client.campaigns.create(campaign_data)
-        return campaign['id']
-    except RequestException as e:
-        logger.error(
-            "Mailchimp failed to create campaign for '{0}'".format(
-                campaign_data.get("settings").get("title")
-            ),
-            extra={"error": str(e)}
-        )
-    return False
-
-
-def set_campaign_content(mailchimp_client, campaign_id, content_data):
-    try:
-        return mailchimp_client.campaigns.content.update(campaign_id, content_data)
-    except RequestException as e:
-        logger.error(
-            "Mailchimp failed to set content for campaign id '{0}'".format(campaign_id),
-            extra={"error": str(e)}
-        )
-    return False
-
-
-def send_campaign(mailchimp_client, campaign_id):
-    try:
-        mailchimp_client.campaigns.actions.send(campaign_id)
-        return True
-    except RequestException as e:
-        logger.error(
-            "Mailchimp failed to send campaign id '{0}'".format(campaign_id),
-            extra={"error": str(e)}
-        )
-    return False
-
-
 def main(data_api_client, mailchimp_client, lot_data, number_of_days):
     logger.info(
         "Begin process to send DOS notification emails for '{0}' lot".format(lot_data["lot_slug"]),
@@ -127,7 +88,7 @@ def main(data_api_client, mailchimp_client, lot_data, number_of_days):
     logger.info(
         "Creating campaign for '{0}' lot".format(lot_data["lot_slug"])
     )
-    campaign_id = create_campaign(mailchimp_client, campaign_data)
+    campaign_id = mailchimp_client.create_campaign(campaign_data)
     if not campaign_id:
         return False
 
@@ -135,13 +96,13 @@ def main(data_api_client, mailchimp_client, lot_data, number_of_days):
     logger.info(
         "Setting campaign data for '{0}' lot and '{1}' campaign id".format(lot_data["lot_slug"], campaign_id)
     )
-    if not set_campaign_content(mailchimp_client, campaign_id, content_data):
+    if not mailchimp_client.set_campaign_content(campaign_id, content_data):
         return False
 
     logger.info(
         "Sending campaign for '{0}' lot and '{1}' campaign id".format(lot_data["lot_slug"], campaign_id)
     )
-    if not send_campaign(mailchimp_client, campaign_id):
+    if not mailchimp_client.send_campaign(campaign_id):
         return False
 
     return True
--- a/dmscripts/upload_dos_opportunities_email_list.py
+++ b/dmscripts/upload_dos_opportunities_email_list.py
@@ -27,10 +27,16 @@ def main(data_api_client, mailchimp_client, lot_data, logger):
         "{} emails have been found for lot {}".format(len(emails), lot_data["lot_slug"])
     )
 
+    existing_mailchimp_emails = mailchimp_client.get_email_addresses_from_list(lot_data["list_id"])
     logger.info(
-        "Subscribing new emails to mailchimp list {}".format(lot_data["list_id"])
+        "{} existing emails found on list {}".format(len(existing_mailchimp_emails), lot_data["list_id"])
     )
-    if not mailchimp_client.subscribe_new_emails_to_list(lot_data["list_id"], emails):
+
+    new_emails = list(set(emails) - set(existing_mailchimp_emails))
+    logger.info(
+        "Subscribing {} new emails to mailchimp list {}".format(len(new_emails), lot_data["list_id"])
+    )
+    if not mailchimp_client.subscribe_new_emails_to_list(lot_data["list_id"], new_emails):
         return False
 
     return True
--- a/dmscripts/upload_dos_opportunities_email_list.py
+++ b/dmscripts/upload_dos_opportunities_email_list.py
@@ -10,11 +10,14 @@ def find_user_emails(supplier_users, services):
     return email_addresses
 
 
-def main(data_api_client, lot_data):
+def main(data_api_client, mailchimp_client, lot_data):
     data_helper = SupplierFrameworkData(data_api_client, lot_data["framework_slug"])
     supplier_users = data_helper.get_supplier_users()
 
-    framework_services = data_api_client.find_services_iter(framework=lot_data["framework_slug"], lot=lot_data["lot_slug"])
+    framework_services = data_api_client.find_services_iter(
+        framework=lot_data["framework_slug"], lot=lot_data["lot_slug"]
+    )
     emails = find_user_emails(supplier_users, framework_services)
     print emails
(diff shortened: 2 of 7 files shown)


dmscripts/send_dos_opportunities_email.py
```diff
@@ -1,23 +1,22 @@
...
-        if datetime.strptime(brief['publishedAt'], DATETIME_FORMAT).date() >= start_date
-        and datetime.strptime(brief['publishedAt'], DATETIME_FORMAT).date() <= end_date
-    ]
+    """Get all briefs for a lot which were published between 2 dates."""
+    valid_briefs = []
+    for brief in data_api_client.find_briefs_iter(status="live", lot=lot_slug, human=True):
+        brief_published_at = datetime.strptime(brief['publishedAt'], DATETIME_FORMAT).date()
+        if start_date <= brief_published_at <= end_date:
+            valid_briefs.append(brief)
+    return valid_briefs
```
risicle: Yeah, the original way was better.

scripts/send_dos_opportunities_email.py
```diff
@@ -81,6 +85,14 @@
         else:
             number_of_days = 1
 
+    # Override list for non-production environment
+    if arguments['<stage>'] != "production":
+        logger.info(
+            "The environment is not production. Emails will be sent to test list unless you set the list id manually."
+        )
+        for lot in lots:
+            lot.update({"list_id": "096e52cebb"})
```
risicle: :\

risicle:
What's the thinking around extending our mailchimp interfaces vs. migrating to notify?

tests/test_send_dos_opportunities_email.py
```diff
@@ -265,19 +193,17 @@ def test_main_gets_live_briefs_for_three_days(get_live_briefs_between_two_dates)
...
+    dm_mailchimp_client = mock.MagicMock(DMMailChimpClient)
+    result = main(mock.MagicMock(), dm_mailchimp_client, LOT_DATA, 3)
+
     assert result is True
-    assert create_campaign.call_count == 0
-    assert set_campaign_content.call_count == 0
-    assert send_campaign.call_count == 0
+    assert dm_mailchimp_client.create_campaign.call_count == 0
+    assert dm_mailchimp_client.set_campaign_content.call_count == 0
+    assert dm_mailchimp_client.send_campaign.call_count == 0
```
risicle: Y'can just do `.called is False`, y'know...

scripts/upload_dos_opportunities_email_list.py
```diff
@@ -0,0 +1,67 @@
...
+            "framework_slug": "digital-outcomes-and-specialists-2"
+        },
+        {
+            "lot_slug": "digital-outcomes",
+            "list_id": "prod_list_id2" if arguments['<stage>'] == "production" else "f0077c516d",
+            "framework_slug": "digital-outcomes-and-specialists-2"
+        },
+        {
+            "lot_slug": "user-research-participants",
+            "list_id": "prod_list_id3" if arguments['<stage>'] == "production" else "d35601203b",
```
risicle: all this hardcoded presumptuous stuff is weird, man.

risicle: (I mean, what the poor australians are supposed to think about all this, god knows...)

dmscripts/upload_dos_opportunities_email_list.py
```diff
@@ -0,0 +1,46 @@
...
+
+def find_user_emails(supplier_users, services):
+    """Return email addresses for any supplier who has a service in services."""
+    email_addresses = []
+    for service in services:
+        for user in supplier_users.get(service['supplierId'], []):
+            email_addresses.append(user['email address'])
+    return email_addresses
+
+def lowercase_list(list):
```
risicle: Builtins names as variable names/argument names are always a recipe for disaster.

dmscripts/upload_dos_opportunities_email_list.py
```diff
@@ -0,0 +1,46 @@
+from dmscripts.helpers.supplier_data_helpers import SupplierFrameworkData
+
+
+def find_user_emails(supplier_users, services):
+    """Return email addresses for any supplier who has a service in services."""
+    email_addresses = []
+    for service in services:
+        for user in supplier_users.get(service['supplierId'], []):
+            email_addresses.append(user['email address'])
+    return email_addresses
```
risicle:
```
list(chain.from_iterable(
    (user['email address'] for user in supplier_users.get(service['supplierId'], []))
    for service in services
))
```

dmscripts/send_dos_opportunities_email.py
idavidmcdonald: Think I agree. I also checked with flake8 and it had no issues with complexity so have changed back.

scripts/upload_dos_opportunities_email_list.py
risicle: (i know you didn't start the pattern)

dmscripts/upload_dos_opportunities_email_list.py
idavidmcdonald: Good point. Have replaced but couldn't think of anything better than a generic `x`

dmscripts/upload_dos_opportunities_email_list.py
idavidmcdonald: Your suggestion looks a bit shorter and more pythonic but could you explain what you mean by far less error prone please (as in what is the risk of errors and why on the current solution)

scripts/send_dos_opportunities_email.py
idavidmcdonald: Any suggestions on how to improve?

dmscripts/upload_dos_opportunities_email_list.py
risicle: If you do actually want to call it e.g. `list` there's nothing wrong with using `list_`.

dmscripts/send_dos_opportunities_email.py
risicle: /me shudders at flake8's complexity measures

risicle: ...and if it had complained about the list comprehension's complexity measure and not the manual loop that would have been damning evidence against it, because there are far *far* more ways the manual loop can be subtly wrong.

scripts/upload_dos_opportunities_email_list.py
idavidmcdonald: Just so I'm clear about the stuff you find weird, are you referring specifically to having hardcoded list ID's in the code, or the inline `if` statements or the general structure of the `lots` list?

Because I may actually be responsible for some of these patterns ;)

dmscripts/upload_dos_opportunities_email_list.py
risicle: Spot the bug

```
mydict = {}
for something in somethings:
    inner = banana_list + [3]
    for another in anothers(something):
        inner.append(another)
    mydict[another] = inner

mylist = []
for bunch in bunches
    for grape in bunch:
        inner.append(grape)
    mylist.append(inner)
```

risicle: ...where what the latter is trying to do is

```
[list(bunch) for bunch in bunches]
```

risicle: Essentially, with the imperative way of doing it, you're commanding a little robot with little instructions of what to do. With the functional way of doing it, you're more *describing the result* you want to achieve. As such it's also far easier for the next developer coming along to glean the programmer's intention.

scripts/upload_dos_opportunities_email_list.py
risicle: No just the general mixture of admin policy and code, the code knowing what "staging" is etc., having hardcoded ids... My comment about the australians is that they probably have a different setup and certainly have different ids for things.

idavidmcdonald: Great point, it is a bit nasty (but on the plus side fairly contained in it's scope to the `lots` dict).

We could remove the if statement based on environment by doing something like:
```
CONFIG = {
    "preview": {
        "digital-specialists": "list id",
        "digital-outcomes": "list id2",
        "user-research-participants": "list id3",
    },
    "production": {
        "digital-specialists": "list id4",
        "digital-outcomes": "list id5",
        "user-research-participants": "list id6",
    }
} # would have all stages in real code
```

```
lots = [
        {
            "lot_slug": "digital-specialists",
            "list_id": CONFIG[arguments['<stage>']]['digital-specialists'],
            "framework_slug": "digital-outcomes-and-specialists-2"
        }
] # would have all lots in real code
```

It would be a bit nicer but without moving list ID's into the database or storing them in credentials repo for every stage for every lot and passing them into the script via jenkins instead we still have a mix of code and policy.

Do you feel the current solution is tolerable or it's worth looking into any of the above?

risicle:
On Thursday 18 May 2017 07:52:30 David McDonald wrote:
> Do you feel the current solution is tolerable or it's worth looking into any of the above?

Yeah don't worry about it it's not your problem it's just a general thing, there are all kinds of places we mix configuration/policy and code. I wouldn't really know where to begin at this point. Just perhaps be aware that it should make you feel dirty.

dmscripts/upload_dos_opportunities_email_list.py
idavidmcdonald: Fair enough in terms of it may be easier to code a mistake but I'd say all that matters is you have correct test coverage and your tests pass

idavidmcdonald:
I think I've dealt with all your points @risicle (and I do indeed feel dirty with my lovely code and policy in the same file). Do let me know if you are happy or you have any other suggestions.

The one thing I hadn't yet answered was in regards to your comment:
'What's the thinking around extending our mailchimp interfaces vs. migrating to notify?'

We can't use notify for these emails because they are not transactional emails but marketing emails. Notify can not send marketing emails and we therefore require sending them through something like mailchimp which includes unsubscribe links at the bottom of our emails automatically (unless we build our own subscription management which is a no no).

dmscripts/upload_dos_opportunities_email_list.py
risicle: :warning:

I couldn't disagree more. Firstly you're disregarding the next person to come along to the code and want to know what this block is up to. Written my way it is *much* easier to see that you're just performing a familiar pattern (and haven't screwed it up) because youre *describing the result you want*.

Also the "tests pass" thing totally depends on how good you are at writing tests, and *nobody* is as good as they think they are.

risicle:
> Notify can not send marketing emails and we therefore require sending them through something like mailchimp which includes unsubscribe links at the bottom of our emails automatically

*Oh.*

dmscripts/upload_dos_opportunities_email_list.py
benvand: https://google.github.io/styleguide/pyguide.html#List_Comprehensions

Happy with the current implementation. Easy to read and understand (especially for non python natives) and as this is a script readability can be favoured over speed optimisations.

benvand:
👍

PR #123, opened 2017-05-17
Remove service and supplier names from script output

TheDoubleK:
Running on Jenkins we're getting "'ascii' codec can't encode character..." errors.  There's not really any need to have supplier or service names in the output, as any tracking of errors will rely on IDs rather than names anyway.


MuriloDalRi:
Best branch name I've seen so far.

PR #124, opened 2017-05-19
Could be serviceSummary or serviceDescription and count bad words

TheDoubleK:
For G-Cloud 9 the `serviceSummary` key has changed to `serviceDescription`, so either should work here.

 Also adds a handy counter to get a summary of all the bad words found.

### DIFF
--- a/scripts/scan-g-cloud-services-for-bad-words.py
+++ b/scripts/scan-g-cloud-services-for-bad-words.py
@@ -8,6 +8,8 @@ Usage:
 """
 
 import sys
+from collections import Counter
+
 sys.path.insert(0, '.')
 import os
 if sys.version_info > (3, 0):
@@ -55,6 +57,8 @@ KEYS_TO_CHECK = {
                   'webInterfaceAccessibilityDescription', 'webInterfaceAccessibilityTesting', 'webInterfaceUsage'],
 }
 
+BAD_WORDS_COUNTER = Counter()
+
 
 def main(stage, data_api_token, bad_words_path, framework_slug, output_dir):
     api_url = get_api_endpoint_from_stage(stage)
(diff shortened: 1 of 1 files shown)


scripts/scan-g-cloud-services-for-bad-words.py
```diff
@@ -115,17 +119,31 @@ def check_services_with_bad_words(output_dir, framework_slug, client, suppliers,
...
                             if get_bad_words_in_value(word, service.get(key)):
                                 output_bad_words(
-                                    supplier["supplierId"], framework_slug, service["id"],
-                                    service["serviceName"], service["serviceSummary"], key,
-                                    service.get(key), word, writer)
+                                    supplier["supplierId"],
+                                    framework_slug,
+                                    service["id"],
+                                    service["serviceName"],
+                                    service.get("serviceSummary", service["serviceDescription"]),
```
samuelhwilliams: Needs to be `service.get("serviceDescription")` or it will fail on services without that key.

TheDoubleK: If it has neither then maybe it should just die though?  we shouldn't have any services without either key.

samuelhwilliams: It will die even if it has serviceSummary, because it tries to fetch serviceDescription before checking whether it has serviceSummary.

TheDoubleK: I don't think it does... I ran this against G9 and it was happy.

samuelhwilliams: It's happy with G9 because G9 uses serviceDescription, so direct access works. If you run it against G8 or earlier it will break.

TheDoubleK:
All done.

PR #125, opened 2017-05-23
Bump mailchimp client to 2.0.11

idavidmcdonald:
The bugfixes introduced since 2.0.8 mean exceptions will not
be raised when validating email addresses that contain
apostrophes in the local part of the email address. These
are in fact valid email addresses that mailchimp accepts.

In our production database we have emails as such as we don't
want our script to fail when it tries to send them to the API.


TheDoubleK:
Seems legit :+1:

PR #126, opened 2017-06-02
Fix intermittent test fail due to call argument ordering

benvand:


TheDoubleK:
I'm going to merge this to fix the intermittent test failure on my earlier PR :/

PR #128, opened 2017-06-12
Add g-cloud services to extract-model-data.py

Wynndow:
CCS have requested a CSV of all current G9 services, including the service id, supplier, name of service and lot. This is so they can confirm there spend data before publishing it. The data that this will give them will allow them to cross reference against their own datasets and ensure they have the correct figures.

### DIFF
--- a/scripts/get-model-data.py
+++ b/scripts/get-model-data.py
@@ -109,6 +109,23 @@ CONFIGS = [
         'sort_by': ['frameworkSlug', 'supplierId', 'lotSlug']
     },
     {
+        'name': 'g_cloud_services',
+        'base_model': 'services',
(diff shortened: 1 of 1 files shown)


TheDoubleK:
`No description given` :disappointed:

Why do we need this? Why G8 and G9? The change is probably fine, but without context it's hard to say if it's the right thing/a thing that we need.

Wynndow:
Sorry Kev, have added more info.

Wynndow:
[CCS have requested a csv file of all current G9 service IDs. This would include Service ID, Supplier, name of service and Lot. (Request has come from Milan to get done(.](https://trello.com/c/6cUjHAmr/174-ccs-have-requested-a-csv-file-of-all-current-g9-service-ids-this-would-include-service-id-supplier-name-of-service-and-lot-reque)

PR #130, opened 2017-06-23
Add G-Cloud 9 to script description

TheDoubleK:
This has worked with G9 for a while now, but the description still said only G8 and DOS2.


TheDoubleK:
This is only a comment change and no-one has looked at it for 12 days so I'm going to just merge.

PR #131, opened 2017-07-05
Add default.nix

risicle:
See https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/719 for discussion of general issues.

It's a pity, but one of the dependencies I was most hoping to sort out here was `wkhtmltopdf` - but as of the nix 17.03 branch this is disabled for darwin for a couple of reasons. So for now, mac users needing generate pdfs will need to provide their own `wkhtmltopdf`. Sorry about that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 __pycache__
 output/
 venv*
+
+# local environment customizations
+local.nix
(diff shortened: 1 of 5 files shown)


risicle:
See https://github.com/alphagov/digitalmarketplace-manual/pull/6 for proposed corresponding manual page.

TheDoubleK:
:+1:

PR #132, opened 2017-07-13
Tell interested suppliers that a question has been answered

CrystalPea:
Suppliers want to be notified when there's a new answer for opportunities they are interested in

Trello ticket: https://trello.com/c/2noI0EFB/472-tell-interested-suppliers-that-a-question-has-been-answered

### DIFF
--- a/dmscripts/helpers/env_helpers.py
+++ b/dmscripts/helpers/env_helpers.py
@@ -1,9 +1,9 @@
 def get_api_endpoint_from_stage(stage, app='api'):
-    """Return the full URL of given API or Search API environment.
+    """Return the full URL of given API, Search API or web URL environment.
 
     :param stage: environment name. Can be one of 'preview', 'staging',
                   'production' or 'dev' (aliases: 'local', 'development').
-    :param app: should be either 'api' or 'search-api'
+    :param app: should be either 'api', 'search-api' or 'www'
 
     """
 
@@ -16,6 +16,7 @@ def get_api_endpoint_from_stage(stage, app='api'):
     dev_ports = {
         'api': 5000,
         'search-api': 5001,
+        'www': 80
     }
 
     if stage in ['local', 'dev', 'development']:
--- a/dmscripts/helpers/env_helpers.py
+++ b/dmscripts/helpers/env_helpers.py
@@ -1,9 +1,9 @@
 def get_api_endpoint_from_stage(stage, app='api'):
-    """Return the full URL of given API, Search API or web URL environment.
+    """Return the full URL of given API or Search API environment.
 
     :param stage: environment name. Can be one of 'preview', 'staging',
                   'production' or 'dev' (aliases: 'local', 'development').
-    :param app: should be either 'api', 'search-api' or 'www'
+    :param app: should be either 'api' or 'search-api'
 
     """
 
@@ -16,12 +16,28 @@ def get_api_endpoint_from_stage(stage, app='api'):
     dev_ports = {
         'api': 5000,
         'search-api': 5001,
-        'www': 80
     }
 
     if stage in ['local', 'dev', 'development']:
         return 'http://localhost:{}'.format(dev_ports[app])
 
+    return stage_domains[stage]
+
+
+def get_web_url_from_stage(stage):
+    """Return the full URL of given web environment.
+
+    :param stage: environment name. Can be one of 'preview', 'staging',
+                  'production' or 'dev' (aliases: 'local', 'development').
+    """
+    if stage in ['local', 'dev', 'development']:
+        return 'http://localhost'
+
+    stage_domains = {
+        'preview': 'https://www.preview.marketplace.team',
+        'staging': 'https://www.staging.marketplace.team',
+        'production': 'https://www.digitalmarketplace.service.gov.uk',
+    }
     return stage_domains[stage]
 
 
--- a/dmscripts/notify_suppliers_of_new_questions_answers.py
+++ b/dmscripts/notify_suppliers_of_new_questions_answers.py
@@ -64,7 +64,7 @@ def invert_a_dictionary_so_supplier_id_is_key_and_brief_id_is_value(dictionary_t
 
 def get_supplier_email_addresses_by_supplier_id(data_api_client, supplier_id):
     response = data_api_client.find_users(supplier_id=supplier_id)
-    return [user['emailAddress'] for user in response['users']]
+    return [user['emailAddress'] for user in response['users'] if user['active']]
 
 
 def create_context_for_supplier(stage, supplier_briefs):
--- a/dmscripts/notify_suppliers_of_new_questions_answers.py
+++ b/dmscripts/notify_suppliers_of_new_questions_answers.py
@@ -105,7 +105,7 @@ def send_supplier_emails(email_api_key, email_addresses, supplier_context, logge
     )
 
 
-def main(data_api_url, data_api_token, email_api_key, stage, dry_run):
+def main(data_api_url, data_api_token, email_api_key, stage, dry_run, exclude_supplier_ids=[]):
     logger.info("Begin to send brief update notification emails")
 
     # get today at 8 in the morning
@@ -124,6 +124,10 @@ def main(data_api_url, data_api_token, email_api_key, stage, dry_run):
 
     # find the IDs of interested suppliers {supplier_id: [briefid1, briefid2]}
     interested_suppliers = get_ids_of_interested_suppliers_for_briefs(data_api_client, briefs)
+    interested_suppliers = dict(
+        (supplier_id, briefs) for supplier_id, briefs in interested_suppliers.iteritems()
+        if supplier_id not in exclude_supplier_ids
+    )
     logger.info(
         "{} suppliers found interested in these briefs".format(len(interested_suppliers))
     )
--- a/dmscripts/notify_suppliers_of_new_questions_answers.py
+++ b/dmscripts/notify_suppliers_of_new_questions_answers.py
@@ -125,7 +125,7 @@ def main(data_api_url, data_api_token, email_api_key, stage, dry_run, exclude_su
     # find the IDs of interested suppliers {supplier_id: [briefid1, briefid2]}
     interested_suppliers = get_ids_of_interested_suppliers_for_briefs(data_api_client, briefs)
     interested_suppliers = dict(
-        (supplier_id, briefs) for supplier_id, briefs in interested_suppliers.iteritems()
+        (supplier_id, briefs) for supplier_id, briefs in interested_suppliers.items()
         if supplier_id not in exclude_supplier_ids
     )
     logger.info(
--- a/dmscripts/notify_suppliers_of_new_questions_answers.py
+++ b/dmscripts/notify_suppliers_of_new_questions_answers.py
@@ -42,7 +42,7 @@ def get_ids_of_interested_suppliers_for_briefs(data_api_client, briefs):
         suppliers_who_asked_a_question = get_ids_of_suppliers_who_asked_a_clarification_question(data_api_client, brief)
         interested_suppliers[brief['id']] = list(set(suppliers_who_applied + suppliers_who_asked_a_question))
 
-    return interested_suppliers
+    return invert_a_dictionary_so_supplier_id_is_key_and_brief_id_is_value(interested_suppliers)
 
 
 def invert_a_dictionary_so_supplier_id_is_key_and_brief_id_is_value(dictionary_to_invert):
@@ -59,6 +59,10 @@ def get_supplier_email_addresses_by_supplier_id(data_api_client, supplier_id):
     return [user['emailAddress'] for user in response['users']]
 
 
+def create_context_for_supplier(list_of_brief_ids):
+    pass
+
+
 def main(data_api_client, number_of_days):
     logger.info("Begin to send brief update notification emails")
 
@@ -67,11 +71,14 @@ def main(data_api_client, number_of_days):
     # get yesterday at 8 in the morning
     start_date = end_date - timedelta(days=number_of_days)
 
-    # we need to find the briefs
+    # we want to find all questions and answers that were submitted between start and end dates
     briefs = get_live_briefs_with_new_questions_and_answers_between_two_dates(data_api_client, start_date, end_date)
 
-    # we want to find all questions and answers that were submitted between start and end dates
+    # find the IDs of interested suppliers {supplier_id: [briefid1, briefid2]}
+    interested_suppliers = get_ids_of_interested_suppliers_for_briefs(data_api_client, briefs)
 
-    # look for people who have asked clarification questions
-    # data_api_client.find_audit_events(
-    # audit_type=dmapiclient.audit.AuditTypes.send_clarification_question, audit_date=start_date.strftime('%Y-%m-%d'))
+    for supplier_id, brief_ids in interested_suppliers.items():
+        # get a context for each supplier email
+        for email_address in get_supplier_email_addresses_by_supplier_id(data_api_client, supplier_id):
+            # use notify client to send email with email address, template id and context
+            pass
--- a/dmscripts/notify_suppliers_of_new_questions_answers.py
+++ b/dmscripts/notify_suppliers_of_new_questions_answers.py
@@ -91,7 +91,7 @@ def get_html_content(context):
     })
 
 
-def send_supplier_emails(email_api_key, email_addresses, supplier_context):
+def send_supplier_emails(email_api_key, email_addresses, supplier_context, logger):
 
     send_email(
         email_addresses,
@@ -100,7 +100,8 @@ def send_supplier_emails(email_api_key, email_addresses, supplier_context):
         EMAIL_SUBJECT,
         EMAIL_FROM_ADDRESS,
         EMAIL_FROM_NAME,
-        ["supplier-new-brief-questions-answers"]
+        ["supplier-new-brief-questions-answers"],
+        logger=logger
     )
 
 
@@ -117,9 +118,15 @@ def main(data_api_url, data_api_token, email_api_key, stage, dry_run):
 
     # we want to find all questions and answers that were submitted between start and end dates
     briefs = get_live_briefs_with_new_questions_and_answers_between_two_dates(data_api_client, start_date, end_date)
+    logger.info(
+        "{} briefs with new question and answers found between {} and {}".format(len(briefs), start_date, end_date)
+    )
 
     # find the IDs of interested suppliers {supplier_id: [briefid1, briefid2]}
     interested_suppliers = get_ids_of_interested_suppliers_for_briefs(data_api_client, briefs)
+    logger.info(
+        "{} suppliers found interested in these briefs".format(len(interested_suppliers))
+    )
 
     for supplier_id, brief_ids in interested_suppliers.items():
         # Get the brief objects for this supplier
@@ -143,6 +150,6 @@ def main(data_api_url, data_api_token, email_api_key, stage, dry_run):
                     'brief_ids_list': ", ".join(map(str, brief_ids))
                 }
             )
-            send_supplier_emails(email_api_key, email_addresses, supplier_context)
+            send_supplier_emails(email_api_key, email_addresses, supplier_context, logger)
 
     return True
--- a/dmscripts/notify_suppliers_of_new_questions_answers.py
+++ b/dmscripts/notify_suppliers_of_new_questions_answers.py
@@ -5,6 +5,7 @@ from datetime import datetime, date, timedelta
 from dmscripts.helpers import env_helpers, logging_helpers
 from dmscripts.helpers.html_helpers import render_html
 from dmscripts.helpers.logging_helpers import logging
+from dmutils.email.exceptions import EmailError
 from dmutils.email.dm_mandrill import send_email
 from dmutils.formats import DATETIME_FORMAT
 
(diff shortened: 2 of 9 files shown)


dmscripts/notify_suppliers_of_new_questions_answers.py
```diff
@@ -0,0 +1,159 @@
...
+from datetime import datetime, date, timedelta
+
+from dmscripts.helpers import env_helpers, logging_helpers
+from dmscripts.helpers.html_helpers import render_html
+from dmscripts.helpers.logging_helpers import logging
+from dmutils.email.dm_mandrill import send_email
+from dmutils.formats import DATETIME_FORMAT
+
+
+EMAIL_SUBJECT = "New question and answer responses on the Digital Marketplace"
```
TheDoubleK: For some reason the content doc for this is private., but this subject line doesn't make any sense to me.  What are "answer responses"?  

Either "New question responses..." or (better) "New answers..." seem better to me.

~~It would be nice to customise the subject to include the title (or at least reference number) of the brief rather than be generic (I imagine getting six emails a day with the exact same subject line will look pretty odd/crappy/spammy).~~

OK so it's one email with possibly multiple briefs listed.  Still, I don't think the subject line is right.

dmscripts/notify_suppliers_of_new_questions_answers.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def get_ids_of_interested_suppliers_for_briefs(data_api_client, briefs):
+    interested_suppliers = {}
+    for brief in briefs:
+        suppliers_who_applied = get_ids_of_suppliers_who_started_applying(data_api_client, brief)
+        suppliers_who_asked_a_question = get_ids_of_suppliers_who_asked_a_clarification_question(data_api_client, brief)
+        interested_suppliers[brief['id']] = list(set(suppliers_who_applied + suppliers_who_asked_a_question))
+
+    return invert_a_dictionary_so_supplier_id_is_key_and_brief_id_is_value(interested_suppliers)
```
TheDoubleK: Could this just be `return dict((v,k) for k,v in interested_suppliers.iteritems())`?  Or is there more going on?  Something to do with duplicate values?

dmscripts/notify_suppliers_of_new_questions_answers.py
```diff
@@ -0,0 +1,159 @@
...
+        EMAIL_FROM_NAME,
+        ["supplier-new-brief-questions-answers"],
+        logger=logger
+    )
+
+
+def main(data_api_url, data_api_token, email_api_key, stage, dry_run, exclude_supplier_ids=[]):
+    logger.info("Begin to send brief update notification emails")
+
+    # get today at 8 in the morning
```
TheDoubleK: What time are we planning on running this script?  It feels like it's potentially going to hit the API a *lot* when it does.  I guess if the numbers of live briefs are relatively low it might not be too bad, but we should keep a close eye on if it unduly loads the API.

scripts/notify-suppliers-of-new-questions-answers.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python
+"""
+Send email notifications to supplier users when a question has been answered
+
+Usage: notify-suppliers-of-new-questions-answers.py <stage> <api_token> <mandrill_api_key> [options]
+
+Options:
+    --dry-run                               List notifications that would be sent without sending the emails
+    --exclude-supplier-ids=EXCLUDE-SUPPLIERS   Comma separated list of suppliers IDs to be excluded. This is incase the
```
TheDoubleK: If it used the Notify API then duplicate messages can be excluded automatically by Notify ¯\\_(ツ)_/¯

TheDoubleK: Also is a comma-separated list the best thing?  What if there are 100s of suppliers who already received the thing and only a couple left to go?  Maybe better to read from a file? (Although, if running on Jenkins I guess that is more difficult..)

dmscripts/notify_suppliers_of_new_questions_answers.py
```diff
@@ -0,0 +1,159 @@
...
+    )
+
+    # find the IDs of interested suppliers {supplier_id: [briefid1, briefid2]}
+    interested_suppliers = get_ids_of_interested_suppliers_for_briefs(data_api_client, briefs)
+    interested_suppliers = dict(
+        (supplier_id, briefs) for supplier_id, briefs in interested_suppliers.iteritems()
+        if supplier_id not in exclude_supplier_ids
+    )
+    logger.info(
+        "{} suppliers found interested in these briefs".format(len(interested_suppliers))
```
TheDoubleK: Should the message be "{} suppliers found interested in these briefs who are not excluded"?

dmscripts/notify_suppliers_of_new_questions_answers.py
```diff
@@ -0,0 +1,159 @@
...
+    briefs = get_live_briefs_with_new_questions_and_answers_between_two_dates(data_api_client, start_date, end_date)
+    logger.info(
+        "{} briefs with new question and answers found between {} and {}".format(len(briefs), start_date, end_date)
+    )
+
+    # find the IDs of interested suppliers {supplier_id: [briefid1, briefid2]}
+    interested_suppliers = get_ids_of_interested_suppliers_for_briefs(data_api_client, briefs)
+    interested_suppliers = dict(
+        (supplier_id, briefs) for supplier_id, briefs in interested_suppliers.iteritems()
+        if supplier_id not in exclude_supplier_ids
```
TheDoubleK: Should it rather be a list of email addresses that are excluded?  What if the job fails and has sent only to some of the accounts for a particular supplier?  The other half will miss out.

email_templates/suppliers_new_briefs_questions_answers.html
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
+<html>
+<head lang="en">
+    <meta charset="UTF-8">
+</head>
+<body>
+Dear supplier,
+<br /><br />
+In the last day, buyers answered questions about:
```
TheDoubleK: I don't think this needs a comma here.

dmscripts/notify_suppliers_of_new_questions_answers.py
```diff
@@ -0,0 +1,159 @@
...
+def _get_link_domain(stage):
+    return env_helpers.get_api_endpoint_from_stage(stage, app='www')
+
+
+def create_context_for_supplier(stage, supplier_briefs):
+    return {
+        'briefs': [
+            {
+                'brief_title': brief['title'],
+                'brief_link': '{0}/{1}/opportunities/{2}'.format(
```
TheDoubleK: Might be better to add the `utm_id` to the link here rather than doing it twice in the email template?

tests/test_notify_suppliers_of_new_questions_answers.py
```diff
@@ -0,0 +1,408 @@
...
+        {"publishedAt": "2017-03-18T06:00:00.669156Z"},
+        {"publishedAt": "2017-03-19T06:00:00.669156Z"},  # Sunday
+        {"publishedAt": "2017-03-20T06:00:00.669156Z"},
+    ]}
+]
+
+FILTERED_BRIEFS = [ALL_BRIEFS[3], ALL_BRIEFS[4], ALL_BRIEFS[5]]
+
+
+MODULE_UNDER_TEST = 'dmscripts.notify_suppliers_of_new_questions_answers'
```
TheDoubleK: I don't think this is a pattern we've ever used before... not sure what benefit it brings other than a lot of ALL_CAPS in the mock setups?

TheDoubleK: Is it about getting lines short enough that PEP8 doesn't complain?  If so, 😢

TheDoubleK: If we really need it could we `.format()` rather than concatenate?

katstevens: This was my idea 😄 Yes, mostly to shorten the line length. I got into the habit of using `MODULE_UNDER_TEST +` or `DIFFERENT_MODULE_NAME + ` to ensure I was mocking the right function from the right place (e.g. a function imported from elsewhere, an external library etc). The only reason for using concatenate was increasing readability inside the patch decorator.

TheDoubleK:
I just noticed on the Trello card that you're planning to send the emails at 8am, the same time as the cut-off for working out which things to include.  I'd be tempted to have at least an hour's gap between the the time of sending and the latest changes to pick up as a "just in case" for time weirdness, daylight saving stuff and out-of-sync clocks.  I have no concrete examples to offer of how things could go wrong, just a gut feel that there's a risk of sending the same notifications twice if you have all the times at 8 on the dot.

dmscripts/notify_suppliers_of_new_questions_answers.py
katstevens: I was wondering about that. Could we use Jenkins' smart scheduling where it chooses a quiet time between hours X and Y to run the job?

dmscripts/notify_suppliers_of_new_questions_answers.py
katstevens: Good point - for brevity we could use user IDs instead of supplier IDs?

dmscripts/helpers/env_helpers.py
```diff
@@ -1,9 +1,9 @@
 def get_api_endpoint_from_stage(stage, app='api'):
-    """Return the full URL of given API or Search API environment.
+    """Return the full URL of given API, Search API or web URL environment.
 
     :param stage: environment name. Can be one of 'preview', 'staging',
                   'production' or 'dev' (aliases: 'local', 'development').
-    :param app: should be either 'api' or 'search-api'
+    :param app: should be either 'api', 'search-api' or 'www'
```
katstevens: I did originally have a separate function for `www`, but it was nearly identical so I used this one to keep things DRYer. I'd be happy to rename the function to something more generic but I think you're right, we should be passing the domain as an arg to the script. It must be part of the environment context somewhere already, right?

dmscripts/notify_suppliers_of_new_questions_answers.py
katstevens: Actually, with the Mandrill API we're sending all the emails for a supplier at once (multiple recipients in the `to:` field). So if it fails, it will fail for all users for a supplier.

TheDoubleK: Ah OK, supplier IDs makes sense then.

dmscripts/notify_suppliers_of_new_questions_answers.py
TheDoubleK: That gives you a range around a particular hour, and we use it for all our scheduled jobs, but you still need to specify the hour around which it's smart-scheduled, and I'd say that shouldn't be 8.

dmscripts/helpers/env_helpers.py
TheDoubleK: I'm not sure you'll find the www anywhere in environment to be honest, I don't think we have any other scripts that need to know about the website itself - they generally do work with APIs to batch manipulate data.

TheDoubleK: It feels like it could live in the email template maybe?  That's the only thing that needs to know the web domain.  So the template could have the various www locations listed?  Or is that weird?

tests/test_notify_suppliers_of_new_questions_answers.py
TheDoubleK: It's probably fine as it is.  I was just taken aback by the ALL_CAPS.  It does feel like we shouldn't have to hack these things just to keep linters happy, but that's where we're at it seems.

dmscripts/notify_suppliers_of_new_questions_answers.py
CrystalPea: We can take it up with content, we followed the subject line from them. 'Question and answer' is a name of the feature as a whole I think.

dmscripts/notify_suppliers_of_new_questions_answers.py
CrystalPea: Yep, it is dealing with duplicates indeed.

dmscripts/notify_suppliers_of_new_questions_answers.py
CrystalPea: the exclusion would be a super rare event, I am not sure if it's worth mentioning it everyday, but then I'm but a rookie 😝. What do you think @idavidmcdonald @katstevens ?

scripts/notify-suppliers-of-new-questions-answers.py
CrystalPea: We actually planned to use notify for this script and we decided to move to mandrill midway through because of how email to each supplier group of addresses would have different number of briefs. There were also big problems with formatting. @idavidmcdonald and @pcraig3 can tell you more about this.

As to input of the excluded supplier ids into command line you have a point there. I din't know the technicalities around passing in a file though.

email_templates/suppliers_new_briefs_questions_answers.html
CrystalPea: This is a content decision, not really decided by us, should we go to content with that?

dmscripts/helpers/env_helpers.py
katstevens: I think the idea was that the email could be sent on any environment and still link to the right domain (e.g. during QA on staging). However there are other email templates that use hardcoded URLs, so this may not be a huge problem. Or do you mean determining the URL with something like `{% if stage == "preview" %}` in the template itself?

dmscripts/notify_suppliers_of_new_questions_answers.py
katstevens: I believe we were following the pattern in the existing `dos_opportunities.html` email template. Happy to add the param in the function instead, at the risk of introducing inconsistency? (Or we could change the DOS email too?)

dmscripts/notify_suppliers_of_new_questions_answers.py
katstevens: Good call - will check with product about this.

dmscripts/notify_suppliers_of_new_questions_answers.py
katstevens: No harm in making it clear exactly what's going on 😄 I'll add a commit to change this.

scripts/notify-suppliers-of-new-questions-answers.py
TheDoubleK: Let's leave it as a command line arg I reckon.  Like you say, with luck it should never need to be used so not worth over-engineering it.

dmscripts/helpers/env_helpers.py
TheDoubleK: Yep I was thinking of something like `{% if stage == "preview" %}` in the template.

dmscripts/notify_suppliers_of_new_questions_answers.py
TheDoubleK: The comma is probably not worth mentioning, but I think the subject line is, it definitely seems odd to me.

dmscripts/notify_suppliers_of_new_questions_answers.py
TheDoubleK: Hmm, consistency is good, but so is doing something once rather than multiple times right next to each other.

So I'm tempted to say I'd change both to include the `utm_id` in the `create_context` rather than in the template.

TheDoubleK: Sorry :/

dmscripts/notify_suppliers_of_new_questions_answers.py
idavidmcdonald: @tracy-content Kev was wondering about the subject line for this email. Do you have a minute to double check it please?

dmscripts/notify_suppliers_of_new_questions_answers.py
idavidmcdonald: I'd be tempted to pass the whole brief in as the context. Then we can build the URL in the frontend entirely. It feels weird to have `utm` stuff in the backend. However, this wouldn't get rid of the duplication (unless you do something smart like set a jinja variable to be the url and then populate using it). Thoughts?

email_templates/suppliers_new_briefs_questions_answers.html
idavidmcdonald: @tracy-content Another content question for you please :)

scripts/notify-suppliers-of-new-questions-answers.py
idavidmcdonald: Yeah, I hope it will never happen. If it crops up more than two or three times we can come up with an easier solution.

dmscripts/helpers/env_helpers.py
katstevens: How about splitting the difference and having a new `get_web_url_from_stage` function in `env_helpers`? As we're now going to append the `utm` param in the code, it makes sense to me to generate the whole URL first then pass it to the template.

katstevens:
Re: scheduling: we've agreed to schedule the job at 9.01am to avoid any potential duplication (or wrinkles around daylight savings). Regards the api load - the script makes three queries via the api client. I'm not sure how expensive each one will be, but we can keep an eye on it and potentially move the times back a couple of hours if necessary?

dmscripts/helpers/env_helpers.py
katstevens: We've moved the construction of the URL entirely (including the UTM param) to the Python code so it's all in one place. This should keep the template more readable, considering there are potentially 4 different domains to choose from.

dmscripts/notify_suppliers_of_new_questions_answers.py
tracy-content: @idavidmcdonald looking at it again I think Kev has a point. 

So, can I ask for one change to this story please? Can we change the email subject line.

From:
New question and answer responses on the Digital Marketplace

To:
New answers to supplier questions on the Digital Marketplace

It feels like maybe we need to clarify what answers, so I've included 'to supplier questions'. It might not be necessary, but I'm including it for clarity.

CrystalPea: Ok, I will do that now @tracy-content ☺️ .

email_templates/suppliers_new_briefs_questions_answers.html
CrystalPea: I just talked with @tracy-content about this and she said that the comma is fine. @TheDoubleK and @idavidmcdonald could you confirm if this is ready for merging, then? Or some more stuff to address?

scripts/notify-suppliers-of-new-questions-answers.py
```diff
@@ -0,0 +1,44 @@
...
+    ok = main(
+        data_api_url=get_api_endpoint_from_stage(arguments['<stage>'], 'api'),
+        data_api_token=arguments['<api_token>'],
+        email_api_key=arguments['<mandrill_api_key>'],
+        stage=arguments['<stage>'],
+        dry_run=arguments['--dry-run'],
+        exclude_supplier_ids=list_of_excluded_supplier_ids
+    )
+
+    if not ok:
```
TheDoubleK: As `main()` just returns `True` I'm not sure how we'd ever end up returning an error?

dmscripts/notify_suppliers_of_new_questions_answers.py
```diff
@@ -0,0 +1,154 @@
...
+    interested_suppliers = get_ids_of_interested_suppliers_for_briefs(data_api_client, briefs)
+    interested_suppliers = dict(
+        (supplier_id, briefs) for supplier_id, briefs in interested_suppliers.items()
+        if supplier_id not in exclude_supplier_ids
+    )
+    logger.info(
+        "{} suppliers found interested in these briefs who are not excluded".format(len(interested_suppliers))
+    )
+
+    for supplier_id, brief_ids in interested_suppliers.items():
```
TheDoubleK: Is it worth having a `try.... except` here to continue if, e.g. there's a failed API call or failed email send during the run?  Might be worth continuing to try rather than dying altogether. And then the `except` part could keep a count of failures, and this could be the number returned by the function.

And then a `0` return value means everything was OK, and anything non-zero should trigger a failure exit code (to flag a failure to the Jenkins job).

TheDoubleK:
OK I left a final comment about exit codes and catching errors during the run.  It's not a massive deal-breaker,m but I do think we need to be able to ensure that the Jenkins build will be marked red and ping Slack if any of the emails fail to send, and I'm not sure that will happen with the current setup (i.e. I'm not 100% sure if, e.g. an API error halting the script, would trigger `not ok` and cause the script to return 1).

TheDoubleK:
So on reflection, of course this will fail red if it dies with an uncaught error.

The question is whether we want to give up immediately there's a single failure or if it might be worth continuing with the rest of the job.

That's up to you really so I'm going to approve this now, although it feels to me like we might not want to give up for a single failed request/email.

There's an example of one of our scripts counting errors and returning the count here: https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/upload-counterpart-agreements.py#L79

TheDoubleK:
I did think of suggesting passing in IDs to send to rather than exclude but didn't have the heart at this stage :)

Better for trying things out too - you can just send to yourself rather than everyone.

PR #133, opened 2017-08-04
Add script to programatically update an ES alias

Wynndow:
Also see this PR on the Jenkins repo: [https://github.com/alphagov/digitalmarketplace-jenkins/pull/18](https://github.com/alphagov/digitalmarketplace-jenkins/pull/18)

Currently adding or moving an elasticsearch index alias is a manual job
involving constructing a bespoke request with curl.

You can do the same thing by giving this script the arguments required.
It will fetch the correct auth token for the search api from the
credentials repo.

It's being written for use by Jenkins, but could be used locally.

If invoked with `--delete-old-index=yes` it will remove whichever index
is losing the `<alias>-old`alias.

### DIFF
--- a/dmscripts/helpers/auth_helpers.py
+++ b/dmscripts/helpers/auth_helpers.py
@@ -1,0 +1,17 @@
+import os
+import subprocess
+import yaml
+
+
+def get_auth_token(api, stage):
+    DM_CREDENTIALS_REPO = os.environ.get('DM_CREDENTIALS_REPO')
+    token_prefix = 'J' if subprocess.check_output(["whoami"]) == 'jenkins' else 'D'
+    creds = subprocess.check_output([
+        "{}/sops-wrapper".format(DM_CREDENTIALS_REPO),
+        "-d",
+        "{}/vars/{}.yaml".format(DM_CREDENTIALS_REPO, stage)
+    ])
+    auth_tokens = yaml.load(creds)[api]['auth_tokens']
+    auth_token = [token for token in auth_tokens if token.startswith(token_prefix)]
+
+    return auth_token[0]
--- a/scripts/update-index-alias.py
+++ b/scripts/update-index-alias.py
@@ -21,26 +21,19 @@ from docopt import docopt
 
 
 def update_index_alias(alias, target, stage, endpoint, delete_old_indexes):
-    url = "https://{}/{}".format(endpoint, alias)
     auth_token = _get_auth_token(stage)
     headers = {
         'Authorization': "Bearer {}".format(auth_token),
         'Content-type': 'application/json'
     }
+
+    _rename_current_alias_to_old(alias, endpoint, headers)
+
+    url = "https://{}/{}".format(endpoint, alias)
     data = {'type': 'alias', 'target': "{}".format(target)}
 
     response = requests.put(url, headers=headers, json=data)
-
-    try:
-        response.raise_for_status()
-    except HTTPError as e:
-        print("HTTPError updating alias: {}".format(e.args[0]))
-        sys.exit(1)
-    except Exception as e:
-        print("Error updating alias: {}".format(e))
-        sys.exit(2)
-
-    print('Successfully updated alias')
+    _check_response_status(response, 'updating alias')
 
     if delete_old_indexes == 'yes':
         _delete_old_indexes(endpoint, auth_token)
@@ -58,45 +51,54 @@ def _get_auth_token(stage):
     return dev_token[0]
 
 
+def _rename_current_alias_to_old(alias, endpoint, headers):
+    old_alias = "old-{}".format(alias)
+    url = "https://{}/{}".format(endpoint, old_alias)
+    data = {'type': 'alias', 'target': "{}".format(alias)}
+
+    response = requests.put(url, headers=headers, json=data)
+
+    _check_response_status(response, 'renaming old alias')
+
+
 def _delete_old_indexes(endpoint, auth_token):
     status_url = "https://{}/_status".format(endpoint.replace('search-api', 'www'))
-    response = requests.get(status_url)
 
-    try:
-        response.raise_for_status()
-    except HTTPError as e:
-        print("HTTPError fetching indexes: {}".format(e.args[0]))
-        sys.exit(1)
-    except Exception as e:
-        print("Error fetching indexes: {}".format(e))
-        sys.exit(2)
+    response = requests.get(status_url)
+    _check_response_status(response, 'fetching indexes')
 
     current_indexes = json.loads(response.content)['search_api_status']['es_status']
     indexes_to_delete = [
         index for index in current_indexes.keys() if not current_indexes[index]['aliases']
     ]
 
-    if len(indexes_to_delete) >= len(current_indexes):
-        print('Error: Can not delete all indexes')
+    if len(current_indexes) - len(indexes_to_delete) < 2:
+        print('Error: Deleting too many indexes')
         sys.exit(3)
 
     base_url = "https://{}/{}"
     headers = {
         'Authorization': "Bearer {}".format(auth_token),
     }
+
     for index in indexes_to_delete:
         response = requests.delete(base_url.format(endpoint, index), headers=headers)
+        _check_response_status(response, "deleting unaliased index {}".format(index))
 
-        try:
-            response.raise_for_status()
-        except HTTPError as e:
-            print("HTTPError deleting index {}: {}".format(index, e.args[0]))
-            sys.exit(1)
-        except Exception as e:
-            print("Error deleting index {}: {}".format(index, e))
-            sys.exit(2)
+    print('Completed deleting unaliased indexes')
 
-    print('Completed deleting indexes')
+
+def _check_response_status(response, action):
+    try:
+        response.raise_for_status()
+    except HTTPError as e:
+        print("HTTPError {}: {}".format(e.args[0], action))
+        sys.exit(1)
+    except Exception as e:
+        print("Error {}: {}".format(e), action)
+        sys.exit(2)
+
+    print("Success {}".format(action))
 
 
 if __name__ == "__main__":
--- a/scripts/update-index-alias.py
+++ b/scripts/update-index-alias.py
@@ -8,11 +8,11 @@ access the credentials repo.
 To delete the old index (the index losing the '<alias>-old' alias), set --delete-old-index=yes
 
 Usage:
-scripts/update-index-alias.py <alias> <target> <search-api-endpoint> [--stage=<stage> --delete-old-index=<value>]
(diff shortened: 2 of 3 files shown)


scripts/update-index-alias.py
```diff
@@ -0,0 +1,107 @@
...
+    DM_CREDENTIALS_REPO = os.environ.get('DM_CREDENTIALS_REPO')
+    creds = subprocess.check_output([
+        "{}/sops-wrapper".format(DM_CREDENTIALS_REPO),
+        "-d",
+        "{}/vars/{}.yaml".format(DM_CREDENTIALS_REPO, stage)
+    ])
+    return yaml.load(creds)['search_api']['auth_tokens'][2]
+
+
+def _delete_old_indexes(endpoint, auth_token):
```
samuelhwilliams: I'm not very comfortable with it deleting all old indexes, particularly if this will become part of the Jenkins job. I would be far more comfortable with it having an option to delete just the index that has been replaced, although I suppose this is arguably the most useful index to keep.

Should we instead introduce something like a `g-cloud-old` alias? This alias would be applied to the current `g-cloud` aliased index, and then when the script runs, it shuffles indexes back along this pathway. I.e. the new index takes the `g-cloud` alias, the current `g-cloud` index takes the `g-cloud-old` alias, and then the script deletes the index that was covered by the `g-cloud-old` alias.

This would meet the function of cycling out old indexes, but give us the ability to quickly restore the last version of the index in case of mistakes or problems.

scripts/update-index-alias.py
```diff
@@ -0,0 +1,107 @@
...
+"""
+A script to update the search index an alias is assoiciated with.
+
+To delete old indexes (those with no alias), set --delete-old-indexes=yes
+
+Usage:
+scripts/update-index-alias.py <alias> <target> <stage> <search-api-endpoint> [options]
+
+Options:
+    --delete-old-indexes=<delete-old-indexes>  [default: ]
```
Wynndow: Yep. I was trying to get the default to be something 'falsey'. This default gives an empty string which is.
Problem is, if you set it to anything else, like 'no' for example, it's a truthy value. That's why it explicitly checks for 'yes' as a string literal in the code. Perhaps there's a better way of doing it, but setting the default to 'no' won't hurt.

scripts/update-index-alias.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def _get_auth_token(stage):
+    DM_CREDENTIALS_REPO = os.environ.get('DM_CREDENTIALS_REPO')
+    creds = subprocess.check_output([
+        "{}/sops-wrapper".format(DM_CREDENTIALS_REPO),
+        "-d",
+        "{}/vars/{}.yaml".format(DM_CREDENTIALS_REPO, stage)
+    ])
+    return yaml.load(creds)['search_api']['auth_tokens'][2]
```
Wynndow: Yep, I'll update

scripts/update-index-alias.py
Wynndow: I think this is a good idea, i'll get it updated.

Wynndow: Have updated. The script will now add the alias `old-g-cloud` to the index that currently has the alias. It then moves the `g-cloud` alias to the target index. It then deletes all indexes with no alias.

scripts/update-index-alias.py
samuelhwilliams: Oh, I see, that's a specific piece of syntax for docopt. Meh. Have you used argparse? I generally prefer it for writing scripts, but it does preclude having the syntax documentation at the top of the script.

scripts/update-index-alias.py
samuelhwilliams: I still feel like deleting all old/un-aliased indexes probably falls outside the remit of this script, which at its core is only about switching over index aliases. If/When we start having more things indexed than just the G-Cloud catalogue, this could be a pretty destructive action with unintended/unanticipated consequences.

I'd rather have to manually delete errant indexes one-by-one (which are highly unlikely to occur in production, anyway) than have a minesweeper go over them all regularly and delete indiscriminately. I'm OK with this script deleting the one oldest index relevant to the current target index (in this instance G-Cloud), but it should probably leave everything else.

scripts/update-index-alias.py
```diff
@@ -0,0 +1,111 @@
...
+        "-d",
+        "{}/vars/{}.yaml".format(DM_CREDENTIALS_REPO, stage)
+    ])
+    auth_tokens = yaml.load(creds)['search_api']['auth_tokens']
+    dev_token = [token for token in auth_tokens if token[0] == 'D']
+    return dev_token[0]
+
+
+def _rename_current_alias_to_old(alias, endpoint, headers):
+    old_alias = "old-{}".format(alias)
```
samuelhwilliams: I'd probably prefer to have it called `"{}-old"` so that it appears in the same place as the current index. Only really matters when we start thinking about having lots more frameworks/catalogues indexed, but I think in that situation they'd make more sense to keep together.

scripts/update-index-alias.py
```diff
@@ -0,0 +1,111 @@
...
+
+    _rename_current_alias_to_old(alias, endpoint, headers)
+
+    url = "https://{}/{}".format(endpoint, alias)
+    data = {'type': 'alias', 'target': "{}".format(target)}
+
+    response = requests.put(url, headers=headers, json=data)
+    _check_response_status(response, 'updating alias')
+
+    if delete_old_indexes == 'yes':
```
samuelhwilliams: I think it would be nice to be a bit more forgiving here. Maybe use `if strtobool(delete_old_indexes)` using `from distutils.util import strtobool`. Alternatively use argparse with that function as the type (my preferred), but either/or would be good.

Wynndow: Nice. This was what I was looking for. Have implemented.

scripts/update-index-alias.py
Wynndow: Yep, makes sense, done.

scripts/update-index-alias.py
Wynndow: This seems most sensible. I've updated so it will only delete the index which is losing the `<alias>-old` alias.

scripts/update-index-alias.py
```diff
@@ -0,0 +1,111 @@
...
+    auth_token = _get_auth_token(stage)
+    headers = {
+        'Authorization': "Bearer {}".format(auth_token),
+        'Content-type': 'application/json'
+    }
+
+    if delete_old_index:
+        old_index_to_delete = _get_old_index_to_delete(alias, endpoint)
+
+    _rename_current_alias_to_old(alias, endpoint, headers)
```
allait: The name is a bit misleading, as far as I can tell it's not renaming the current alias, it's *moving* the old one to match it.

scripts/update-index-alias.py
```diff
@@ -0,0 +1,111 @@
...
+
+    if delete_old_index:
+        _delete_old_index(old_index_to_delete, endpoint, auth_token)
```
allait: So if this step fails on first run and interrupts the script, then we're left with `alias-old == alias`. If the script is run again it will try to move `alias-old` to match `alias` (which it already does) and then retry the delete, which would fail (since search-api doesn't allow deleting indexes with an alias). The only way to resolve this would be manual intervention to move the aliases and clean up the old index.

I suggest we reorder the steps:
1. Store the currently aliased index name
2. Store the old aliased index name
3. Move the alias to the new index
4. Move alias-old to the previously aliased index
5. Delete previous target of alias-old

This way we should be able to rerun the job after failure and the worst case is that it will leave some old indexes around.

scripts/update-index-alias.py
```diff
@@ -101,6 +81,18 @@ def _check_response_status(response, action):
...
 
+def _get_auth_token(stage):
+    DM_CREDENTIALS_REPO = os.environ.get('DM_CREDENTIALS_REPO')
+    creds = subprocess.check_output([
+    "{}/sops-wrapper".format(DM_CREDENTIALS_REPO),
+    "-d",
+    "{}/vars/{}.yaml".format(DM_CREDENTIALS_REPO, stage)
+    ])
+    auth_tokens = yaml.load(creds)['search_api']['auth_tokens']
+    dev_token = [token for token in auth_tokens if token[0] == 'D']
```
samuelhwilliams: `token.startswith('D')` might be a clearer syntax to use.

samuelhwilliams:
I'd like to be able to use this on my local environment (just tried to as part of testing), but it kicks back because creds repo doesn't have a file for local tokens/etc. Can we add a check for stage==dev/local/etc and pre-populate the `myToken` tokens? And/or add optional parameters to supply the tokens as command line args. I guess my most preferred would be for stage to be optional (defaulting to dev), and auth tokens similarly optional and defaulting to `myToken`), so that it can be easily run against local environments (which is where most manual use will take place, I suspect).

scripts/update-index-alias.py
Wynndow: nice.

Wynndow:
Yeah, I'll add this in. I like the optional stage, defaulting to dev thing.

scripts/update-index-alias.py
Wynndow: Cool, all fixed

scripts/update-index-alias.py
```diff
@@ -0,0 +1,107 @@
...
+        print("HTTPError {}: {}".format(e.args[0], action))
+        sys.exit(1)
+    except Exception as e:
+        print("Error {}: {}".format(e), action)
+        sys.exit(2)
+
+    print("Success {}".format(action))
+
+
+def _get_auth_token(stage):
```
samuelhwilliams: It would actually be quite to have this in dmscripts/heplers actually, now that I think about it.

dmscripts/helpers/auth_helpers.py
```diff
@@ -0,0 +1,16 @@
...
+
+def get_dev_auth_token(api, stage):
+    DM_CREDENTIALS_REPO = os.environ.get('DM_CREDENTIALS_REPO')
+    creds = subprocess.check_output([
+        "{}/sops-wrapper".format(DM_CREDENTIALS_REPO),
+        "-d",
+        "{}/vars/{}.yaml".format(DM_CREDENTIALS_REPO, stage)
+    ])
+    auth_tokens = yaml.load(creds)[api]['auth_tokens']
+    dev_token = [token for token in auth_tokens if token.startswith('D')]
```
TheDoubleK: If this is running on Jenkins it should use the Jenkins token beginning with 'j', surely?

Wynndow: Hey Kev! That's a good point. I could update it to check the user who's running the script and pull the correct token. 

Not sure I'd actually made the connection that we had a 'J' token that was for Jenkins. Obvious when you think about it....

PR #134, opened 2017-08-09
Add a script to find a random user account

allait:
Outputs details of a random active unlocked user with a given role.

For supplier users, providing additional <framework> and <lot> arguments allows
filtering by suppliers who have published services on the given framewok and lot.

### DIFF
--- a/scripts/get-user.py
+++ b/scripts/get-user.py
@@ -1,0 +1,98 @@
+#!/usr/bin/env python
+"""Outputs details of a random active unlocked user with a given role.
+
+For supplier users, providing additional <framework> and <lot> arguments allows
+filtering by suppliers who have published services on the given framewok and lot.
+
+Usage:
+    get-user.py <role> [<framework>] [<lot>] [options]
+
+    --api-url=<api_url>  API URL [default: http://localhost:5000]
+    --api-token=<api_token>  API token [default: myToken]
+
+Example:
+    ./get-user.py supplier g9 cloud-hosting
+    ./get-user.py --api-url=localhost:5000 --api-token=myToken admin
+    ./get-user.py --api-url=localhost:5000 --api-token=myToken supplier g-cloud-9
+    ./get-user.py --api-url=localhost:5000 --api-token=myToken supplier dos2 digital-outcomes
+"""
+
+from __future__ import print_function
+
+import random
+import re
+import sys
+
+import dmapiclient
+from docopt import docopt
+
+sys.path.insert(0, '.')  # noqa
+
+
+def get_full_framework_slug(framework):
+    iteration = re.search('(\d+)', framework)
+    if framework.startswith('g'):
+        prefix = 'g-cloud'
+    elif framework.startswith('d'):
+        prefix = 'digital-outcomes-and-specialists'
+    else:
+        return framework
+
+    if iteration:
+        return "{}-{}".format(prefix, iteration.group(1))
+    else:
+        return prefix
+
+
+def get_supplier_id(api_client, framework, lot):
+    services = [
+        s for s in api_client.find_services(framework=framework, lot=lot)['services']
+        if s['status'] in ['published']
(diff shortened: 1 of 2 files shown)


scripts/get-user.py
```diff
@@ -0,0 +1,90 @@
...
+    return random.choice([
+        u for u in api_client.find_users(role=role, supplier_id=supplier_id)['users']
+        if u['active'] and not u['locked']
+    ])
+
+
+def get_user(api_url, api_token, role, framework, lot):
+    api_client = dmapiclient.DataAPIClient(api_url, api_token)
+
+    if role == 'supplier':
```
samuelhwilliams: Be nice to force this to lowercase, like you do with the frameworks check. Probably better to do in the `__main__` block mind.

scripts/get-user.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python
+"""Outputs details of a random active unlocked user with a given role.
+
+For supplier users, providing additional <framework> and <lot> arguments allows
+filtering by suppliers who have published services on the given framewok and lot.
```
samuelhwilliams: This implies the framework/lot args are optional. If you ask for a supplier, the script dies if you don't provide at least a framework.

allait:
@TheDoubleK oh, I haven't seen the README list. Do people actually use it? It looks like it's the same information (apart from flagging things which run on Jenkins) you get from listing the files in the scripts/ directory and getting `--help` from the individual scripts.

It feels to me like it'd be valuable if it described common tasks and which scripts relate to what for example (and which order to run them in), but as is I'm not really sure what value it adds. I think everyone should be comfortable looking at the file list and getting individual script's help and if they're not I'd rather we write documentation for that.

But I'm happy to add it if people do find the README useful.

scripts/get-user.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def get_user(api_url, api_token, role, framework, lot):
+    api_client = dmapiclient.DataAPIClient(api_url, api_token)
+
+    if role == 'supplier':
+        framework = get_full_framework_slug(framework)
+        print('Framework: {}'.format(framework))
+        supplier_id = get_supplier_id(api_client, framework, lot)
+        print('Supplier id: {}'.format(supplier_id))
```
samuelhwilliams: Printing out framework+supplier id, but no the lot (if given). Be nice to spit everything back out just as confirmation.

scripts/get-user.py
```diff
@@ -0,0 +1,90 @@
...
+    ])
+
+
+def get_user(api_url, api_token, role, framework, lot):
+    api_client = dmapiclient.DataAPIClient(api_url, api_token)
+
+    if role == 'supplier':
+        framework = get_full_framework_slug(framework)
+        print('Framework: {}'.format(framework))
+        supplier_id = get_supplier_id(api_client, framework, lot)
```
samuelhwilliams: If the lot is invalid, eg `cloud-blah` for g9, it dies with a fairly obtuse error message. Be nice to catch this and make it clearer what the problem is. It also dies if you get the case wrong, eg `Cloud-hosting`.

scripts/get-user.py
```diff
@@ -0,0 +1,90 @@
...
+
+if __name__ == "__main__":
+    arguments = docopt(__doc__)
+    user = get_user(
+        api_url=arguments['--api-url'],
+        api_token=arguments['--api-token'],
+        role=arguments['<role>'],
+        framework=arguments['<framework>'],
+        lot=arguments['<lot>'],
+    )
```
samuelhwilliams: Probably all of these args should be forced to lowercase, just to prevent issues later on (as mentioned above).

scripts/get-user.py
allait: Good shout, made them optional so now you can ask for a completely random supplier user as well.

PR #135, opened 2017-08-10
Add a description of send_dos_opportunities_email.py script to README

CrystalPea:
Script with description still missing:

notify-suppliers-whether-application-made-for-framework.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,10 @@ The `scripts` folder in this repository contains scripts that interact with the 
 
   Checks all free-text fields in G-Cloud services for "bad words" and generates a CSV report of any bad words found.
 
+* `send_dos_opportunities_email.py`
+
+  **Runs Mon-Fri 8:00am on Jenkins**. Send emails with new opportunities on a lot to suppliers who are on this lot.
+
 * `send-stats-to-performance-platform.py`
 
   Fetches application statistics for a framework from the API and pushes them to the Performance Platform.
--- a/dmscripts/notify_suppliers_of_new_questions_answers.py
+++ b/dmscripts/notify_suppliers_of_new_questions_answers.py
@@ -147,6 +147,14 @@ def main(data_api_url, data_api_token, email_api_key, stage, dry_run, supplier_i
                     'brief_ids_list': ", ".join(map(str, brief_ids))
                 }
             )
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -101,6 +101,10 @@ The `scripts` folder in this repository contains scripts that interact with the
 
   Checks all free-text fields in G-Cloud services for "bad words" and generates a CSV report of any bad words found.
 
+* `send_dos_opportunities_email.py`
```
samuelhwilliams: This script uses underline notation; all other scripts in this repo appear to use dashes. Could we update the script (assuming this isn't just a typo) to conform to that standard, please?

README.md
```diff
@@ -101,6 +101,10 @@ The `scripts` folder in this repository contains scripts that interact with the
...
+
+  **Runs Mon-Fri on Jenkins**. Send emails with new opportunities on a lot to suppliers who are on this lot.
```
samuelhwilliams: Might be nice to include the time it runs (I think it's something like 8am or 9am?).

README.md
CrystalPea: @samuelhwilliams you are totally right that this name is not consonant with the rest, but if we change it, we also need to change the Jenkins job and this probably is outside of scope of this manual update. Should we add it to the tech debt?

README.md
CrystalPea: Good call, I will add the runtime ☺️

README.md
samuelhwilliams: If you don't think it is a quick fix, then yes we should make a record of it somewhere. Tech debt is probably the most appropriate place to put it. 👍

CrystalPea: Ok, added to tech debt: https://trello.com/c/3Ca1I5m6/152-change-filename-of-senddosopportunitiesemailpy-in-scripts

MuriloDalRi:
I think we can merge this, no?

CrystalPea:
I think you are right @MuriloDalRi, should I do it now?

MuriloDalRi:
You were not here on Monday when we had a tech retro. We agreed to either merge or close PRs older than two weeks since 4th wall was getting really cluttered.

CrystalPea:
ok, it is merged now :)

PR #138, opened 2017-08-17
Add a way to build a Docker image for running scripts

allait:
We've been running scripts locally and on Jenkins using virtualenv. This has been working mostly fine so far but has a few downsides:

* We're not explicitly managing the python version. This lead us to use pyenv on Jenkins because Ubuntu 14.04 shipped with an old Python 2 version without proper SSL certificate verification support. This also creates problem for python 3 since in order to switch individual scripts to run with 3.6 instead of 2.7 we'd need to pin the python version in each Jenkins job explicitly. Switching the python version also requires deleting existing venv folders from the job workspace, which would have to be done manually.
* Some of the scripts rely on additional tools / packages (like wkhtmltopdf) which have to be separately installed locally and on Jenkins. Their version is managed differently in different environments and there's no way to express that dependency in this repo.

Building a docker image with the scripts allows us to specify a Python version, upgrade existing Jenkins jobs to Python 3 without managing a python interpreter with `pyenv` and potentially makes it much easier to deploy and run scripts on PaaS (since we already have a deployment pipeline for docker images).

This adds Dockerfile, .dockerignore and make *-docker targets. Since scripts requirements are quite different from the application ones the Docker image is based on the python one, not our own base image.

We're only building a Python 3 image since we've agreed that we want to make scripts python3-only in the future and jobs that aren't compatible yet can use the existing `pyenv` setup for now.

This is also only setting up the python packages it doesn't install any script-specific dependencies (like wkhtmltopdf). Those will have to be added later once the scripts have been tested under python3.

### Add freeze-requirements

At the moment, scripts running in Jenkins are using the local venv created by the Jenkins job, which means they won't update any packages unless a version is changed explicitly.

This means that recreating that venv could end up with a different set of package versions than what's currently installed.

We need to freeze our nested package requirements to make sure we're not accidentally updating packages between builds.

Similar to the application changes, this renames requirements.txt to requirements-app.txt. Updates Makefile to add freeze-requirements.

requirements-app.txt is not a great name for the scripts repo, but it's consistent with what we call it in other repos.

This also switches Travis to use `make test` target instead of running py.test directly. `make requirements` is setting up dependencies in the virtualenv, but running `py.test` directly isn't activating that virtualenv explicitly. The only reason this worked on Travis is because it's activating a clean virtualenv before running any steps for the python builds.

**NOTE:** This is upgrading dmutils to bring a newer version of the cryptography package in Python 3. There's a major change related to S3 in the new utils version, so scripts using it might need to be updated.


README.md
```diff
@@ -127,6 +127,33 @@ The `scripts` folder in this repository contains scripts that interact with the
...
+```
+
+`docker run digitalmarketplace/scripts` without an explicit command will display the Python version,
+release tag the container was built from and a list of available scripts.
+
+If the script is connecting to any local apps/services you need to forward the ports to the docker
+container. The easiest way to do this is to use `--net=host` argument for `docker run`:
+
+```
+docker run --net=host digitalmarketplace/scripts index-services.py dev ...
```
Wynndow: Tiny change, should be `scripts/index-services.py dev ...`

allait: Oops. I've added `scripts/` to PATH in one of the image iterations, but decided against it in the end. Fixed now, good catch!

TheDoubleK:
The Dockerfile is all a bit dark arts to me, but everything looks sensible as far as I can tell.

Were we going to add some warnings to scripts that we suspect might not work properly after this update as part of the PR? If not then :+1:

PR #140, opened 2017-08-22
758-2-add-contract-info-to-data-script

benvand:
https://trello.com/c/ohWdLfJO/758-2-add-contract-info-to-data-script

Add new method `expand_json_subfields` to handle incoming json and add fields to the parent object
 * Add `awardedContractStartDate` to parent `brief_response`
 * Add `awardedContractValue` to parent `brief_response`

Add new fields to base model csvs
 * `briefs.csv` `awardedBriefResponseId`
 * `brief_responses.csv` `id,supplierOrganisationSize,awardedContractStartDate,awardedContractValue`

Test

### DIFF
--- a/dmscripts/models/queries.py
+++ b/dmscripts/models/queries.py
@@ -64,3 +64,10 @@ def add_counts(join, group_by, model, data, directory):
         )['_count'].fillna(0)
 
     return data
+
+
+def assign_json_subfields(field, subfields, data):
+    """Apply subfields from field to data."""
+    for subfield in subfields:
+        data[subfield] = data.apply(lambda row: row[field].get(subfield, '') if row[field] else '', axis=1)
+    return data
--- a/scripts/get-model-data.py
+++ b/scripts/get-model-data.py
@@ -145,7 +145,8 @@ CONFIGS = [
             'budgetRange',
             'startDate',
             'contractLength',
-            'isACopy'
+            'isACopy',
+            'awardedBriefResponseId'
         ),
         'get_data_kwargs': {'with_users': 'true'},
         'process_fields': {
@@ -165,8 +166,13 @@ CONFIGS = [
             'createdAt',
             'supplierName',
             'submittedAt',
-            'essentialRequirements'
+            'essentialRequirements',
+            'id',
+            'awardedContractStartDate',
+            'awardedContractValue',
+            'supplierOrganisationSize'
         ),
+        'assign_json_subfields': {'awardDetails': ['awardedContractStartDate', 'awardedContractValue']},
         'get_data_kwargs': {},
         'process_fields': {
             'createdAt': format_datetime_string_as_date,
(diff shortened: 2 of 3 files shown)


dmscripts/models/queries.py
```diff
@@ -64,3 +64,10 @@ def add_counts(join, group_by, model, data, directory):
         )['_count'].fillna(0)
 
     return data
+
+
+def expand_json_subfields(field, subfields, data):
```
CrystalPea: I like this, very useful for further expansion of data collection 👍

dmscripts/models/queries.py
```diff
@@ -64,3 +64,10 @@ def add_counts(join, group_by, model, data, directory):
...
+    """Apply subfields from field to data."""
+    for subfield in subfields:
+        data[subfield] = data.apply(lambda row: row[field].get(subfield, '') if row[field] else '', axis=1)
+    return data
```
samuelhwilliams: Also, this seems unnecessary. The data is mutated in-place, so returning data and reassigning it to data is redundant, I think?

dmscripts/models/queries.py
katstevens: How about `flatten_json_subfields`?

katstevens:
Probably out of scope for this PR, but it would be amazing to pull the big chunk of code out of the top level `get-model-data.py` script and into the `dmscripts/` folder so we could test it properly? Otherwise looks good to me.

dmscripts/models/queries.py
benvand: Done for consistency with other methods in the file.
https://google.github.io/styleguide/pyguide.html

dmscripts/models/queries.py
benvand: Flatten sounds about right in this instance
https://stackoverflow.com/questions/20957023/flattening-a-binary-tree-to-an-array-in-java

benvand: Done

samuelhwilliams: I don't think flatten is the right word - it suggests you're actually moving them, rather than adding them to the parent and leaving them where they are.

benvand: `assign_json_subfields` ?

benvand: `denormalize_json_subfields` ?

samuelhwilliams: `copy_json_subfields_to_parent` :trollface:

benvand: But we don't refer to data as 'parent' anywhere else in either file so then the logical thing to do is to call it `copy_json_subfields_to_parent` but we don't _refer_ to data in any of the other transformation method names.
`copy_json_subfields_to_parent` Sounds overly verbose compared with everything else in that file.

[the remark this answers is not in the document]
benvand: Alrighty. @katstevens @samuelhwilliams Thanks for the reviews. I'm gonna go down the middle with `assign_json_subfields`. In the context of the file it feels like the right compromise between verbosity and semantic correction. Can I get an approval plz?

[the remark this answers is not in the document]
katstevens: Yep, that sounds fine to me - remember to change the call & the test name too 😄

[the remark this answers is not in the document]
samuelhwilliams: Sounds good @benvand. I already approved in my previous review so feel free to go ahead based on that. 👍

PR #141, opened 2017-08-23
Prepare scripts for move to Python3

katstevens:
Addresses some tech debt in the scripts code for Python3 compatibility (see https://trello.com/c/8VQwNhrN/158-scripts-running-in-python-3). The changes below were autogenerated by running `2to3`, so please sanity check!

- ensure print statements have parentheses
- clean up unicode strings
- remove unnecessary imports
- ensure iterator functions that returned a list in Python2 still return a list if it's needed (rather than e.g. a filter or zip object)
- converted any `map(None, ...)` statements to list comprehensions with conditionals
- use `isinstance(foo, collections.Callable)` instead of `callable(foo)`

Also extracted some common functionality into a helper for extra DRYness 🌵 

Tests pass in both Python2 and Python3. Most of the briefs scripts require a Mandrill key so I've only tested those scripts that have a `--dry-run` option.

### DIFF
--- a/dmscripts/generate_framework_master_csv.py
+++ b/dmscripts/generate_framework_master_csv.py
@@ -85,7 +85,7 @@ class GenerateMasterCSV(GenerateCSVFromAPI):
             ]
             # This creates placeholders for the dynamic lot fieldnames.
             lot_placeholders = [0 for i in self._get_dynamic_field_names()]
-            supplier_dict = dict(list(zip(field_names, supplier_info + lot_placeholders)))
+            supplier_dict = dict(zip(field_names, supplier_info + lot_placeholders))
             # Get service data and process dynamic lot values
             service_data = self.get_supplier_service_data(supplier_id)
             for service in service_data:
--- a/dmscripts/generate_framework_master_csv.py
+++ b/dmscripts/generate_framework_master_csv.py
@@ -85,7 +85,7 @@ class GenerateMasterCSV(GenerateCSVFromAPI):
             ]
             # This creates placeholders for the dynamic lot fieldnames.
             lot_placeholders = [0 for i in self._get_dynamic_field_names()]
-            supplier_dict = dict(zip(field_names, supplier_info + lot_placeholders))
+            supplier_dict = dict(list(zip(field_names, supplier_info + lot_placeholders)))
             # Get service data and process dynamic lot values
             service_data = self.get_supplier_service_data(supplier_id)
             for service in service_data:
--- a/dmscripts/helpers/logging_helpers.py
+++ b/dmscripts/helpers/logging_helpers.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import
+
 import sys
 import logging
 from logging import DEBUG, INFO, WARNING, ERROR, CRITICAL, getLogger  # noqa
--- a/dmscripts/helpers/logging_helpers.py
+++ b/dmscripts/helpers/logging_helpers.py
@@ -1,5 +1,3 @@
-from __future__ import absolute_import
-
 import sys
 import logging
 from logging import DEBUG, INFO, WARNING, ERROR, CRITICAL, getLogger  # noqa
--- a/dmscripts/notify_buyers_when_requirements_close.py
+++ b/dmscripts/notify_buyers_when_requirements_close.py
@@ -36,7 +36,7 @@ def notify_users(email_api_key, brief):
                 [user['emailAddress'] for user in brief['users'] if user['active']],
                 email_body,
                 email_api_key,
-                u'Next steps for your ‘{}’ requirements'.format(brief['title']),
+                u"Next steps for your '{}' requirements".format(brief['title']),
                 'enquiries@digitalmarketplace.service.gov.uk',
                 'Digital Marketplace Admin',
                 ['requirements-closed'],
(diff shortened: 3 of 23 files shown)


dmscripts/generate_buyer_email_list.py
```diff
@@ -13,7 +13,7 @@ def inner_function(user):
         try:
             briefs = client.find_briefs(user_id=user['id'])['briefs']
             if briefs:
-                title_statuses = [u"; ".join([u"{} - {}".format(brief['title'], brief['status']) for brief in briefs])]
```
allait: We should probably keep these until we switch the script using this to Py3. It's a stylistic change for Python 3, but it can break unicode strings in Python 2 environments.

dmscripts/generate_framework_master_csv.py
```diff
@@ -85,7 +85,7 @@ def _update_with_supplier_data(self, output):
             ]
             # This creates placeholders for the dynamic lot fieldnames.
             lot_placeholders = [0 for i in self._get_dynamic_field_names()]
-            supplier_dict = dict(zip(field_names, supplier_info + lot_placeholders))
+            supplier_dict = dict(list(zip(field_names, supplier_info + lot_placeholders)))
```
allait: No harm in adding this, but it was fine before as well since the iterator is consumed only once.

dmscripts/helpers/logging_helpers.py
```diff
@@ -1,5 +1,3 @@
-from __future__ import absolute_import
```
allait: Same as the `u''`. I'm not really sure what these are for in this case, but they make no difference for Python 3 so we might as well keep them until we're sure we don't need it.

dmscripts/helpers/service_faker_helpers.py
```diff
@@ -16,15 +16,15 @@ def eligible_g9_declaration_base():
...
+        "employersInsurance": "Not applicable - your organisation does not need employer’s liability insurance because"
+                              " your organisation employs only the owner or close family members. ",
         "publishContracts": True, "readUnderstoodGuidance": True, "understandTool": True,
         "understandHowToAskQuestions": True, "accurateInformation": True, "informationChanges": True,
         "accuratelyDescribed": True, "proofOfClaims": True,
         "nameOfOrganisation": "Mr Malachi Mulligan. Fertiliser and Incubator.",
         "tradingNames": "Omphalos dutiful yeoman services", "registeredAddressBuilding": "Omphalos",
         "registeredAddressTown": "Lambay Island", "registeredAddressPostcode": "N/A", "firstRegistered": "5/6/1904",
-        "currentRegisteredCountry": u"Éire", "companyRegistrationNumber": "00000014", "dunsNumber": "987654321",
+        "currentRegisteredCountry": "Éire", "companyRegistrationNumber": "00000014", "dunsNumber": "987654321",
```
allait: I'm surprised this didn't break in Py2 tests.

scripts/get-user.py
```diff
@@ -17,8 +17,6 @@
     ./get-user.py --api-url=localhost:5000 --api-token=myToken supplier dos2 digital-outcomes
 """
 
-from __future__ import print_function
```
allait: The script will definitely break on Py2 without it, but since it's only used by developers locally that's probably fine at this point.

dmscripts/helpers/service_faker_helpers.py
katstevens: Probably because the coverage is not good 😿

dmscripts/generate_buyer_email_list.py
katstevens: Cool - I'm happy to revert those.

dmscripts/generate_framework_master_csv.py
katstevens: Aha, good spot. `2to3` can be overly cautious when wrapping these, I did remove a few unnecessary ones (e.g. where the iterator was only used in a for loop) but missed this one.

scripts/get-user.py
katstevens: Really? I did try this one in Py2 and it seemed fine.

dmscripts/notify_buyers_when_requirements_close.py
```diff
@@ -36,7 +36,7 @@ def notify_users(email_api_key, brief):
                 [user['emailAddress'] for user in brief['users'] if user['active']],
                 email_body,
                 email_api_key,
-                u'Next steps for your ‘{}’ requirements'.format(brief['title']),
+                u"Next steps for your '{}' requirements".format(brief['title']),
```
allait: I think 2to3 replaced the curly quotes with straight ones inside the message. Let's restore them since they're nicer.

tests/test_notify_buyers_when_requirements_close.py
```diff
@@ -98,7 +98,7 @@ def test_notify_users(send_email):
             '/buyers/frameworks/framework-slug/requirements/lot-slug/100/responses'
         ),
         'KEY',
-        u'Next steps for your \u2018My brief title\u2019 requirements',
+        "Next steps for your 'My brief title' requirements",
```
allait: ... and revert the test for the curly quotes

dmscripts/notify_buyers_when_requirements_close.py
katstevens: Haha that wasn't 2to3, that was me 😸 I guess if we're reinstating the `u""` it doesn't matter whether they're curly or not - will change back.

scripts/get-user.py
allait: Oh yeah. I forgot it doesn't mind the brackets on their own.

PR #142, opened 2017-08-25
add migrate-supplier-data-from-declarations script

risicle:
Story https://trello.com/c/v0ouuEMF/61-migrate-the-data-from-the-g9-dos2-declaration-to-account-level-2

Most of the information about this is in that story, including example "dry run" output run on production yesterday with outputs of what changes it *would* make.

### DIFF
--- a/scripts/oneoff/migrate-supplier-data-from-declarations.py
+++ b/scripts/oneoff/migrate-supplier-data-from-declarations.py
@@ -2,17 +2,18 @@
 """Selectively migrate supplier information from a recent framework declaration to new fields on supplier itself
 
 Usage:
-    scripts/oneoff/migrate-supplier-data-from-declarations.py <stage> <data_api_token> <ch_api_token>
-        [<user>] [--frameworks <framework_slugs>] [--dry-run]
+    scripts/oneoff/migrate-supplier-data-from-declarations.py <stage> <data_api_token> <ch_api_token> <framework_slugs>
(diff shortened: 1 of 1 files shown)


scripts/oneoff/migrate-supplier-data-from-declarations.py
```diff
@@ -0,0 +1,200 @@
...
+
+    configure_logger({"script": loglevel_INFO})
+
+    client = DataAPIClient(get_api_endpoint_from_stage(arguments['<stage>']), arguments['<data_api_token>'])
+    user = arguments['<user>'] or getpass.getuser()
+    framework_slugs = arguments.get("--frameworks") and arguments["--frameworks"].split(",")
+    dry_run = bool(arguments.get("--dry-run"))
+
+    frameworks = tuple(
+        fw for fw in sorted(
```
TheDoubleK: I reckon if a list of frameworks was passed in on the command line then preserve the ordering there.  Or even just hard-code the framework list with the order we want in the script for this one-off use.

risicle:
@TheDoubleK regardez

TheDoubleK:
We could merge this now, but could wait for now in case there are other tweaks once we run it for real, and then only merge in the actual version that was run.  Up to you though - merging while it's all up-to-date might not be a bad thing :+1:

TheDoubleK:
This is the version of the one-off script that was run against production - merging it as-is for records.

PR #146, opened 2017-08-31
Fix script not to attempt to send emails when no active users

CrystalPea:


dmscripts/notify_suppliers_of_new_questions_answers.py
```diff
@@ -147,6 +147,14 @@ def main(data_api_url, data_api_token, email_api_key, stage, dry_run, supplier_i
                     'brief_ids_list': ", ".join(map(str, brief_ids))
                 }
             )
+        elif len(email_addresses) == 0:
```
idavidmcdonald: Two super minor things (don't feel obliged, they are more of a style thing, your code is all correct).

You don't need to check the length, you could instead use `if not email_addresses` as an empty list is considered `False` in terms of a boolean check.

You could change this so after line 141 when we fetch the `email_addresses`, we do the check then. This is because this supplier would still be included in a dry run even though in reality they would not be emailed. It also puts this bit of code nearer to the bit where we get the email addresses using `get_supplier_email_addresses_by_supplier_id` so related things are near each other.

Happy either way though.

idavidmcdonald:
Oh and your branch will need a rebase on top of master

PR #147, opened 2017-08-31
Log list of supplier IDs that we intend to send emails to

idavidmcdonald:
If our script fails middway, we rerun it and include a specific
list of supplier IDs only to send to.  If the script fails middway
though we don't know which IDs these are as they are never logged
out. This commit will log them meaning they can be easily be
copied and pasted back into the job (from the supplier ID of
failure).

### DIFF
--- a/dmscripts/notify_suppliers_of_new_questions_answers.py
+++ b/dmscripts/notify_suppliers_of_new_questions_answers.py
@@ -128,7 +128,10 @@ def main(data_api_url, data_api_token, email_api_key, stage, dry_run, supplier_i
             if supplier_id in supplier_ids
         )
     logger.info(
-        "{} suppliers found interested in these briefs".format(len(interested_suppliers))
+        "{} suppliers found interested in these briefs with the following IDs: {}".format(
+            len(interested_suppliers),
+            ",".join(map(str, interested_suppliers.keys()))
+        )
     )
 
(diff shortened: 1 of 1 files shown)


dmscripts/notify_suppliers_of_new_questions_answers.py
```diff
@@ -130,6 +130,9 @@ def main(data_api_url, data_api_token, email_api_key, stage, dry_run, supplier_i
     logger.info(
         "{} suppliers found interested in these briefs".format(len(interested_suppliers))
     )
+    logger.info(
+        "Supplier IDs found: {}".format(",".join(map(str, interested_suppliers.keys())))
+    )
```
katstevens: Can we combine the two log statements above?

katstevens: Also let's check it's fine in Python3 (I think it will be fine...)

CrystalPea: Doesn't Travis check for Python3 compatibility?

idavidmcdonald: Travis will check the unit tests in python 3 but won't actually do a functional run of the script so that is what is missing. I have tested a dry run on production using python 3 which should be sufficient for the moment I think.

PR #152, opened 2017-09-19
751 Email buyers about their recently closed briefs

katstevens:
Ticket: https://trello.com/c/poIdAQc0/751-3-create-script-to-send-emails-to-buyers-with-recently-closed-briefs-at-scheduled-intervals

We want to send emails via Notify to buyers of briefs that closed 4 weeks ago, telling them to award their brief. If they have not responded after 8 weeks (i.e. the brief is still closed), send another email.

- Top level script with
    - required args: the Data API, Notify API, Notify template ID 
    - optional args: list of buyer IDs (for resending failures), date_closed (to send for briefs closed on a specific date), offset days (to allow either 4 weeks or 8 weeks), dry run (to log without calling Notify)
- Script to query closed briefs (filtering by user ID list if supplied) and build the context dict to send to the Notify API
- 2 Notify templates have been created (to allow the 8-week email to be sent to the same user, which Notify would not allow if we used the same template)
- Some refactoring of existing functions for DRYness

TODO: 
- ~~some commits need squashing.~~
- ~~we need to add a 1-liner to the README/docs~~
- Jenkins jobs https://github.com/alphagov/digitalmarketplace-jenkins/pull/34
- ~~Change landing page 404 to error message if brief has already been awarded (Briefs FE)~~
- ~~GA events for the award flow forms (in the Briefs FE)~~ https://github.com/alphagov/digitalmarketplace-briefs-frontend/pull/36


dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,139 @@
+# -*- coding: utf-8 -*-
+from datetime import datetime, date, timedelta
+
+import dmapiclient
+from dmutils.email.dm_notify import DMNotifyClient
+from dmutils.email.exceptions import EmailError
```
Wynndow: `DMNotifyClient` and `EmailError` are both imported by the `__init__.py` of `dmutils.email` so you can just do `from dmutils.email import DMNotifyClient, EmailError`

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,139 @@
...
+from dmutils.formats import DATE_FORMAT
+
+from dmscripts.helpers import logging_helpers, date_helpers, brief_data_helpers
+from dmscripts.helpers.logging_helpers import logging
```
Wynndow: do you need to import this if you've already imported `logging_helpers`? I think you only use it on the next line so could just do `logger = logging_helpers.configure_logger({'dmapiclient': logging_helpers.logging.INFO})`

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,139 @@
...
+                        user['emailAddress'], notify_template_id, context_data, allow_resend=False
+                    )
+                except EmailError:
+                    failed_users.append(user['id'])
+
+    return failed_users
+
+
+def _check_that_brief_closing_date_before_the_offset(date_closed, offset_days):
+    if date_closed:
```
Wynndow: I found this a bit tough to get my head around. I think because `date_closed` is used a lot in the function as well as `main`. Might be worth adding a comment?

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,139 @@
...
+    if not dry_run:
+        notify_client = DMNotifyClient(notify_api_key, logger=logger)
+
+    closed_briefs = brief_data_helpers.get_closed_briefs(data_api_client, date_closed)
+    if not closed_briefs:
+        logger.info("No briefs closed on {date_closed}", extra={"date_closed": date_closed})
+        return True
+
+    # If user list supplied, check that the briefs have at least one user in the list
+    closed_briefs = _filter_briefs_by_optional_user_id_list(closed_briefs, user_id_list)
```
Wynndow: This reassignment of `closed_briefs` could be confusing. Could call it something else like `closed_briefs_for_supplied_users`. Or something.

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,139 @@
...
+def _filter_brief_users_by_optional_user_id_list(brief, user_id_list):
+    """Filter the brief's users by restricted user ID list if present
+    """
+    if user_id_list:
+        return filter(lambda user: user['id'] in user_id_list, brief['users'])
+    else:
+        return brief['users']
+
+
+def _filter_briefs_by_optional_user_id_list(closed_briefs, user_id_list):
```
Wynndow: I'm not sure this function name reflects what it does. It filters them by if they have users, and optionally by the user id list?

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,139 @@
...
+    """Return briefs if they have any users
```
idavidmcdonald: A brief should always have a user. There is a small chance that someone exists who is inactive and responsible for a brief but there should always exist someone.

With that being the case, we wouldn't always need to call this and it would be much clearer that we only need to call this function if a user_id_list is present i.e. a user_id_list is not optional.

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,139 @@
...
+
+from dmscripts.helpers import logging_helpers, date_helpers, brief_data_helpers
+from dmscripts.helpers.logging_helpers import logging
+
+logger = logging_helpers.configure_logger({'dmapiclient': logging.INFO})
+
+
+def _filter_brief_users_by_optional_user_id_list(brief, user_id_list):
+    """Filter the brief's users by restricted user ID list if present
+    """
```
idavidmcdonald: This docblock could be tidied up to be one line?

dmscripts/notify_buyers_to_award_closed_briefs.py
katstevens: Good to know! Perhaps I should change in (wherever the place I copied the import from was) too...

dmscripts/notify_buyers_to_award_closed_briefs.py
katstevens: Ugh, this was also a c&p from somewhere else - I thought Pea and I had refactored this out but I guess it got missed. Will fix

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,139 @@
...
+    data_api_client = dmapiclient.DataAPIClient(data_api_url, data_api_access_token)
+    if not dry_run:
+        notify_client = DMNotifyClient(notify_api_key, logger=logger)
+
+    closed_briefs = brief_data_helpers.get_closed_briefs(data_api_client, date_closed)
+    if not closed_briefs:
+        logger.info("No briefs closed on {date_closed}", extra={"date_closed": date_closed})
+        return True
+
+    # If user list supplied, check that the briefs have at least one user in the list
```
idavidmcdonald: The use of the filtering briefs stuff in here is hard to understand. I think that is because it is not too clear what is normal functionality and what is functionality only needed in case of a script failure and to be used as a developer manually rerunning something.

It might just be a case of changing the comment to something like (if this is correct):
If user list supplier, restrict/filter briefs to only include those owned by those users

dmscripts/notify_buyers_to_award_closed_briefs.py
katstevens: Aha - scripts is still using utils v.27.1.2. I don't think there's any harm bumping it up to v28?

Wynndow: Nope, wouldn't have thought so. The breaking change is about decoding invitation tokens.

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,139 @@
...
+from dmutils.email.exceptions import EmailError
+from dmutils.formats import DATE_FORMAT
+
+from dmscripts.helpers import logging_helpers, date_helpers, brief_data_helpers
+from dmscripts.helpers.logging_helpers import logging
+
+logger = logging_helpers.configure_logger({'dmapiclient': logging.INFO})
+
+
+def _filter_brief_users_by_optional_user_id_list(brief, user_id_list):
```
idavidmcdonald: Would feel nicer if we only called this if we wish to filter the briefs i.e. user_id_list would be a required argument.

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,139 @@
...
+                    notify_client.send_email(
+                        user['emailAddress'], notify_template_id, context_data, allow_resend=False
+                    )
+                except EmailError:
+                    failed_users.append(user['id'])
+
+    return failed_users
+
+
+def _check_that_brief_closing_date_before_the_offset(date_closed, offset_days):
```
idavidmcdonald: This function is doing more than the name says

Whilst it does check if the closing date is before the offset, it also currently responsible for working out what the closing date is based on the offset. 

This functionality should be split into separate areas, especially as when you read the main function, we set `date_closed = _check_that_brief_closing_date_before_the_offset` which is not great for readability.

dmscripts/helpers/brief_data_helpers.py
```diff
@@ -0,0 +1,9 @@
+from datetime import datetime
+from dmutils.formats import DATETIME_FORMAT
+
+
+def get_closed_briefs(data_api_client, date_closed):
```
idavidmcdonald: Nice on moving this into a helper

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,139 @@
...
+    if date_closed:
+        date_closed = date_helpers.get_date_closed(date_closed)
+        if date_closed > (datetime.utcnow() - timedelta(days=offset_days)).date():
+            return False
+        return date_closed
+    return date.today() - timedelta(days=offset_days)
+
+
+def main(
+    data_api_url, data_api_access_token, notify_api_key, notify_template_id, date_closed=None, dry_run=None,
```
idavidmcdonald: It might be nice to reorder the params such that:

`data_api_url, data_api_access_token, notify_api_key, notify_template_id, offset_days=28, dry_run=None, date_closed=None, user_id_list=None`

This way the two params that users would be changing when the script has not previously failed are at the front of the list and the final two are the failure cases. This mainly helps with readability as it tells you a little bit about the importance of the params.

dmscripts/notify_buyers_to_award_closed_briefs.py
katstevens: I'll rename and add a docstring - something like `Return a list of briefs if they have at least 1 user in the given ID list`?

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,139 @@
...
+            )
+        else:
+            logger.info(
+                "Notifying {no_of_users} user(s) about brief ID: {brief_id} - '{brief_title}'",
+                extra=logging_context
+            )
+            failed_users = notify_users(notify_client, notify_template_id, brief, user_id_list)
+            if failed_users:
+                failed_users_for_each_brief[brief['id']] = failed_users
+                all_failed_users.extend(failed_users)
```
idavidmcdonald: Wonder if we need both of these?
Are you expecting people to use information from both of the two error formats?

scripts/notify-buyers-to-award-closed-briefs.py
```diff
@@ -0,0 +1,52 @@
...
+if __name__ == '__main__':
+    arguments = docopt(__doc__)
+
+    list_of_buyer_ids = []
+    if arguments['--buyer-ids']:
+        list_of_buyer_ids = list(map(int, arguments['--buyer-ids'].split(',')))
+    if arguments['--offset-days']:
+        offset_days = int(arguments['--offset-days'])
+    else:
+        offset_days = 28
```
idavidmcdonald: If we are setting a default value here do we also need the default value on the main function?

dmscripts/notify_buyers_to_award_closed_briefs.py
katstevens: Yep, agreed. I'm also going to add more info about each param in the docstring

dmscripts/notify_buyers_to_award_closed_briefs.py
katstevens: Yeah, I wasn't sure about this - depends what's more likely to be useful. A full list of users who failed so we can c+p it straight into the script to re-run, or do we want to know the individual briefs as well (so we can figure out why those users failed)?

tests/helpers/test_date_helpers.py
```diff
@@ -0,0 +1,16 @@
...
+
+import dmscripts.helpers.date_helpers as date_helpers
+
+
+def test_get_date_closed():
+    def check_date_closed(value, expected):
+        with freeze_time('2015-01-02 03:04:05'):
+            assert date_helpers.get_date_closed(value) == expected
+
+    for value, expected in [
```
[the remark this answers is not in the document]
katstevens: Definitely - that means we can simplify this test as well. Have added a commit.

tests/test_notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,312 @@
...
+                {'brief_id': 100, 'brief_title': 'My brief title', 'utm_date': '20170101'}, allow_resend=False
+            ),
+            mock.call(
+                'c@example.com', 'NOTIFY_TEMPLATE_ID',
+                {'brief_id': 100, 'brief_title': 'My brief title', 'utm_date': '20170101'}, allow_resend=False
+            ),
+        ]
+
+
+@mock.patch('dmscripts.notify_buyers_to_award_closed_briefs.DMNotifyClient')
```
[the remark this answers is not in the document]
katstevens: Yep! Have done so for the logger as well (consistent with other scripts tests)

tests/test_notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,312 @@
...
+        get_closed_briefs.return_value = [self.brief1, self.brief2]
+        notify_users.return_value = []
+
+        with freeze_time('2016-01-29 03:04:05'):
+            assert notify_buyers_to_award_closed_briefs.main(
+                'URL', 'API_KEY', 'NOTIFY_KEY', 'NOTIFY_TEMPLATE_ID', date_closed='2016-01-01', dry_run=None
+            )
+            get_closed_briefs.assert_called_once_with(mock.ANY, datetime.date(2016, 1, 1))
+            notify_client.assert_called_once_with('NOTIFY_KEY', logger=mock.ANY)
+            notify_users.assert_has_calls([
```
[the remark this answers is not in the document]
katstevens: Argh how did that one sneak past me? 🙀

tests/test_notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,312 @@
...
+                    extra={
+                        'brief_title': 'Mochi making machine (for govt)',
+                        'brief_id': 200,
+                        'no_of_users': 1,
+                    }
+                )
+            ]
+
+    def test_main_notifies_about_briefs_closed_on_date_8_weeks_ago_using_offset_days(
+            self, logger, get_closed_briefs, notify_users, notify_client):
```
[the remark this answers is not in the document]
katstevens: Now fixed (PEP8 is happy, at least)

dmscripts/notify_buyers_to_award_closed_briefs.py
katstevens: Done

tests/helpers/test_brief_data_helpers.py
```diff
@@ -0,0 +1,29 @@
...
+import datetime
+
+from dmscripts.helpers import brief_data_helpers
+
+
+class TestGetClosedBriefs:
+
+    def test_get_closed_briefs_filters_by_date_closed(self):
+        api_client = mock.Mock()
+        api_client.find_briefs_iter.return_value = iter([
```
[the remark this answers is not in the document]
katstevens: I admit I blindly copied this from the old test module. Will sharpen it up a little.

tests/test_notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,312 @@
...
+
+
+class TestNotifyUsers:
+
+    brief = {
+        'id': 100,
+        'title': 'My brief title',
+        'lotSlug': 'lot-slug',
+        'frameworkSlug': 'framework-slug',
+        'users': [
```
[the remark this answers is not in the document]
katstevens: We're assuming multiple buyer users could exist in `notify_buyers_when_requirements_close` - this was intended to follow that pattern.

tests/test_notify_buyers_to_award_closed_briefs.py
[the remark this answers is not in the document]
katstevens: Now fixed

scripts/notify-buyers-to-award-closed-briefs.py
katstevens: Good point, now fixed

dmscripts/notify_buyers_to_award_closed_briefs.py
idavidmcdonald: Fair point.

Probably worth just putting this out there and then seeing how our script fairs and if we ever get any errors, checking we can then recover from them. I think the fact that notify won't let you send the same email twice to a user may be helpful (I believe)... meaning in most cases we could likely just run the whole script again if we get a random failure from notify (as opposed to a persistent one such as bad data).

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,149 @@
...
+                    notify_client.send_email(
+                        user['emailAddress'], notify_template_id, context_data, allow_resend=False
+                    )
+                except EmailError:
+                    failed_users.append(user['id'])
+
+    return failed_users
+
+
+def _get_brief_closing_date(closing_date_arg, offset_days):
```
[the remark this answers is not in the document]
katstevens: Agreed - will swap the args around.

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def main(
+    data_api_url, data_api_access_token, notify_api_key, notify_template_id, offset_days,
+    dry_run=None, date_closed=None, user_id_list=None
+):
+    """
+    Send emails to buyers via Notify, reminding them to award their closed briefs
+
+    offset_days:    send emails for briefs that closed X days ago (defaults to 28)
```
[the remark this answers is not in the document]
katstevens: Good point - will fix.

katstevens:
@TheDoubleK I've addressed the comments David left - do you want to take a look over this? I can walk you through it tomorrow if necessary. (Some of the commits probably need squashing but I will do that last in case there are any other changes.)

dmscripts/helpers/brief_data_helpers.py
benvand: Could probably do with a more descriptive name
`data_api_client.find_briefs_iter(status='closed', with_users=True)`
gets closed briefs. This method gets briefs closed on a specific day

dmscripts/helpers/date_helpers.py
```diff
@@ -0,0 +1,13 @@
+from datetime import datetime, timedelta
+from dmutils.formats import DATE_FORMAT
+
+
+def get_date_closed(date_closed):
```
benvand: This is quite a generic name, I'd consider putting something in the docstring about the default and why it's yesterday or maybe even something like:
```
def get_date_closed(date_closed, default=None):
   return datetime.strptime(date_closed, DATE_FORMAT).date() if date_closed else default
```
Then when you call it pass in the default.

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,149 @@
...
+
+def _filter_brief_users_by_optional_user_id_list(brief, user_id_list):
+    """Filter the brief's users by restricted user ID list if present"""
+    if user_id_list:
+        return filter(lambda user: user['id'] in user_id_list, brief['users'])
+    else:
+        return brief['users']
+
+
+def _filter_briefs_by_user_id_list(briefs, user_id_list):
```
benvand: This fits on one line < 120

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,149 @@
...
+from dmutils.email import DMNotifyClient
+from dmutils.email.exceptions import EmailError
+from dmutils.formats import DATE_FORMAT
+
+from dmscripts.helpers import logging_helpers, date_helpers, brief_data_helpers
+
+logger = logging_helpers.configure_logger({'dmapiclient': logging_helpers.logging.INFO})
+
+
+def _filter_brief_users_by_optional_user_id_list(brief, user_id_list):
```
benvand: This could also go on one line with a python ternary like so:
```return filter(lambda user: user['id'] in user_id_list, brief['users']) if user_id_list else brief['users']```

File: dmscripts/notify_buyers_to_award_closed_briefs.py (hunk 24 above)
Comment (benvand) [reply dmscripts/notify_buyers_to_award_closed_briefs.py]: Actually this could be:

```
return briefs if any(_filter_brief_users_by_optional_user_id_list(brief, user_id_list)) else []

```

File: dmscripts/notify_buyers_to_award_closed_briefs.py [hunk 26]
```diff
@@ -0,0 +1,149 @@
...
+
+        for user in brief_users:
+            if user['active']:
+                try:
+                    notify_client.send_email(
+                        user['emailAddress'], notify_template_id, context_data, allow_resend=False
+                    )
+                except EmailError:
+                    failed_users.append(user['id'])
+
```
benvand: Could use a set here because I don't think we care how many times the user appears but just an observation. Fine either way

dmscripts/helpers/date_helpers.py
katstevens: I'll add an explanation in the docstring - it feels a bit weird adding a `default=None` param when `date_closed` itself might be `None` anyway.

dmscripts/notify_buyers_to_award_closed_briefs.py
katstevens: Sweeeet, so it does! Will change

dmscripts/notify_buyers_to_award_closed_briefs.py
katstevens: I think some readability might be lost here? As it is, it's pretty clear that we apply the filter if the user id list is given, otherwise we return the unchanged list.

dmscripts/helpers/date_helpers.py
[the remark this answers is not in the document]
TheDoubleK: Ah OK, but I just noticed it's moving functionality from the existing "notify when briefs close" script.  Still, if it's a general date helper I think the name isn't quite right.

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,149 @@
...
+            _filter_brief_users_by_optional_user_id_list(brief, user_id_list)
+        )
+    ]
+
+
+def _create_context_for_brief(brief):
+    return {
+        'brief_id': brief['id'],
+        'brief_title': brief['title'],
+        'utm_date': date.today().strftime("%Y%m%d")
```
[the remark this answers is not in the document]
katstevens: 🙀 Templates now fixed!

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+
+def notify_users(notify_client, notify_template_id, brief, user_id_list):
+    failed_users = []
+    if brief['users']:
```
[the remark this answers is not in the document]
katstevens: Good point. If there are somehow no users for a brief, the script should fail loudly anyway.

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,149 @@
...
+    user_id_list:   if supplied, only send emails to buyer with these user IDs
+    """
+    logger.info("Data API URL: {data_api_url}", extra={'data_api_url': data_api_url})
+
+    date_closed = _get_brief_closing_date(offset_days, date_closed)
+    if not date_closed:
+        logger.error('Not allowed to notify about briefs that closed less than {} days ago', offset_days)
+        return False
+
+    data_api_client = dmapiclient.DataAPIClient(data_api_url, data_api_access_token)
```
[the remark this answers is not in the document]
katstevens: It's definitely inconsistent 😿. Scripts where we create the client inside the `dmscripts` bit:
- Notify buyers when requirements close
- Notify suppliers of new questions and answers

I'd say there's an argument for keeping as much code inside the `dmscripts` function as possible - the top level script would ideally have no logic and just pass through args and any additional environment config to a nice testable function?

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,149 @@
...
+    if not dry_run:
+        notify_client = DMNotifyClient(notify_api_key, logger=logger)
```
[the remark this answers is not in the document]
katstevens: Good call. We originally had it as you suggested and I can't remember why we moved it, possibly overenthusiastic refactoring!

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,149 @@
...
+    data_api_url, data_api_access_token, notify_api_key, notify_template_id, offset_days,
+    dry_run=None, date_closed=None, user_id_list=None
+):
+    """
+    Send emails to buyers via Notify, reminding them to award their closed briefs
+
+    offset_days:    send emails for briefs that closed X days ago
+    dry_run:        log instead of sending emails
+    date_closed:    if supplied, send emails for briefs that closed on this date
+    user_id_list:   if supplied, only send emails to buyer with these user IDs
```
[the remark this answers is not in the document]
katstevens: Done

dmscripts/notify_buyers_to_award_closed_briefs.py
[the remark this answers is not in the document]
TheDoubleK: What you're saying is that your team completely ignored the established convention during the last mission when you added those new Notify scripts :p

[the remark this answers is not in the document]
TheDoubleK: There was no inconsistency before then.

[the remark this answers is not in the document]
TheDoubleK: And you're going to mock out the API client in tests anyway.

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,144 @@
...
+def notify_users(notify_client, notify_template_id, brief, user_id_list):
+    failed_users = []
+
+    context_data = _create_context_for_brief(brief)
+    brief_users = _filter_brief_users_by_optional_user_id_list(brief, user_id_list)
+
+    for user in brief_users:
+        if user['active']:
+            try:
+                notify_client.send_email(
```
TheDoubleK: Really really sorry, I should have mentioned/spotted this before, but...

I think *this* is the call that should be wrapped with an `if dry_run...`, and not the higher-level call to `notify_users`.  As things stand the `dry-run` will not test the creation of context data or filtering users.  You want the dry run to run as much of the script as possible, and only skip the API calls that update data/send an email or whatever to give the fullest possible test of what would happen, just without doing the actual thing.

Sorry.

katstevens: No that makes perfect sense! We were trying to keep all the logging at the same depth, so the flag got bumped up into `main`. I will have a look and see if I can get it closer to the `notify_client` call.

katstevens:
Just realised that we're not sending the `lot-slug` to Notify so the URL is going to be wrong for some briefs 🙀.

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,147 @@
...
+        'brief_title': brief['title'],
+        'brief_id': brief['id'],
+        'no_of_users': len(brief['users']),
+    }
+    email_context_data = _create_context_for_brief(brief)
+    if user['active']:
+        try:
+            if dry_run:
+                logger.info(
+                    "Would notify {no_of_users} user(s) about brief ID: {brief_id} - '{brief_title}'",
```
TheDoubleK: The log messages here don't seem quite right now - these are going to be output per-user rather than per-brief, so probably should be something more like `"Would notify {email_address} about brief ID: {brief_id} - '{brief_title}'`

For the dry run we can output actual email addresses as it helps with debugging, and perhaps we don't need to log each email for the non-dry-run case?  (That's how we do it here: https://github.com/alphagov/digitalmarketplace-scripts/blob/master/dmscripts/upload_counterpart_agreements.py#L83)

Or maybe for the non-dry-run we log the headline number somewhere before the call to `send_email_to_brief_user_via_notify`?

TheDoubleK: (Either way we probably don't want actual email addresses output in live logs)

katstevens: Good spot. How about logging user IDs instead of emails though (like we do with errors)? Still easy enough to debug but no personal info logged on live?

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,147 @@
...
+
+    # If user list supplied, only continue for briefs that have at least one user in that list
+    if user_id_list:
+        closed_briefs = _filter_briefs_by_user_id_list(closed_briefs, user_id_list)
+
+    logger.info("Notifying users about {briefs_count} closed brief(s)", extra={'briefs_count': len(closed_briefs)})
+
+    failed_users_by_brief_id = {}
+    for brief in closed_briefs:
+        failed_users_for_this_brief = []
```
TheDoubleK: Yep having the headline "Notifying X users about brief Y" log message would be fine here I think (the same with or without dry-run), and then just log individual emails for the dry-run inside the `send_email...` call.

dmscripts/notify_buyers_to_award_closed_briefs.py
katstevens: The dry run now looks something like this:

```
2017-10-09 13:55:27,636 script INFO Notifying users about 1 closed brief(s)
2017-10-09 13:55:27,636 script INFO Would notify user ID 34321 about brief ID 5148: 'Digital Discovery: Exploring the Potential for Business Energy Efficiency Website'
```

And the wet run would be (including the Notify client's logging):

```
2017-10-09 13:55:27,636 script INFO Notifying users about 1 closed brief(s)
2017-10-09 13:55:27,636 script INFO Notifying user ID 34321 about brief ID 5148: 'Digital Discovery: Exploring the Potential for Business Energy Efficiency Website'
2017-10-09 13:55:27,636 script INFO Sent email M7o-Z5TLAM6McxNgHy0tk-aAnkzXB6vKeLh0AzLHuwA= to xID-VUdAEenUtL2p_b8rLRiUO4eRYxV1WK7TNDI1zUc= (id: 75216ef4-e42a-4751-ac46-a49887f71cca, template: bc1f93a8-9529-48b6-9a82-af61bc6da332) through Notify
```

dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -0,0 +1,147 @@
...
+        for user in brief['users']:
+            failed_user_id = send_email_to_brief_user_via_notify(
```
TheDoubleK: It feels a bit risky relying on our Notify client to return `None` for a successful send - that feels like something that could maybe change in future and would catch us out here.

katstevens: I'm confused - we catch errors from the Notify client, and if there's no exception then it's `send_email_to_brief_user_via_notify` that returns None?

dmscripts/notify_buyers_to_award_closed_briefs.py
katstevens: I don't think we really need to know how many users are being emailed for each brief (yet!) so I've just taken it out. We can review this when we start having multiple brief users?

dmscripts/notify_buyers_to_award_closed_briefs.py
TheDoubleK: OK it's me who's totally confused here, sorry. For some reason when I wrote that comment I had a moment of madness and thought that Python does the Ruby thing of returning the last thing from a method by default, which in this case would have been the return value of `notify_client.send_email()`. Clearly I wasn't well at he time.

dmscripts/notify_buyers_to_award_closed_briefs.py
TheDoubleK: Yep this seems good to me.

dmscripts/notify_buyers_to_award_closed_briefs.py
TheDoubleK: Yep.

TheDoubleK:
All looks good now, let's get this in... it's been a marathon!  
:+1:

katstevens:
Woohoo! Thanks again Kev for your part in this marathon slog of a PR...

PR #157, opened 2017-10-11
105-create-complete-opportunity-data-script-to-publish-on-the-digital-marketplace-automated

benvand:
https://trello.com/c/5rEnbBR5/105-create-complete-opportunity-data-script-to-publish-on-the-digital-marketplace-automated

Create a new csv.
Tidy up some of our csv creation.

### DIFF
--- a/dmscripts/models/process_rules.py
+++ b/dmscripts/models/process_rules.py
@@ -8,3 +8,9 @@ def format_datetime_string_as_date(dt):
 
 def remove_username_from_email_address(ea):
     return '{}'.format(ea.split('@').pop()) if ea else None
+
+def construct_brief_url(brief_id):
+    return (
+        'https://www.digitalmarketplace.service.gov.uk/'
+        'digital-outcomes-and-specialists/opportunities/{}'.format(brief_id)
+    )
--- a/dmscripts/models/process_rules.py
+++ b/dmscripts/models/process_rules.py
@@ -9,6 +9,7 @@ def format_datetime_string_as_date(dt):
 def remove_username_from_email_address(ea):
     return '{}'.format(ea.split('@').pop()) if ea else None
 
+
 def construct_brief_url(brief_id):
     return (
         'https://www.digitalmarketplace.service.gov.uk/'
--- a/dmscripts/models/queries.py
+++ b/dmscripts/models/queries.py
@@ -26,6 +26,12 @@ def base_model(base_model, keys, get_data_kwargs, client, logger=None, limit=Non
 
 
 def model(model, directory):
+    """Return a Pandas DataFrame loaded from a csv of a given DM model.
+
+    :param model: The model we are working with, used as the name of the .csv
+    :param directory: The directory in which to find the model data csv.
+    :return: Pandas DataFrame of model data loaded from csv.
+    """
     return pandas.read_csv(csv_path(directory, model))
 
 
--- a/dmscripts/models/queries.py
+++ b/dmscripts/models/queries.py
@@ -35,13 +35,26 @@ def model(model, directory):
     return pandas.read_csv(csv_path(directory, model))
 
 
-def join(join, directory):
-    models = [model['model'] for model in join]
-    data, joined = (pandas.read_csv(csv_path(directory, model)) for model in models)
-    left_on, right_on = (model['key'] for model in join)
+def join(data, model, left_on, right_on, directory, data_duplicate_suffix=None):
+    """Left join the model data csv denoted by 'model' to 'data'.
 
-    return data.merge(joined, left_on=left_on, right_on=right_on,
-                      suffixes=['_{}'.format(model) for model in models])
+    :param data: The current pandas DataFrame we are working with.
+    :param model: The name of a csv in the data directory. This will be joined to data.
+    :param left_on: The field in 'data' we are joining on.
+    :param right_on: The field in the model csv we are joining on.
+    :param directory: The data directory.
+    :param data_duplicate_suffix: An optional suffix for fields duplicated in both datasets.
+    :return: pandas DataFrame of model joined to data.
+    """
+    csv_to_be_joined = pandas.read_csv(csv_path(directory, model))
+
+    return data.merge(
+        csv_to_be_joined,
+        how='left',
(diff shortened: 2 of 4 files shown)


dmscripts/models/queries.py
```diff
@@ -5,6 +5,16 @@
...
+
+    :param base_model: A Digital Marketplace model (client must have a 'find_{model}_iter' method)
+    :param keys: The attributes we require of that model.
+    :param get_data_kwargs: Additional kwargs for the get request the client will make.
+    :param client: Instantiated Digital Marketplace APIClient
+    :param logger:
+    :param limit: Maximum number of requests for the client to perform.
+    :return: A pandas DataFrame of the requested data. Columns as model attributes, rows as instances.
+    """
     mt = ModelTrawler(base_model, client)
```
MuriloDalRi: I love that we call this Model Trawler! :)

scripts/get-model-data.py
```diff
@@ -286,14 +401,16 @@
...
+        if 'process_fields' in config:
+            data = queries.process_fields(config['process_fields'], data)
         if 'rename_fields' in config:
             data = queries.rename_fields(config['rename_fields'], data)
 
         # sort list by some dict value
         if 'sort_by' in config:
             data = queries.sort_by(config['sort_by'], data)
 
+        data = data.drop_duplicates()
```
katstevens: Do we test this anywhere? I'd like to understand what sort of duplicates we might get as a result of the joins (we shouldn't get duplicates for straightforward model dumps, right?).

scripts/get-model-data.py
```diff
@@ -222,6 +236,105 @@
...
+            'requirementsLength': 'Open For',
+            'applicationsFromSMEs': 'Applications from SMEs',
+            'applicationsFromLargeOrganisations': 'Applications from Large Organisations',
+            'totalApplications': 'Total Organisations',
+            'status': 'Status',
+            'awarded_supplierName': 'Winning supplier',
+            'awarded_supplierOrganisationSize': 'Size of supplier',
+            'awarded_awardedContractValue': 'Contract amount',
+            'awarded_awardedContractStartDate': 'Contract start date'
+        }
```
katstevens: One day (not today) I would love to see all this config held in a separate place to the code that churns out the CSVs themselves. We could have a lovely `ModelExportCSV` class that initialises with the config, queries the data, does some manipulation/renaming etc, has testable methods, avoids all that logic in the top level... _wanders off into dream land_

```
class ModelExportCSV(name, model, keys, joins=None, process_fields=None, ...):
    def get_query_data(self, ...):
          ...   
    def rename_fields(self, ...):
          ...  
    def output_to_dir(self, ...):
          ...  
    def run(self):
          # Do everything
          ...

opportunity_data_export = ModelExportCSV(**CONFIGS['opportunity_data'])
opportunity_data_export.run()
```

Something for the next 🔥 💔 maybe!

scripts/get-model-data.py
benvand: Hey, no we shouldn't get duplicates from the standard model dumps (which is why this is safe here). And no this isn't tested anywhere. To test it I ran the csv generation locally using master and this branch and made sure the respective  csvs had the same number of rows. 

This new csv does have duplicates. It lists briefs but is actually created from the brief_responses csv.
We have to load up all the brief responses to get the counts for how many applied so until we do the drop duplicates there is a row for every brief response. 

At this point in the code the brief response rows are exact duplicates because we have filtered out columns that are brief response specific using the `keys` section of the config.

Doing this at the end is like doing a sql `SELECT DISTINCT ON`

Bit fiddly! 😅

dmscripts/models/queries.py
```diff
@@ -66,6 +98,16 @@ def add_counts(join, group_by, model, data, directory):
     return data
 
 
+def add_aggregation_counts(data, group_by, join, count_name, query=None):
+    left_on, right_on = join
+
+    count_data = (data.query(query) if query else data).groupby(group_by)[group_by].count()
```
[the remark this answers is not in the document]
benvand: I know :nauseated_face: 🐼 s is weird

scripts/get-model-data.py
benvand: Yeah, I'm gonna make a tech debt ticket for this

dmscripts/models/queries.py
benvand: http://www.ebay.co.uk/bhp/fishing-trawler 😂

scripts/get-model-data.py
katstevens: Cool, that makes sense. However in the long term I think I'd feel better if `drop_duplicates` was only done for this new dump or ones like it - maybe add a comment or a note to the tech debt ticket to make it clear why we might get duplicates in the first place? (Your explanation above would do just fine!)

dmscripts/models/queries.py
```diff
@@ -35,13 +35,26 @@ def model(model, directory):
...
+
+    :param data: The current pandas DataFrame we are working with.
+    :param model: The name of a csv in the data directory. This will be joined to data.
+    :param left_on: The field in 'data' we are joining on.
+    :param right_on: The field in the model csv we are joining on.
+    :param directory: The data directory.
+    :param data_duplicate_suffix: An optional suffix for fields duplicated in both datasets.
+    :return: pandas DataFrame of model joined to data.
+    """
+    csv_to_be_joined = pandas.read_csv(csv_path(directory, model))
```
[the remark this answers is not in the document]
benvand: Done 👍

dmscripts/models/process_rules.py
```diff
@@ -8,3 +8,9 @@ def format_datetime_string_as_date(dt):
 
 def remove_username_from_email_address(ea):
     return '{}'.format(ea.split('@').pop()) if ea else None
+
+def construct_brief_url(brief_id):
+    return (
+        'https://www.digitalmarketplace.service.gov.uk/'
+        'digital-outcomes-and-specialists/opportunities/{}'.format(brief_id)
```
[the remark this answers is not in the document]
benvand: I've used a global now, it was that or hardcode 😬

PR #160, opened 2017-10-12
Script to upload file to s3 given stage and bucket

benvand:


scripts/upload-file-to-s3.py
```diff
@@ -0,0 +1,56 @@
...
+Usage:
+    scripts/upload-file-to-s3.py
+        <stage> <file_path> <bucket_category> [<remote_path>] [<remote_name>] [--dry-run] [--public]
+
+Options:
+    -h --help   Show this screen.
+    --stage=<stage>  The stage being updated
+    --file_path=<file_path>  Path of file to be saved in S3
+    --bucket_category=<bucket_category>  Bucket category, used to get name ie agreements/ communications/ documents
+    --remote_path=<remote_path>  Path to save the file at inside the bucket
```
Wynndow: S3 seems to use the idea of objects having a key rather than there being a file hierarchy (http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html). You could rename it here to reflect that. Although the concept of a file path is probably less confusing...

scripts/upload-file-to-s3.py
```diff
@@ -0,0 +1,56 @@
...
+    --download_name=<download_name>  Suggested name for a browser to download, part of Content-Disposition header
```
Wynndow: Can you give an example of this being used? Not sure I understand it.

scripts/upload-file-to-s3.py
```diff
@@ -0,0 +1,56 @@
...
+    download_name = arguments.get('<remote_name>', '')
+    dry_run = arguments['--dry-run']
+    public = arguments['--public']
+
+    # Set defaults
+    acl = 'public-read' if public else 'private'
+    download_name = download_name if download_name else os.path.basename(file_path)
+
+    # Grab bucket
+    bucket = S3(get_bucket_name(stage, bucket_category))
```
Wynndow: Instead of giving a bucket category, is there value in just supplying a bucket name? Would mean the script is useable if we want to upload to a bucket outside of that pattern at any point.

scripts/upload-file-to-s3.py
benvand: Hey so this sets the content-disposition header attribute of the file which determines whether it will be opened in browser or downloaded as a file.
https://github.com/alphagov/digitalmarketplace-utils/blob/a33fad66d50bc0e9f06193cac28c015a4a292121/dmutils/s3.py#L53-L53

Wynndow: Ah ok, so it's just a header to tell a browser what to do with the thing it's getting?

PR #163, opened 2017-10-16
Repair broken status column in successful_brief_responses.csv

benvand:


scripts/get-model-data.py
```diff
@@ -207,12 +207,13 @@
             'supplierId',
             'supplierName',
             'submittedAt',
-            'status',
+            'status_briefs',
```
Wynndow: Does this need to be `status_brief`, as per the `data_duplicate_suffix` above?

benvand: This is briefs as per the model name in the join. Actually, I'm going to make that clearer.

benvand: Done

PR #167, opened 2017-10-17
Update indexing script to be more user friendly

Wynndow:
1. Allow auth token helpers to match dev aliases

2. Update docstring to better reflect script use

### DIFF
--- a/dmscripts/helpers/auth_helpers.py
+++ b/dmscripts/helpers/auth_helpers.py
@@ -2,9 +2,11 @@ import os
 import subprocess
 import yaml
 
+DEV_ALIASES = ('dev', 'development', 'local')
+
 
 def get_auth_token(api, stage):
-    if stage == 'development':
+    if stage.lower() in DEV_ALIASES:
         return 'myToken'
 
     DM_CREDENTIALS_REPO = os.environ.get('DM_CREDENTIALS_REPO', '../digitalmarketplace-credentials')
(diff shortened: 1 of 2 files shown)


dmscripts/helpers/auth_helpers.py
```diff
@@ -2,9 +2,10 @@
 import subprocess
 import yaml
 
+DEV_ALIASES = ('dev', 'development', 'local')
 
 def get_auth_token(api, stage):
-    if stage == 'development':
+    if stage in DEV_ALIASES:
```
samuelhwilliams: Might be good to throw in `stage.lower()`

scripts/index-services.py
```diff
@@ -2,18 +2,21 @@
 """Read services from the API endpoint and write to search-api for indexing.
 
 Usage:
-    index-services.py <stage> --frameworks=<frameworks> [options]
+    index-services.py <stage> --index=<index> --frameworks=<frameworks> [options]
 
-    --serial                                      Do not run in parallel (useful for debugging)
-    --index=<index>                               Search API index name [default: g-cloud-9]
+    <stage>                                       One of dev, preview, staging or production
+    --index=<index>                               Search API index name
```
samuelhwilliams: Maybe say that it should be of the format `<framework>-YYYY-MM-DD`.

Wynndow: Does it actually need to be though? For local, would it matter?

samuelhwilliams: It doesn't matter specifically.

Wynndow: Cool. Have updated.

PR #168, opened 2017-10-17
get-model-data.py::opportunity-data extra fields

benvand:
https://trello-attachments.s3.amazonaws.com/59ca72f6013a566bf761ba93/59ce13f29c7d49a341a910ba/c812072a78c51d2c3d702b3efa9c3e9b/opportunity-data.csv


scripts/get-model-data.py
```diff
@@ -287,10 +287,11 @@
             'location',
             'publishedAt',
             'requirementsLength',
-            'totalApplications',
+            'contractLength',
             'applicationsFromSMEs',
             'applicationsFromLargeOrganisations',
-            'status_briefs',
+            'totalApplications',
+            'status_data',
```
katstevens: Is this key now coming from the status plus the `data_duplicate_suffix`?

benvand: Yes, `_briefs` is no longer used.

PR #169, opened 2017-10-18
Add direct award project search data to get-model-scripts

Wynndow:
For this ticket: https://trello.com/c/eo8gJE4C

Tests won't pass until this PR on the apiclient is in: https://github.com/alphagov/digitalmarketplace-apiclient/pull/105

### Allow model trawler to find `get` iter methods  …
The model trawler finds all iter methods on the api client with a
'find_' prefix.

Direct award project services has an iter with a 'get_' prefix.

### Add new process rules for processing model data  …
`extract_id_from_user_info` is useful for getting a string of user ids
from a list of user data objects.

`convert_none_to_zero` is useful if you have missing data which you'd
rather be presented as an integer.

### Add `how` parameter to `join` query  …
The join query uses the pandas merge function. Merge can use 'left',
'right', 'inner' or 'outer'.

This defaults to left but allows you to chose depending on your needs

### Add `group_by` to queries  …
Sometimes you want to group and count your data in place, without
joining to anything else.

### Add `get_model_by_fk` query  …
Sometimes you want to get data from the api that requires the id of a
different model to be supplied in the request. If you want to get that
data for a bunch of different model id's.

An example of this is direct_award_project_searches. The url on the api
to get them is `/direct-award/projects/<int:project_id>/searches`.

If we want to retrieve all the searches from the api we first need a
list of all the projects, and the iterate over their id's.

It's also useful to be able to filter the original models before
fetching the data via their id's.

It's also useful to be able to group and count them before returning
the data, in case it doesn't have a key to group by later.

### Add configs for direct award search data  …
The main configs of interest for tracking the use of saved searches are
`direct_award_projects_with_locked_service_count` and
`direct_award_saved_and_locked_searches_count_by_user`.

The others being committed here are useful as stepping stones to compile
this data, but may well also be useful in their own right.

### Update tests to removed pytest warnings  …
Warnings were:

`[pytest] section in setup.cfg files is deprecated, use [tool:pytest]
instead.`

`yield tests are deprecated, and scheduled to be removed in pytest 4.0`


tests/test_notify_buyers_when_requirements_close.py
```diff
@@ -37,7 +37,7 @@ def check_date_closed(value, expected):
         (None, datetime.date(2015, 1, 1)),
         ('2016-01-02', datetime.date(2016, 1, 2))
     ]:
-        yield check_date_closed, value, expected
+        check_date_closed(value, expected)
```
galund: I guess this could have been replaced with a parameterized test, but I'm honestly not sure there's much benefit. https://docs.pytest.org/en/latest/parametrize.html

scripts/get-model-data.py
```diff
@@ -377,21 +484,37 @@
...
         # transform values that we want to transform
         if 'assign_json_subfields' in config:
             for field, subfields in config['assign_json_subfields'].items():
                 data = queries.assign_json_subfields(field, subfields, data)
+
         if 'duplicate_fields' in config:
             for field, new_name in config['duplicate_fields']:
                 data = queries.duplicate_fields(data, field, new_name)
 
+        if 'process_fields' in config:
```
benvand: 🤔 Iiiinnteresting, so this needs to come above group by because of the `convert_none_to_zero` fair enough

benvand:
Ah there was one thing re:
> Direct award project services has an iter with a 'get_' prefix.

Can we get a tech debt ticket to change that if it goes against all our other patterns? Is that what the problem is?

Looks like we have 
`find_brief_responses` given a brief id
`find_services` given a supplier id etc

Should be
`find_direct_award_project_services` given project id/ user id

scripts/get-model-data.py
Wynndow: Yep. The whole thing is a bit weird. It's this sort of declarative thing where you specify a config for the data you want, but you also need to keep in mind the order things are getting applied in. 

It'd be nice if there was some sort of list of actions, which were different queries to apply, and they got applied in order.

Wynndow:
Ha, yeah. Rebasing was....interesting. 

Yep, I'll add a tech debt ticket to look into it. For now I've fudged around it with the last commit here on the apiclient https://github.com/alphagov/digitalmarketplace-apiclient/pull/105/commits

I was chatting with Sam about the get/find thing. The get was used as we have the PK of the thing we're getting data for, and finds are just getting all the data. But it looks like there are some `find` methods that also take the PK of a thing. It needs looking at more deeply I think.

Wynndow:
@benvand Have rebased a run locally and all looks well. If you could run and compare against your stuff that'd be swell.

benvand:
```
(venv) benvandersteen@gds5077:~/gds/branches/digitalmarketplace/digitalmarketplace-scripts$ git diff data/opportunity-data.csv data/opportunity-data-c.csv 
```

PR #171, opened 2017-10-26
114 send emails to supplier when its been withdrawn

benvand:


dmscripts/notify_suppliers_of_brief_withdrawal.py
```diff
@@ -0,0 +1,45 @@
...
+
+    withdrawn_briefs = data_api_client.find_briefs(withdrawn_on=withdrawn_date).get("briefs")
+
+    if brief_id:
+        withdrawn_briefs = filter(lambda i: i['id'] == brief_id, withdrawn_briefs)
+
+    for brief in withdrawn_briefs:
+        email_addresses = get_brief_response_emails(data_api_client, brief['id'])
+        if not email_addresses:
+            continue
```
katstevens: Do we gain much from breaking the loop here? Should we log if a withdrawn brief has no applications?

dmscripts/notify_suppliers_of_brief_withdrawal.py
```diff
@@ -0,0 +1,45 @@
...
+        for email_address in email_addresses:
+            if not dry_run:
+                mail_client.send_email(
+                    email_address, template_id, brief_email_context, allow_resend=False
+                )
+            logger.info("{}EMAIL: 'Withdrawal of Brief ID: {} to {}".format(
+                '[Dry-run]' if dry_run else '',
+                brief['id'],
+                hash_string(email_address),
+            ))
```
katstevens: A lot of the other scripts pass log vars as an `extra={}` param, though I'm not 100% clear why. Any particular reason why we're not doing that here though?

scripts/notify-suppliers-of-brief-withdrawal.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python
+"""
+If a brief has been withdrawn suppliers need to be notified that their applications have also been cancelled. This
+script notifies suppliers of all briefs withdrawn on a given date. Alternatively, supplier with a given brief
```
katstevens: `Alternatively, supplier with a given brief...` this kind of trails off? 😸

scripts/notify-suppliers-of-brief-withdrawal.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # Get arguments
+    stage = arguments['<stage>']
+    api_token = arguments['<api_token>']
+    govuk_notify_api_key = arguments['<govuk_notify_api_key>']
+    govuk_notify_template_id = arguments['<govuk_notify_template_id>']
+    withdrawn_date = arguments.get('<withdrawn_date>', None)
+    brief_id = arguments.get('<brief_id>', None)
+    dry_run = arguments['--dry-run']
+    verbose = arguments['--verbose']
```
katstevens: Ooooh `verbose`, nice 👍

tests/test_notify_suppliers_of_brief_withdrawal.py
```diff
@@ -0,0 +1,94 @@
...
+
+BRIEF_RESPONSES = {"briefResponses": [
+    {"id": 4321, "respondToEmailAddress": "email@me.now"},
+    {"id": 4389, "respondToEmailAddress": "email@them.now"}
+]}
+
+
+EXPECTED_BRIEF_CONTEXT = {
+    'brief_title': "Tea Drinker",
+    'brief_link': 'https://www.preview.marketplace.team/digital-outcomes-and-specialists/opportunities/123'  # noqa
```
katstevens: `# noqa`?

tests/test_notify_suppliers_of_brief_withdrawal.py
```diff
@@ -0,0 +1,94 @@
...
+    assert create_context_for_brief.call_args_list == [mock.call("preview", WITHDRAWN_BRIEFS["briefs"][0])]
+    assert notify_client.send_email.call_args_list == [
+        mock.call("email@me.now", "notify_template_id", EXPECTED_BRIEF_CONTEXT, allow_resend=False),
+        mock.call("email@them.now", "notify_template_id", EXPECTED_BRIEF_CONTEXT, allow_resend=False)
+    ]
+
+
+@mock.patch('dmutils.email.DMNotifyClient', autospec=True)
+@mock.patch('dmapiclient.DataAPIClient', autospec=True)
+def test_main_calls_correct_client_methods(data_api_client, notify_client):
```
katstevens: I'm not sure what the difference is between this test and the one above? Could they be combined with a more descriptive test name e.g. `test_main_notifies_all_withdrawn_briefs_by_default` (to differentiate it from the `brief_id` test below?

tests/test_notify_suppliers_of_brief_withdrawal.py
```diff
@@ -0,0 +1,94 @@
...
+def test_single_call_when_brief_id_specified(data_api_client, notify_client):
+    """Script should only look up brief responses for given brief id when brief id is specified."""
+    data_api_client.find_briefs.return_value = WITHDRAWN_BRIEFS
+    brief_id = 235
+    result = tested_script.main(
+        data_api_client, notify_client, 'notify_template_id', 'preview', mock.Mock(), date.today(), brief_id=brief_id
+    )
+
+    assert result is True
+    data_api_client.find_brief_responses.assert_called_once_with(brief_id=brief_id, status='submitted')
```
katstevens: Should we also assert that we only send one email?

dmscripts/notify_suppliers_of_brief_withdrawal.py
benvand: | Do we gain much from breaking the loop here? 
No but I like how it reads, it explicitly informs the user that we take no action if we have no one to email.

| Should we log if a withdrawn brief has no applications?
I guess my thinking was that it _might_ come in useful to someone at some point but no one has asked for it and logging for a path that intentionally does nothing feels a bit weird

tests/test_notify_suppliers_of_brief_withdrawal.py
benvand: So the idea is that the part of the script that I am testing here is the part that gets the brief responses and therefore the emails. 
At this point it could send any number of emails and still be valid but it should only make one call to the brief responses end point.

tests/test_notify_suppliers_of_brief_withdrawal.py
benvand: This test explicitly checks client methods and nothing else. The one above mocks _everything_ and checks the correct methods in the script are called.

The first is at the boundary of the methods - method within the scripts
The second is at the boundary of the script - clients

tests/test_notify_suppliers_of_brief_withdrawal.py
benvand: Fixed this

scripts/notify-suppliers-of-brief-withdrawal.py
benvand: 😎

scripts/notify-suppliers-of-brief-withdrawal.py
benvand: Fixed this

dmscripts/notify_suppliers_of_brief_withdrawal.py
benvand: Will fix

tests/test_notify_suppliers_of_brief_withdrawal.py
katstevens: OK that makes sense, but both tests assert the `data_api_client` and `notify_client` calls at the moment, and the first one mocks out `get_brief_response_emails` but doesn't assert its call. 

To help devs with poor memories (i.e. me) I'd rename the tests to make it clear what you're doing - maybe something like `test_main_calls_script_methods` and `test_main_calls_external_clients`?

tests/test_notify_suppliers_of_brief_withdrawal.py
katstevens: Ahh you're right of course - multiple suppliers could apply to one brief, that makes sense now. I think I got confused because we don't assert the call for `data_api_client.get_brief_response_emails` when it's called multiple times (because we're mocking out the helper function?).

dmscripts/notify_suppliers_of_brief_withdrawal.py
katstevens: OK, fair enough 😸

dmscripts/notify_suppliers_of_brief_withdrawal.py
```diff
@@ -0,0 +1,45 @@
...
+            env_helpers.get_web_url_from_stage(stage),
+            brief['frameworkFramework'],
+            brief['id']
+        )
+    }
+
+
+def main(data_api_client, mail_client, template_id, stage, logger, withdrawn_date=None, brief_id=None, dry_run=False):
+
+    withdrawn_briefs = data_api_client.find_briefs(withdrawn_on=withdrawn_date).get("briefs")
```
Wynndow: Also, that endpoint paginates, so if there are more than 100 withdrawn briefs in a day (unlikely I know, but if we scaled massively it might be possible) it wouldn't send emails for every brief?

dmscripts/notify_suppliers_of_brief_withdrawal.py
```diff
@@ -0,0 +1,45 @@
...
+        'brief_title': brief['title'],
+        'brief_link': '{0}/{1}/opportunities/{2}'.format(
+            env_helpers.get_web_url_from_stage(stage),
+            brief['frameworkFramework'],
+            brief['id']
+        )
+    }
+
+
+def main(data_api_client, mail_client, template_id, stage, logger, withdrawn_date=None, brief_id=None, dry_run=False):
```
Wynndow: Should `withdrawn_date` be a kwarg like this? If for some reason it's not given a value, you'll end up with ALL briefs, not just withdrawn.

The other part of the script that does the parsing of arguments etc sets `withdrawn_at` to yesterday if it's not already set, so this could be a positional argument so that it's harder to pass in None by accident.

I might be being too paranoid though....

benvand: Great shout (I'll probably have nightmares about that scenario tonight 😂 )

Wynndow: You could add a guard clause in at the top of the script instead of changing it to a positional arg. Something like?
```
if not withdrawn_on:
    Abort
```

benvand: Yeah I'll add it to the request to the api

dmscripts/notify_suppliers_of_brief_withdrawal.py
benvand: Ah, don't need to.
Our custom formatter only has:
`LOG_FORMAT = '%(asctime)s %(name)s %(levelname)s %(message)s'`
and the first arg is passed as message anyway.
Can use it with %s formatting and pass string formatting arguments as *args though.

tests/test_notify_suppliers_of_brief_withdrawal.py
benvand: Done

benvand:
@Wynndow @katstevens 
Comments addressed and commits squashed

dmscripts/notify_suppliers_of_brief_withdrawal.py
katstevens: So when we use `extra={}` elsewhere we're not gaining anything?

dmscripts/notify_suppliers_of_brief_withdrawal.py
Wynndow: Is this not an issue?

dmscripts/notify_suppliers_of_brief_withdrawal.py
Wynndow: In the api client?

tests/test_notify_suppliers_of_brief_withdrawal.py
katstevens: Cool, much better 😎.  The first test should still be asserting the calls for the mocked `get_brief_response_emails` script method though?

katstevens: Boom 👍

PR #172, opened 2017-10-27
Adds in G-Cloud 9 index example

DilwoarH:


scripts/index-services.py
```diff
@@ -15,8 +15,11 @@
     --search-api-url=<search-api-url>             Override search API URL (otherwise automatically populated)
     --search-api-token=<search_api_access_token>  Override search API token (otherwise automatically populated)
 
-Example:
+Example 1:
     ./index-services.py dev --index=g-cloud-9-2017-10-17 --frameworks=g-cloud-9
+
+Example 2:
+    ./index-services.py dev --api-token=myToken --search-api-token=myToken --frameworks=g-cloud-9 --index "g-cloud-9"
```
samuelhwilliams: We shouldn't need to manually specify the api tokens when indexing locally; did you have to? Ideally I think it would be good for us to get away from the idea of manually entering tokens. We have helpers that should allow us to automatically inject them as appropriate.

If we are keeping this, can we keep the parameters in the same order/format to be consistent? I.e. index first (using the key=val style rather than key 'val'), then frameworks, then optionals.

TheDoubleK: I agree, this example shouldn't be necessary. The index in general probably shouldn't be the same as the framework name - we use an alias for that.

galund:
based on the comments, suggest we close this?

TheDoubleK:
Yep.

PR #174, opened 2017-11-08
When creating an index, a mapping name is now required.

galund:
This is the minimum change we need in order to carry on supporting G-Cloud indexes (i.e. being able to create them). A script of some sort (potentially a refactoring of this one, i.e. the G-Cloud services indexing script) will follow for indexing DOS opportunities.

https://trello.com/c/5oIKvS1s/100-search-service-for-briefs-plus-indexing-script


TheDoubleK:
I think Sam beat you to the draw on this one.  He did a more thorough job than you 😛 
for the record, I would have done it this way too 😄

galund:
We'd better discuss

PR #175, opened 2017-11-09
Update indexing script

samuelhwilliams:
## Summary
Update index-services script to have a default mapping (for services) to remain compatible after updates to the search-api/apiclient


TheDoubleK:
I might be missing something, but isn't the "**index-services**" script always going to index `services`?  Why parameterise something that will always be `services`?

PR #176, opened 2017-11-24
Generalize the indexer for briefs

galund:
Generalize the indexer to work with briefs as well as services.
 - must specify document type (`doc-type`) on command line;
 - index and mapping are not updated unless `--create-with-mapping` is specified.

This is not backwards-compatible so we'll have to look at updating Jenkins jobs etc as part of planning the release for this.

https://trello.com/c/5oIKvS1s/100-search-service-for-briefs-plus-indexing-script

### DIFF
--- a/scripts/index-to-search-service.py
+++ b/scripts/index-to-search-service.py
@@ -2,20 +2,18 @@
 """Read services from the API endpoint and write to search-api for indexing.
 
 Usage:
-    index-to-search-service.py <doc-type> <stage> [--create] --index=<index> --frameworks=<frameworks> \
---mapping=<mapping> [options]
+    index-to-search-service.py <doc-type> <stage> --index=<index> --frameworks=<frameworks> [options]
 
     <doc-type>                                    One of briefs or services
     <stage>                                       One of dev, preview, staging or production
-    --create                                      Use this flag to create the index if it doesn't exist. If it does
+    --index=<index>                               Search API index name, usually of the form <framework>-YYYY-MM-DD
+    --frameworks=<frameworks>                     Comma-separated list of framework slugs that should be indexed
+
+Options:
+    --create-with-mapping=<mapping>               Specify a mapping to create the index if it doesn't exist. If it does
                                                   exist, the mapping for the index will be updated. Don't specify this
                                                   option when running from a batch/Jenkins job, as both creating indexes
                                                   and updating mappings should be done under user control.
-    --index=<index>                               Search API index name, usually of the form <framework>-YYYY-MM-DD
-    --frameworks=<frameworks>                     Comma-separated list of framework slugs that should be indexed
-    --mapping=<mapping>                           One of the mappings supported by the Search API.
-
-Options:
     --serial                                      Do not run in parallel (useful for debugging)
     --api-url=<api-url>                           Override API URL (otherwise automatically populated)
     --api-token=<api_access_token>                Override API token (otherwise automatically populated)
@@ -24,7 +22,7 @@ Options:
 
 Example:
     ./index-to-search-service.py services dev --index=g-cloud-9-2017-10-17 --frameworks=g-cloud-9 \
---mapping=g-cloud-9-services
+--create-with-mapping=services-g-cloud-9
 """
 
 import sys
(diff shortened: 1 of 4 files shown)


scripts/index.py
```diff
@@ -57,63 +48,95 @@ def print_progress(counter, start_time):
...
-            result = client.create_index(self.index, mapping=mapping)
+            result = self.search_client.create_index(self.index, mapping=mapping)
             logger.info("Index creation response: {response}", extra={'response': result})
         except dmapiclient.HTTPError as e:
             if 'already exists as alias' in str(e.message):
                 logger.info("Skipping index creation for alias {index}", extra={'index': self.index})
             else:
                 raise
 
+    def request_items(self, frameworks):
```
Wynndow: For my own understanding, why do we need to define this here if it's defined in both the sub classes? Is it an OOP thing?

galund: This is just a way of declaring that it's a required part of the interface. There are other ways of doing that too, but this often seems the easiest. It's a way of documenting that if you are going to make a new subclass, then you have to have this method, is all.

scripts/index.py
```diff
@@ -2,23 +2,23 @@
 """Read services from the API endpoint and write to search-api for indexing.
 
 Usage:
-    index-services.py <stage> --index=<index> --frameworks=<frameworks> [options]
+    index.py <doc-type> <stage> [--create] --index=<index> --frameworks=<frameworks> --mapping=<mapping> [options]
```
risicle: `index-to-search-api.py`? Index can mean many things across multiple services.

scripts/index.py
```diff
@@ -2,23 +2,23 @@
...
-                                                  [default: services]
     --serial                                      Do not run in parallel (useful for debugging)
     --api-url=<api-url>                           Override API URL (otherwise automatically populated)
     --api-token=<api_access_token>                Override API token (otherwise automatically populated)
     --search-api-url=<search-api-url>             Override search API URL (otherwise automatically populated)
     --search-api-token=<search_api_access_token>  Override search API token (otherwise automatically populated)
 
 Example:
-    ./index-services.py dev --index=g-cloud-9-2017-10-17 --frameworks=g-cloud-9
+    ./index.py services dev --index=g-cloud-9-2017-10-17 --frameworks=g-cloud-9
```
risicle: :+1: having an example - I should do this more

scripts/index.py
galund: Dunno, right now there is only one way to index these things, and so it's speculation as to whether _if_ our services had multiple places to be indexed, doing that would be a _separate_ script?

risicle: Well, I mean `index` doesn't just mean search api. Also used to mean postgres indexes. Or index as in "generate an index page". Or index as in `list.index()`. Just a very generic term. Not quite as bad as `type` but getting there.

scripts/index.py
```diff
@@ -57,63 +48,95 @@ def print_progress(counter, start_time):
...
-def do_index(search_api_url, search_api_access_token, data_api_url, data_api_access_token, mapping, serial, index,
-             frameworks):
+    def include_in_index(self, item):
+        # Even draft briefs will be in the index, for now at least
+        return True
+
+
+class ServiceIndexer(IndexerBase):
+    def request_items(self, frameworks):
+        # despite the name, frameworks takes a string containing a comma-separated list of framework slugs
```
risicle: > despite the name, frameworks takes a string containing

Think you mean __framework__ (the name of the kwarg).

scripts/index-to-search-service.py
```diff
@@ -2,23 +2,28 @@
...
-                                                  [default: services]
     --serial                                      Do not run in parallel (useful for debugging)
     --api-url=<api-url>                           Override API URL (otherwise automatically populated)
     --api-token=<api_access_token>                Override API token (otherwise automatically populated)
     --search-api-url=<search-api-url>             Override search API URL (otherwise automatically populated)
     --search-api-token=<search_api_access_token>  Override search API token (otherwise automatically populated)
 
 Example:
-    ./index-services.py dev --index=g-cloud-9-2017-10-17 --frameworks=g-cloud-9
+    ./index-to-search-service.py services dev --index=g-cloud-9-2017-10-17 --frameworks=g-cloud-9
```
Wynndow: I think this example needs a `--mapping`

PR #189, opened 2018-01-23
index-to-search-service.py: update --help/doc to reflect new behaviour of api

risicle:
See https://github.com/alphagov/digitalmarketplace-search-api/pull/135 and https://trello.com/c/2wYwttWZ/171-500s-on-buyer-frontend-for-g-cloud-search-page-numbers-above-100

The api will now throw an error if attempting to create an already-existing index instead of making any attempt at updating the mapping.

### DIFF
--- a/scripts/index-to-search-service.py
+++ b/scripts/index-to-search-service.py
@@ -10,11 +10,11 @@ Usage:
     --frameworks=<frameworks>                     Comma-separated list of framework slugs that should be indexed
 
 Options:
-    --create-with-mapping=<mapping>               Specify a mapping to create the index if it doesn't exist. If it does
-                                                  exist, the mapping for the index will be updated. Don't specify this
-                                                  option when running from a batch/Jenkins job, as both creating indexes
-                                                  and updating mappings should be done under user control. This mapping
-                                                  is a filename (sans .json suffix) found in search-api/mappings
+    --create-with-mapping=<mapping>               Create the named index using mapping <mapping>. Don't specify this
+                                                  option when running from a batch/Jenkins job, as creating indexes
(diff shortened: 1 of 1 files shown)


scripts/index-to-search-service.py
```diff
@@ -10,11 +10,11 @@
...
-    --create-with-mapping=<mapping>               Specify a mapping to create the index if it doesn't exist. If it does
-                                                  exist, the mapping for the index will be updated. Don't specify this
-                                                  option when running from a batch/Jenkins job, as both creating indexes
-                                                  and updating mappings should be done under user control. This mapping
-                                                  is a filename (sans .json suffix) found in search-api/mappings
+    --create-with-mapping=<mapping>               Create the named index using mapping <mapping>. Don't specify this
+                                                  option when running from a batch/Jenkins job, as creating indexes
+                                                  should be done under user control. This mapping is a filename (without
+                                                  .json suffix) as would be found by the search-api in its
+                                                  digitalmarketplace-search-api/mappings directory.
```
katstevens: Just to double check (as I'm still getting up to speed with all this), does this option still apply in the two examples below? Will we ever call this script without the `create-with-mapping` option?

galund: The overnight index job uses the existing index, or an alias of it (i.e. we'll be overwriting existing records in ES) - so yes we will.

Day-to-day, we don't use this option, only if we're changing the mapping, adding fields or (it turns out) changing settings. Or of course if we're adding a new index because of G-Cloud 10 or something like that.

risicle: What he said.

katstevens: Makes sense, thanks for clarifying 😸

PR #190, opened 2018-02-02
Include supplier contact details for dos labs export

idavidmcdonald:
https://trello.com/c/7msYuJ6c/184-links-for-dos-labs-suppliers-not-working

I had to put up a new version of the labs spreadsheet on https://www.digitalmarketplace.service.gov.uk/buyers/frameworks/digital-outcomes-and-specialists-2/requirements/user-research-studios

I used this script (which I assumed was how it was done in the past). I had to do a bunch of manual formatting to make it look pretty for an end user. I will details of this to the manual in the framework lifecycle section.

### DIFF
--- a/scripts/export-dos-labs.py
+++ b/scripts/export-dos-labs.py
@@ -11,14 +11,18 @@ import sys
 sys.path.insert(0, '.')
 
 from docopt import docopt
+from dmapiclient import DataAPIClient
+
 from dmscripts.helpers.env_helpers import get_api_endpoint_from_stage
 from dmscripts.helpers.framework_helpers import find_suppliers_with_details_and_draft_services
-from dmapiclient import DataAPIClient
+from dmscripts.export_dos_labs import append_contact_information_to_services
 
 if sys.version_info[0] < 3:
     import unicodecsv as csv
 else:
     import csv
+
+REQUIRED_CONTACT_DETAILS_KEYS = ['email', 'contactName', 'phoneNumber']
 
 
 def find_all_labs(client):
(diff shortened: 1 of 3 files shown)


scripts/export-dos-labs.py
```diff
@@ -11,25 +11,29 @@
...
+REQUIRED_CONTACT_DETAILS_KEYS = ['email', 'contactName', 'phoneNumber']
+
 
 def find_all_labs(client):
     records = find_suppliers_with_details_and_draft_services(client,
                                                              FRAMEWORK_SLUG,
                                                              lot="user-research-studios",
                                                              statuses="submitted"
                                                              )
+    records = append_contact_information_to_services(records, REQUIRED_CONTACT_DETAILS_KEYS)
```
TheDoubleK: Might be marginally more efficient to append contact information after filtering by onFramework?

tests/test_export_dos_labs.py
```diff
@@ -0,0 +1,30 @@
...
+            'services': [{'serviceName': 'service1'}, {'serviceName': 'service2'}],
+            'supplier': {'contactInformation': [{'email': 'me@me.com', 'other': 'value'}]}
+        },
+        {
+            'services': [{'serviceName': 'service3'}],
+            'supplier': {'contactInformation': [{'email': 'me2@me2.com', 'other': 'value'}]}
+        },
+    ]
+
+    res = append_contact_information_to_services(records, ['email'])
```
TheDoubleK: It might be nice to test with all three of the same keys we use in the script itself?  Or at least more than one key, so we are sure it isn't getting things mixed up.

idavidmcdonald: Will add another key

scripts/export-dos-labs.py
idavidmcdonald: Will change the order (no idea if it actually would be more efficient but no harm)

PR #193, opened 2018-02-07
Notify suppliers when a brief has been awarded

brenetic:
This script will run overnight to collate all brief responses awarded the previous day and notify them by email

https://trello.com/c/LsEbL3jP/306-send-dos-suppliers-an-email-to-let-them-know-the-outcome

### DIFF
--- a/dmscripts/notify_suppliers_of_awarded_briefs.py
+++ b/dmscripts/notify_suppliers_of_awarded_briefs.py
@@ -1,4 +1,4 @@
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, date
 
 from dmutils.email.exceptions import EmailError
 from dmutils.email.helpers import hash_string
@@ -13,7 +13,8 @@ def _create_context_for_brief(stage, brief):
             env_helpers.get_web_url_from_stage(stage),
             brief['framework']['family'],
             brief['id']
-        )
+        ),
+        'utm_date': date.today().strftime("%Y%m%d")
     }
 
 
--- a/dmscripts/notify_suppliers_of_awarded_briefs.py
+++ b/dmscripts/notify_suppliers_of_awarded_briefs.py
@@ -1,0 +1,99 @@
+from datetime import datetime, timedelta
+
+from dmutils.email.exceptions import EmailError
+from dmutils.email.helpers import hash_string
+from dmutils.formats import DATE_FORMAT
+from dmscripts.helpers import env_helpers
+
+
+def _create_context_for_brief(stage, brief):
+    return {
+        'brief_title': brief['title'],
+        'brief_link': '{}/{}/opportunities/{}'.format(
+            env_helpers.get_web_url_from_stage(stage),
+            brief['framework']['family'],
+            brief['id']
+        )
+    }
+
+
+def _get_brief_responses_to_be_notified(data_api_client, brief_response_ids, awarded_at):
+    brief_responses = []
+
+    if brief_response_ids:
+        # Used for re-running failed email sending
+        for brief_response_id in brief_response_ids:
+            brief_responses.append(data_api_client.get_brief_response(brief_response_id)['briefResponses'])
+    else:
+        awarded_at_date = awarded_at or (datetime.utcnow() - timedelta(days=1)).strftime(DATE_FORMAT)
+        awarded_brief_responses = data_api_client.find_brief_responses_iter(awarded_at=awarded_at_date)
+
+        for abr in awarded_brief_responses:
+            # Add the successful BriefResponse
+            brief_responses.append(abr)
+            # Get the unsuccessful BriefResponses
+            submitted_brief_responses = data_api_client.find_brief_responses_iter(
+                brief_id=abr['briefId'], status='submitted'
+            )
+            brief_responses.extend(submitted_brief_responses)
+
+    return brief_responses
+
+
+def _build_and_send_emails(brief_responses, mail_client, stage, dry_run, template_id, logger):
+    failed_brief_responses = []
+
+    # Now email everyone whose Brief got awarded
+    for brief_response in brief_responses:
+        email_address = brief_response["respondToEmailAddress"]
+        if not email_address:
+            continue
+
+        brief_email_context = _create_context_for_brief(stage, brief_response['brief'])
+        try:
+            if not dry_run:
+                mail_client.send_email(
+                    email_address, template_id, brief_email_context, allow_resend=False
+                )
+            logger.info(
+                "{dry_run}EMAIL: Award of Brief Response ID: {brief_response_id} to {email_address}",
+                extra={
+                    'dry_run': '[Dry-run] - ' if dry_run else '',
+                    'brief_response_id': brief_response['id'],
+                    'email_address': hash_string(email_address),
+                }
+            )
+        except EmailError:
+            # Log individual failures in more detail
+            logger.error(
+                "Email sending failed for BriefResponse {brief_response_id} (Brief ID {brief_id})",
+                extra={
+                    "brief_id": brief_response['brief']['id'],
+                    "brief_response_id": brief_response['id']
+                }
+            )
+            failed_brief_responses.append(brief_response['id'])
+
+    return failed_brief_responses
+
+
+def main(
+    data_api_client, mail_client, template_id, stage, logger,
+    dry_run=False, brief_response_ids=None, awarded_at=None
+):
+    brief_responses = _get_brief_responses_to_be_notified(data_api_client, brief_response_ids, awarded_at)
+
+    failed_brief_responses = _build_and_send_emails(brief_responses, mail_client, stage, dry_run, template_id, logger)
+
+    # Log a summary of failures at the end of the job
+    if failed_brief_responses:
+        logger.error(
+            "Email sending failed for the following {count} BriefResponses: {brief_response_ids}",
+            extra={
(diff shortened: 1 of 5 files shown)


tests/test_notify_suppliers_of_awarded_briefs.py
```diff
@@ -0,0 +1,192 @@
...
+    if awarded:
+        brief_response['awardedAt'] = '2018-01-01T23:59:59.999999Z'
+        brief_response['status'] = 'awarded'
+        brief_response["respondToEmailAddress"] = "lucky_winner_{}@example.com".format(id_)
+    if not valid_email:
+        brief_response['respondToEmailAddress'] = ""
+    return brief_response
+
+
+def test_create_context_for_brief():
```
galund: I'd probably `_` prefix this, if it's not intended to be called from outside this module

dmscripts/notify_suppliers_of_awarded_briefs.py
```diff
@@ -0,0 +1,86 @@
+from datetime import datetime, timedelta
+
+from dmutils.email.exceptions import EmailError
+from dmutils.email.helpers import hash_string
+from dmutils.formats import DATE_FORMAT
+from dmscripts.helpers import env_helpers
+
+
+def create_context_for_brief(stage, brief):
```
risicle: If the only reason you're not underscore-prefixing this function is because you're importing it for a test, _I'd argue_ you can still go ahead and prefix-underscore it. It'll make someone think twice before going and depending on it. And if someone _did_ decide to go and change it willy nilly as a result, the test would break, so that would be ok.

dmscripts/notify_suppliers_of_awarded_briefs.py
```diff
@@ -0,0 +1,86 @@
...
+        'brief_title': brief['title'],
+        'brief_link': '{}/{}/opportunities/{}'.format(
+            env_helpers.get_web_url_from_stage(stage),
+            brief['framework']['family'],
+            brief['id']
+        )
+    }
+
+
+def main(
```
risicle: I'd still go ahead and base the name of the function on what it does, even if that means giving it the same name as the module. Consider this - to use this, someone's probably going to be doing a `from dmscripts.notify_suppliers_of_awarded_briefs import main` - and then someone's going to have this annoyingly meaningless `main` floating around in their namespace...

scripts/notify-suppliers-of-awarded-briefs.py
```diff
@@ -0,0 +1,77 @@
...
+    --stage=<stage>                                       Stage to target
+    --api_token=<api_token>                               API Token
+    --govuk_notify_api_key=<govuk_notify_api_key>         Notify API Token
+    --govuk_notify_template_id=<govuk_notify_template_id> Notify template id on account corresponding to token provided
+    --awarded_at=[<awarded_at>]                           Notify applicants to briefs awarded on this date, defaults to
+                                                          # yesterday (date format: YYYY-MM-DD)
+    --brief_response_ids=[<brief_response_ids>]           List of brief response IDs to send to (for resending failures)
+    --dry-run                                             List notifications that would be sent without sending emails
+    --verbose
+    -h, --help                                            Show this screen
```
katstevens: To make docopt happy:
```
Usage:
    notify-suppliers-of-awarded-briefs.py <stage> <api_token> <govuk_notify_api_key> <govuk_notify_template_id>
        [options]

Options:
    --awarded_at=<awarded_at>                       Notify applicants to briefs awarded on this date, defaults to
                                                    # yesterday (date format: YYYY-MM-DD)
    --brief_response_ids=<brief_response_ids>       List of brief response IDs to send to (for resending failures)
   ...
```

dmscripts/notify_suppliers_of_awarded_briefs.py
```diff
@@ -0,0 +1,86 @@
...
+    data_api_client, mail_client, template_id, stage, logger,
+    dry_run=False, brief_response_ids=None, awarded_at=None
+):
+    failed_brief_responses, brief_responses = [], []
+
+    if brief_response_ids:
```
risicle: Could we have a comment above this block please to describe what this is trying to do?

brenetic: There is the `# Used for re-running failed email sending` comment below?

dmscripts/notify_suppliers_of_awarded_briefs.py
```diff
@@ -0,0 +1,86 @@
...
+
+        for abr in awarded_brief_responses:
+            # Add the successful BriefResponse
+            brief_responses.append(abr)
+            # Get the unsuccessful BriefResponses
+            submitted_brief_responses = data_api_client.find_brief_responses_iter(
+                brief_id=abr['briefId'], status='submitted'
+            )
+            for br in submitted_brief_responses:
+                brief_responses.append(br)
```
risicle: y'can just do `brief_responses.extend(submitted_brief_responses)` y'know...

dmscripts/notify_suppliers_of_awarded_briefs.py
risicle: I basically just mean that the purpose of the entire block is just to populate `brief_responses` and that's not totally clear, but it could be given a comment.

dmscripts/notify_suppliers_of_awarded_briefs.py
risicle: (two can play at that game)

dmscripts/notify_suppliers_of_awarded_briefs.py
katstevens: Does `extend` play nice with the API client iterable?

dmscripts/notify_suppliers_of_awarded_briefs.py
```diff
@@ -0,0 +1,86 @@
...
+):
+    failed_brief_responses, brief_responses = [], []
+
+    if brief_response_ids:
+        # Used for re-running failed email sending
+        for brief_response_id in brief_response_ids:
+            brief_responses.append(data_api_client.get_brief_response(brief_response_id)['briefResponses'])
+    else:
+        awarded_at_date = awarded_at or (datetime.utcnow() - timedelta(days=1)).strftime(DATE_FORMAT)
+        awarded_brief_responses = data_api_client.find_brief_responses_iter(awarded_at=awarded_at_date)
```
risicle: So... if the script fails to run one night, those alerts are just never sent out, are they? Similarly, the script gives different results if it's run one minute before midnight or one minute after midnight? I guess I don't have a better answer without recording state back on the server somehow...

katstevens: We'll get a failure alert in Slack, and we can re-run the job for a specific date. 

Our other emailing scripts have the convention of running between 8am-9am about stuff that happened the day before (so the buyer has until 23:59:59.999999 UTC to award a brief in order for an email to get sent the next day).

dmscripts/notify_suppliers_of_awarded_briefs.py
brenetic: functionised 🐳

dmscripts/notify_suppliers_of_awarded_briefs.py
```diff
@@ -0,0 +1,99 @@
...
+    for brief_response in brief_responses:
+        email_address = brief_response["respondToEmailAddress"]
+        if not email_address:
+            continue
+
+        brief_email_context = _create_context_for_brief(stage, brief_response['brief'])
+        try:
+            if not dry_run:
+                mail_client.send_email(
+                    email_address, template_id, brief_email_context, allow_resend=False
```
risicle: What error does this function give us if it _does_ turn out to be a resend? Are we able to discern that error from a real error? Should we log it differently?

katstevens: Just to clarify, there aren't automatic resends. It will be a manual triggering of Jenkins with either the ID list or a date string. So we should be aware that something unusual is happening.

risicle:
On Wednesday 14 Feb 2018 16:37:57 Kat wrote:
> Does `extend` play nice with the API client iterable?
> 
> 

Works with all iterables

tests/test_notify_suppliers_of_awarded_briefs.py
```diff
@@ -0,0 +1,192 @@
...
+            _get_dummy_brief_response(4322, AWARDED_BRIEFS[0]),
+            _get_dummy_brief_response(4323, AWARDED_BRIEFS[0], valid_email=False),
+        ],
+        [   # No submitted BRs for the 2nd awarded brief
+        ]
+    ]
+    tested_script._create_context_for_brief.return_value = EXPECTED_BRIEF_CONTEXT
+
+    with freeze_time('2018-01-02'):
+        assert tested_script.main(data_api_client, notify_client, "notify_template_id", "preview", mock.Mock())
```
risicle: Interestingly (?) you're mocking the _class_ and then passing that in where the script would expect an _instance_ to be passed in. But it seems to work so I guess that's ok (?)

katstevens: Yeah, it's an interesting one. As we initialise the clients in the top level function we never assert the mocked classes directly, only their methods. We've followed the pattern used in our other recent scripts here by initialising clients at top level - I'm on the fence as to whether this is better off done in the testable code.

PR #194, opened 2018-02-13
Add one-off script for renaming supplier registration country

Wynndow:
Part of this trello card: https://trello.com/c/FeBysBI7

Currently suppliers that are registered in the UK have the string 'gb'
set in the `registration_country` field of the suppliers table.

As part of the new supplier sign up flow we're going to be collecting
their registration country via the country picker supplied by registers.

We need to bring our existing data in line with the data the country
picker will be supplying.

### DIFF
--- a/scripts/oneoff/rename-supplier-registered-country.py
+++ b/scripts/oneoff/rename-supplier-registered-country.py
@@ -43,7 +43,7 @@ def rename_country(client, dry_run):
                             {'registrationCountry': NEW_COUNTRY},
                             'rename supplier registered country script',
                         )
-                    )
+                    )()
                     success_counter += 1
                 except HTTPError as e:
                     print("Error updating supplier {}: {}".format(supplier['id'], e.message))
--- a/scripts/oneoff/rename-supplier-registered-country.py
+++ b/scripts/oneoff/rename-supplier-registered-country.py
@@ -13,25 +13,44 @@ Usage:
 import sys
 sys.path.insert(0, '.')
 
+import backoff
+import requests
 from docopt import docopt
-from dmapiclient import DataAPIClient
+from dmapiclient import DataAPIClient, HTTPError
+from dmapiclient.errors import HTTPTemporaryError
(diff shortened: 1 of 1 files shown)


scripts/oneoff/rename-supplier-registered-country.py
```diff
@@ -0,0 +1,45 @@
...
+
+def rename_country(client, dry_run):
+    counter = 0
+    for supplier in client.find_suppliers_iter():
+        if supplier.get('registrationCountry') == OLD_COUNTRY:
+            if not dry_run:
+                client.update_supplier(
+                    supplier['id'],
+                    {'registrationCountry': NEW_COUNTRY},
+                    'script',
```
TheDoubleK: Something more descriptive here maybe... `rename supplier registered country script`?

scripts/oneoff/rename-supplier-registered-country.py
```diff
@@ -0,0 +1,45 @@
...
+
+OLD_COUNTRY = "gb"
+NEW_COUNTRY = "country:GB"
+
+def rename_country(client, dry_run):
+    counter = 0
+    for supplier in client.find_suppliers_iter():
+        if supplier.get('registrationCountry') == OLD_COUNTRY:
+            if not dry_run:
+                client.update_supplier(
```
TheDoubleK: Depending on how risk-averse you're feeling we could do a backoff thing to re-try failed calls, something like this: https://github.com/alphagov/digitalmarketplace-scripts/blob/859f11f78124806c392f9f2ebd1f1e144f0c5caf/scripts/oneoff/migrate-supplier-data-from-declarations.py#L192

But as this can easily be re-run over again anyway maybe it's quicker to just try it and hope for the best.

Wynndow: Have added, might come in useful!

samuelhwilliams:
It would be nice to have this print out some stuff at the start/as it progresses, but also :man_shrugging:. We're only running this once.

PR #196, opened 2018-02-15
Notify suppliers for cancelled and unsuccessful Briefs

katstevens:
Trello: https://trello.com/c/LsEbL3jP/306-send-dos-suppliers-an-email-to-let-them-know-the-outcome

Extends the script to include Briefs that were marked as `unsuccessful` or `cancelled` on the same date. This means an extra query to the `/briefs` endpoint to get all briefs for those statuses (there shouldn't be many - currently less than 10 each since Oct 2017), then filtering those Briefs by date within the script. We then get the associated Brief Responses for those briefs as normal.

Retrying the script with the list of IDs should be unaffected.

### DIFF
--- a/dmscripts/notify_suppliers_of_awarded_briefs.py
+++ b/dmscripts/notify_suppliers_of_awarded_briefs.py
@@ -30,13 +30,22 @@ def _get_brief_responses_to_be_notified(data_api_client, brief_response_ids, awa
         awarded_brief_responses = data_api_client.find_brief_responses_iter(awarded_at=awarded_at_date)
 
         for abr in awarded_brief_responses:
-            # Add the successful BriefResponse
+            # Add the awarded BriefResponse
             brief_responses.append(abr)
-            # Get the unsuccessful BriefResponses
-            submitted_brief_responses = data_api_client.find_brief_responses_iter(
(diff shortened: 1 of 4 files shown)


dmscripts/notify_suppliers_of_awarded_briefs.py
```diff
@@ -30,13 +44,20 @@ def _get_brief_responses_to_be_notified(data_api_client, brief_response_ids, awa
...
-            # Get the unsuccessful BriefResponses
-            submitted_brief_responses = data_api_client.find_brief_responses_iter(
-                brief_id=abr['briefId'], status='submitted'
+            # Get the other non-awarded BriefResponses for the same Brief
+            brief_responses.extend(
+                data_api_client.find_brief_responses_iter(brief_id=abr['briefId'], status='submitted')
             )
-            brief_responses.extend(submitted_brief_responses)
+
+        # Get BriefResponses for any Briefs cancelled or unsuccessful on the same date
```
katstevens: Going to replace this query with the hopefully-more-scalable date filtering for unsuccessful/cancelled: https://github.com/alphagov/digitalmarketplace-apiclient/pull/122

PR #200, opened 2018-02-28
One off script for migrating supplier organisation size

katstevens:
Prep work needed for https://trello.com/c/PZyZdMtG/337-custom-dimension-to-capture-size-of-supplier

As this is a one-off script this hopefully won't need tests, but I can add them if necessary 😄 

I've adapted the `migrate-supplier-data-from-declarations` script to just update one supplier field, `organisation_size` (the retry/backoff stuff is all copied from that script). 

For each supplier, we get the list of their frameworks and choose the most recent non-null declaration to get the `organisationSize` from.

Running this locally took around 8 minutes, I'd expect slightly longer on preview.

### DIFF
--- a/scripts/oneoff/migrate-supplier-organisation-size-from-declarations.py
+++ b/scripts/oneoff/migrate-supplier-organisation-size-from-declarations.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python
+"""Selectively migrate supplier organisation size from a recent framework declaration to the field on supplier itself.
+
+   Based on scripts/oneoff/migrate-supplier-data-from-declarations.py (retry/backoff explained there in more detail)
+
+Usage:
+    scripts/oneoff/migrate-supplier-organisation-size-from-declarations.py <stage> <data_api_token>
+        [<user>] [--dry-run]
+
+Positional arguments:
+    <stage>                                 API stage to perform operation on
+    <data_api_token>                        appropriate API token for <stage>
+
+Optional arguments:
+    -h, --help                              show this help message and exit
+    <user>                                  Audit Trail 'updated_by' user
+    --dry-run                               skip update step
+"""
+import sys
+sys.path.insert(0, '.')
+
+from datetime import datetime
+import logging
+import getpass
+
+import backoff
+from docopt import docopt
+import requests
+from dmapiclient import DataAPIClient, HTTPError
+from dmapiclient.errors import HTTPTemporaryError
+
+from dmscripts.helpers.env_helpers import get_api_endpoint_from_stage
+from dmscripts.helpers.logging_helpers import configure_logger
+from dmscripts.helpers.logging_helpers import INFO as loglevel_INFO
+
+
+logger = logging.getLogger("script")
+
+
+_backoff_wrap = backoff.on_exception(
+    backoff.expo,
+    (HTTPTemporaryError, requests.exceptions.ConnectionError, RuntimeError),
+    max_tries=5,
+)
+
+
+def _catch_404_none(func):
+    try:
+        return func()
+    except HTTPError as e:
+        if e.status_code == 404:
+            return None
+        else:
+            raise
+
+
+def _get_supplier_frameworks(data_api_client, supplier_id):
+    # Get all supplierFrameworks for this supplier that have a declaration
+    try:
+        supplier_frameworks = _catch_404_none(
+            _backoff_wrap(
+                lambda: data_api_client.get_supplier_frameworks(supplier_id)
+            )
+        )
+    except StopIteration:
+        logger.info("Supplier %s: not on any relevant frameworks", supplier_id)
+        return None
+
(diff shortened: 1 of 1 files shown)


scripts/oneoff/migrate-supplier-organisation-size-from-declarations.py
```diff
@@ -0,0 +1,128 @@
...
+        else:
+            raise
+
+
+def _get_supplier_frameworks(data_api_client, supplier_id):
+    # Get all supplierFrameworks for this supplier that have a declaration
+    try:
+        supplier_frameworks = _catch_404_none(
+            _backoff_wrap(
+                lambda: data_api_client.get_supplier_frameworks(supplier_id)
```
galund: Yeah, @risicle might tell you to use `functools.partial`, but eh :-)

scripts/oneoff/migrate-supplier-organisation-size-from-declarations.py
```diff
@@ -0,0 +1,128 @@
...
+        logger.info("Processing supplier %s", supplier["id"])
+        if supplier.get("organisationSize"):
+            logger.info("  already done: {}".format(supplier["id"]))
+            continue
+
+        supplier_frameworks = _get_supplier_frameworks(client, supplier['id'])
+
+        if supplier_frameworks:
+            if len(supplier_frameworks) == 1:
+                supplier_framework = supplier_frameworks[0]
```
galund: Feels like this case just adds to the complication - sorting a list of length one will be very little effort?

scripts/oneoff/migrate-supplier-organisation-size-from-declarations.py
```diff
@@ -0,0 +1,128 @@
...
+
+            logger.info(
+                "Supplier %s: updating with data from framework %s", supplier["id"], supplier_framework["frameworkSlug"]
+            )
+
+            # Get the organisation size
+            org_size = supplier_framework.get('declaration', {}).get('organisationSize')
+            logger.info('  Updating with org size {}'.format(org_size))
+
+            if org_size in ['micro', 'small', 'medium', 'large']:
```
galund: If you're gonna test, you might as well output a warning if it doesn't match? (Ignoring `None` I suppose, that case isn't very interesting.)

scripts/oneoff/migrate-supplier-organisation-size-from-declarations.py
```diff
@@ -0,0 +1,128 @@
...
+                # Update the supplier
+                _update_supplier_organisation_size(client, org_size, supplier['id'], user, dry_run)
+                updated_suppliers_count += 1
```
TheDoubleK: It might be interesting to have an "else" here, just to see if we have any unexpected data in declarations.  But really only out of interest.

scripts/oneoff/migrate-supplier-organisation-size-from-declarations.py
katstevens: All that stuff is copied from the the other script 😸

scripts/oneoff/migrate-supplier-organisation-size-from-declarations.py
katstevens: Yep - going to log it at least.

scripts/oneoff/migrate-supplier-organisation-size-from-declarations.py
katstevens: Yep - this bit of code was a bit gnarlier before, now I agree the check doesn't add much.

TheDoubleK:
One thought - rather than passing in the user who's running the script (or maybe as well as??) we could include the name of the script in the "updated_by".

We've definitely done that in the past I think.

scripts/oneoff/migrate-supplier-organisation-size-from-declarations.py
TheDoubleK: `None` would be interesting, because it's a required question in declarations, so shouldn't be any with `None`.

scripts/oneoff/migrate-supplier-organisation-size-from-declarations.py
galund: Oh well in that case maybe it should be in `dmscripts` :-)

scripts/oneoff/migrate-supplier-organisation-size-from-declarations.py
galund: Was it always required, or will there be people on DOS1 but not DOS2 that didn't fill it out perhaps (or G-Cloud 7 or something maybe?)

katstevens:
A dry run against preview took 44 mins, ouch. All seemed fine, 11707 suppliers would have been updated, 2817 had no declaration at all, 787 had a declaration with an org size of 'None' or something else not in the `['micro', 'small', 'medium', 'large']` list.

PR #201, opened 2018-03-05
This csv should contain the contact details of the supplier in question

benvand:
This script is often used to contact the supplier in question about their declaration or application.
We should include the proper contact details for the supplier.


benvand:
Awesome, thanks @MuriloDalRi ! 👍

PR #202, opened 2018-03-08
Fix timeouts when calling Mailchimp

Wynndow:
This is to pull in this update to the utils - https://github.com/alphagov/digitalmarketplace-utils/pull/365

We're pulling in the updated Mailchimp client which should mitigate timeouts when fetching email lists from Mailchimp.

All of the major versions don't affect the scripts repo.

Once this is merged the scripts docker image will need rebuilding and pushing up to Dockerhub.

### DIFF
--- a/requirements-app.txt
+++ b/requirements-app.txt
@@ -2,7 +2,7 @@
 # with package version changes made in requirements-app.txt
 
 git+https://github.com/madzak/python-json-logger.git@v0.1.5#egg=python-json-logger==v0.1.5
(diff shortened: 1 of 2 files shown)


risicle:
> All of the major versions don't affect the scripts repo.

Surprised by this - most of the major bumps around this time were email-sending-related weren't they?

Wynndow:
A few were, but they were to with with generating and decoding tokens for user creation and password reset, which the scripts repo doesn't do. Couple of others about using the odfpy fork and then not using it anymore. And one about bumping the docker base image, which scripts doesn't use.

PR #204, opened 2018-03-12
Remove unused set-search-alias script

brenetic:
We now use the `update-index-alias.py` script to update search index aliases.


brenetic:
@risicle I can't see it being used anyway in Jenkins

PR #206, opened 2018-03-16
Update script for Python3

samuelhwilliams:
## Summary
Have run 2to3 on scripts/make-g-cloud-live.py to check for basic
incompatibilities and none are reported, so this seems to just be
removing the `u""` strings mainly. Beyond that, it appears to run fine
on Python3, although testing it is somewhat without a framework+draft
services in a good state.

 ## Ticket
https://trello.com/c/mtiS5IAx/3-move-scripts-running-python-2-on-jenkins-to-python3-docker


scripts/make-g-cloud-live.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
```
CrystalPea: does this line tell it to run in python3?

samuelhwilliams: If you call it directly with `./<script name>`, it'll make it look for a Python3 interpreter, yeah. You can still force it to run in Python 2 though. This is called a shebang and is interpreted by bash when executing the file.

CrystalPea: Ahh, awesome, that's a good explanation, thank you :)

samuelhwilliams: I slightly lied. It's not interpreted by bash, it's interpreted by the 'program loader'. I don't know what that is. But more details and links off from here: https://en.wikipedia.org/wiki/Shebang_(Unix) :)

Dockerfile
```diff
@@ -4,7 +4,7 @@ ENV APP_DIR /app
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-                    gcc g++ make git libffi-dev libssl-dev libc6-dev
+                    gcc g++ make git libffi-dev libssl-dev libc6-dev jq==1.5
```
galund: Why do we need `jq` now? I mean it's useful - but not obviously used here?

requirements-app.txt
```diff
@@ -13,6 +13,8 @@ jsonschema==2.5.1
 lorem==0.1.1
 mailchimp3==2.0.11
 pandas==0.19.1
+pygithub==1.38
 python-dateutil==2.4.2
 unicodecsv==0.14.1
 xeger==0.3.1
+yq==2.4.1
```
galund: Can't see the direct dep on yq either?

samuelhwilliams: Gone. I think that got slurped in from my other PR for rotating creds.

Dockerfile
samuelhwilliams: Same here.

scripts/make-g-cloud-live.py
```diff
@@ -54,53 +52,52 @@ def parse_document_url(url, framework_slug):
...
         if dry_run:
-            print(u"    > dry run: skipped copying {}".format(message_suffix))
+            print("    > dry run: skipped copying {}".format(message_suffix))
         else:
-            key = documents_bucket.bucket.copy_key(live_document_path,
-                                                   src_bucket_name=draft_bucket.bucket_name,
-                                                   src_key_name=draft_document_path)
-            key.set_acl('public-read')
-            print(u"    > copied {}".format(message_suffix))
+            documents_bucket.copy(src_bucket=draft_bucket.bucket_name, src_key=draft_document_path,
```
katstevens: Ahhhh boto3 is so much nicer 😻

PR #207, opened 2018-03-16
Update generate counterpart PDFs

samuelhwilliams:
## Summary
Check
`scripts/generate-framework-agreement-counterpart-signature-pages.py`
for compatibility with Python 3 and remove some old scripts that have
been superseded by more generic versions.

 ## Ticket
https://trello.com/c/mtiS5IAx/3-move-scripts-running-python-2-on-jenkins-to-python3-docker


TheDoubleK:
Part of me is nervous about deleting this script, but we *really* shouldn't need to generate any G8 agreements these days, so it's probably fine to delete it.

And it lives in git history anyway I guess.

PR #208, opened 2018-03-24
Automated API token updates

samuelhwilliams:
## Summary
Adds a script that allows easier rotation of tokens for our data and
search API. The script will add three new tokens, or remove all except
the top three, commit the new tokens, push them to GitHub and open a
Pull Request for review. This will integrate with a Jenkins pipeline job
that will add new tokens, re-deploy all apps, then remove the old tokens
and re-deploy all apps to hopefully make it far faster and less
error-prone to cycle tokens.

## Ticket
https://trello.com/c/4crGe6F0/169

### DIFF
--- a/scripts/rotate-api-tokens.sh
+++ b/scripts/rotate-api-tokens.sh
@@ -1,0 +1,106 @@
+#!/bin/bash
+# This script will update Data and Search API tokens for a given environment. It does this by checking out the
+# dm-credentials repository, decrypting the relevant creds files and using 'yq' (a yaml CLI editor) to inject new
+# tokens. These are committed and pushed up, then the GitHub API is used to create a pull request for the developer
+# to manually approve.
+#
+# If running locally you will need to provide AWS_PROFILE=sops or equivalent.
+# If running locally through Docker, you will probably also need to mount your ~/.aws folder at `/root/.aws` for SOPS to be able to decrypt credentials. But be careful with this due to file permissions.
+# If running on Jenkins, AWS allows SOPS to decrypt credentials using its EC2 instance profile - even within Docker.
+#
+# Syntax: ./scripts/rotate-api-tokens.sh [add-new|remove-old] [preview|staging|production]
+
+set -e
+
+COMMAND=$(echo "${1}" | tr "[:lower:]" "[:upper:]")
+STAGE=$(echo "${2}" | tr "[:lower:]" "[:upper:]")
+STAGE_LOWER=$(echo "${STAGE}" | tr "[:upper:]" "[:lower:]")
+GIT_CREDS_UPDATE_BRANCH_NAME=$(date +"%Y-%m-%dT%H-%M-%S-credentials-update")
+
+if [ -z "${GITHUB_ACCESS_TOKEN}" ]; then
+  echo "Required environment variable GITHUB_ACCESS_TOKEN is undefined."
+  exit 1
+fi
+
+if [[ "${STAGE}" != "PREVIEW" && "${STAGE}" != "STAGING" && "${STAGE}" != "PRODUCTION" ]]; then
+  echo "Syntax: ./scripts/rotate-api-tokens.sh [add-new|remove-old] [preview|staging|production]"
+  exit 2
+fi
+
+function generate_api_token() {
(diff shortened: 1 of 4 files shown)


scripts/update-api-tokens.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
```
idavidmcdonald: Could we add an overview at the top of the file to explain what this does please without someone needing to actually scroll through the code

scripts/update-api-tokens.sh
```diff
@@ -0,0 +1,82 @@
...
+  echo "Syntax: ./scripts/update-api-tokens.sh [add-new|remove-old] [preview|staging|production]"
+  exit 1
+fi
+
+function generate_api_token() {
+  echo $(python3 -c "import random, string; print(''.join(random.SystemRandom().choice(string.ascii_letters + string.digits) for _ in range(56)))")
+}
+
+function create_credentials_update_branch() {
+  rm -rf digitalmarketplace-credentials
```
idavidmcdonald: Should we do something like .gitignore `digitalmarketplace-credentials` or get this script to remove it after it's run?

scripts/update-api-tokens.sh
idavidmcdonald: Could also be good to rename this script something more explicit about how this creates prs rather than actually rolling credentials.

scripts/update-api-tokens.sh
samuelhwilliams: Hm interesting. We could do - although in theory this should only run inside a Docker container, so it won't be left behind anyway. If a developer runs this locally outside of Docker, i guess it could get left around, but that shouldn't be done and in that case I'm kinda happy leaving it there to spite them for doing it improperly... 🙈

scripts/update-api-tokens.sh
samuelhwilliams: 👍

scripts/update-api-tokens.sh
Wynndow: In what scenario would there be a credentials dir to remove here?

scripts/update-api-tokens.sh
```diff
@@ -0,0 +1,82 @@
...
+  git clone https://${GITHUB_ACCESS_TOKEN}@github.com/alphagov/digitalmarketplace-credentials.git
```
Wynndow: Might be worth quoting variable expansions here.

scripts/update-api-tokens.sh
samuelhwilliams: Hopefully none. Maybe if a dev ran it locally a few times :trollface:.

Wynndow: Better safe than sorry...

scripts/update-api-tokens.sh
samuelhwilliams: Ah good catch. Should all be quoted now.

samuelhwilliams:
Updated the script so that it also injects the new tokens into the relevant variables defined in jenkins-vars/jenkins.yaml:jenkins_env_variables.

scripts/rotate-api-tokens.sh
```diff
@@ -0,0 +1,106 @@
...
+  echo "Created PR#$(echo "${response_data}" | jq -crM '.number') on alphagov/digitalmarketplace-credentials."
+  echo "https://www.github.com/alphagov/digitalmarketplace-credentials/pull/$(echo "${response_data}" | jq -crM '.number')"
+}
+
+function add_new_tokens() {
+  create_credentials_update_branch
+
+  echo "Adding three new tokens (Application, Jenkins, Developer) to the Data-API and Search-API for ${STAGE}."
+
+  # Insert three new tokens for the api and search-api, and update Jenkins config with the new jenkins token.
```
idavidmcdonald: Wonder if it's worth explicitly mentioning that they are being inserted at the top (it is fairly well documented for the removal case that it's not the first three)

samuelhwilliams: Ugh missed this comment before putting this in. 🤦‍♂️

PR #209, opened 2018-03-26
Update utils for MailChimp client

samuelhwilliams:
## Summary
Pull in latest dm-utils which updates the MailChimp client to use a
generator and have a smaller page size when retrieving email addresses
from an existing list, which should hopefully result in fewer 504
timeouts with MailChimp servers.

 ## Ticket
https://trello.com/c/4cK1wTvq/236

### DIFF
--- a/dmscripts/upload_dos_opportunities_email_list.py
+++ b/dmscripts/upload_dos_opportunities_email_list.py
@@ -10,10 +10,6 @@ def find_user_emails(supplier_users, services):
     return email_addresses
 
 
-def lowercase_list(x):
-    return [item.lower() for item in x]
-
-
 def main(data_api_client, mailchimp_client, lot_data, logger):
     logger.info(
         "Begin mailchimp email list updating process for {} lot".format(lot_data["lot_slug"]),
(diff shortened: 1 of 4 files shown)


dmscripts/upload_dos_opportunities_email_list.py
```diff
@@ -31,14 +27,14 @@ def main(data_api_client, mailchimp_client, lot_data, logger):
         "{} emails have been found for lot {}".format(len(emails), lot_data["lot_slug"])
     )
 
-    existing_mailchimp_emails = mailchimp_client.get_email_addresses_from_list(lot_data["list_id"])
+    existing_mailchimp_emails = list(mailchimp_client.get_email_addresses_from_list(lot_data["list_id"]))
```
galund: Do you actually need to convert it to a list here?..

samuelhwilliams: Sadly it needs to be all slurped into memory at a later point to do a set difference with the known DM-side emails, so I figured it was easiest to just do it upfront than a couple lines later.

dmscripts/upload_dos_opportunities_email_list.py
```diff
@@ -31,14 +27,14 @@ def main(data_api_client, mailchimp_client, lot_data, logger):
...
     logger.info(
         "{} existing emails found on list {}".format(len(existing_mailchimp_emails), lot_data["list_id"])
     )
 
     # lowercase required because mailchimp may capatalise emails addresses returned from the mailchimp API based on it's
     # on how it stores both the lowercase hash but also the original (potentially capatilised) email address
-    new_emails = list(set(lowercase_list(emails)) - set(lowercase_list(existing_mailchimp_emails)))
+    new_emails = set(map(lambda x: x.lower(), emails)) - set(map(lambda x: x.lower(), existing_mailchimp_emails))
```
galund: And if you used `set.difference` here, you wouldn't need to construct the second set *at all*. (Although again this probably won't make any difference, if we're optimising...)

Also the first set could be a frozenset :)

dmscripts/upload_dos_opportunities_email_list.py
samuelhwilliams: We also iterate over the list twice which makes it harder to use a generator. 🙈

samuelhwilliams:
Updated based on comments @galund. Downside is having to remove one message logging the number of email addresses currently on the mailchimp list as it would consume the email addresses before we get around to passing them into mailchimp. Doesn't feel like this is _that_ useful.

dmscripts/upload_dos_opportunities_email_list.py
```diff
@@ -26,19 +22,16 @@ def main(data_api_client, mailchimp_client, lot_data, logger):
         framework=lot_data["framework_slug"], lot=lot_data["lot_slug"]
     )
 
-    emails = find_user_emails(supplier_users, framework_services)
+    emails = frozenset(map(lambda x: x.lower(), find_user_emails(supplier_users, framework_services)))
```
risicle: `frozenset(x.lower() for x in find_user_emails(supplier_users, framework_services))` is a little neater.

samuelhwilliams: 👍

PR #210, opened 2018-03-26
Auth token from environment

galund:
Use auth token from environment, if available.
 - this means we can call scripts that use this function from a docker container without having to make the credentials repo / decryption process available to that container.
 - we will be using a docker container to run all scripts in Python 3 from now on.

https://trello.com/c/yNVT3P2Y/227-updating-index-alias-script-to-python-3


dmscripts/helpers/auth_helpers.py
```diff
@@ -9,17 +9,21 @@ def get_auth_token(api, stage):
...
-    creds = subprocess.check_output([
-        "{}/sops-wrapper".format(DM_CREDENTIALS_REPO),
-        "-d",
-        "{}/vars/{}.yaml".format(DM_CREDENTIALS_REPO, stage)
-    ])
-    auth_tokens = yaml.load(creds)[api]['auth_tokens']
-    auth_token = [token for token in auth_tokens if token.startswith(token_prefix)]
-
-    return auth_token[0]
+    env_var = "DM_{}_TOKEN_{}".format(api.upper(), stage.upper())
```
katstevens: Am I right in assuming these env vars will be set when the Docker image is created, so if we recycle the tokens (as per Sam's new Jenkins job) the new tokens would get picked up in the app redeploy, as they would when fetching them the credentials repo below?

galund: They should be set by the `docker run` call - there's no separate deployment stage here, Jenkins creates the docker container on-the-fly as far as I understand it. (Though it will have cached the image, so that should be quick, the environment variables are passed in with `docker run`.)

TLDR yes I think new creds will get picked up.

dmscripts/helpers/auth_helpers.py
```diff
@@ -9,17 +9,21 @@ def get_auth_token(api, stage):
...
+
+    if not auth_token:
+        DM_CREDENTIALS_REPO = os.environ.get('DM_CREDENTIALS_REPO', '../digitalmarketplace-credentials')
+        token_prefix = 'J' if subprocess.check_output(["whoami"]) == 'jenkins' else 'D'
+        creds = subprocess.check_output([
+            "{}/sops-wrapper".format(DM_CREDENTIALS_REPO),
+            "-d",
+            "{}/vars/{}.yaml".format(DM_CREDENTIALS_REPO, stage)
+        ])
+        auth_tokens = yaml.safe_load(creds)[api]['auth_tokens']
```
galund: Have updated the yaml loading, thanks @MuriloDalRi

MuriloDalRi: 👍 I've added a few PRs to change the other apps as well

PR #211, opened 2018-03-27
Use yaml.safe_load

MuriloDalRi:


galund:
Fixed this in https://github.com/alphagov/digitalmarketplace-scripts/pull/210

PR #212, opened 2018-03-28
boto to boto3

samuelhwilliams:
## Summary
Some of the scripts are still trying to use the `boto` library. dm-utils
has been updated to use boto3 and create its own S3ResponseError, so
let's update this to do the same.

### DIFF
--- a/scripts/upload-counterpart-agreements.py
+++ b/scripts/upload-counterpart-agreements.py
@@ -45,9 +45,8 @@ from dmscripts.helpers.logging_helpers import logging
 
 from docopt import docopt
 from dmapiclient import DataAPIClient, APIError
-from boto.exception import S3ResponseError
 
-from dmutils.s3 import S3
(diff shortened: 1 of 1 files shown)


katstevens:
Grepping `boto` brings up this - does that need updating too?
```
scripts/make-g-cloud-live.py:@backoff.on_exception(backoff.expo, (dmapiclient.HTTPError, boto.exception.S3ResponseError, ), max_tries=5)
```

samuelhwilliams:
Ask David 😛

idavidmcdonald:
Yes and no.

It likely would if we want to get rid of our boto usage in all our scripts. However we are just going to try and fix this script by moving it to phython 3 as it's broken on jenkins at the moment.

We thought about fixing the other and then were weary that it was probably being run in python2 at the moment and decided not to try and fix that too at this point but hopefully should come up with gcloud is made live...

PR #213, opened 2018-03-29
Remove boto import

samuelhwilliams:
## Summary
This script needs to be compatible with Python3, which means not using
boto (but instead, boto3 as used by utils). The `S3ResponseError` has
been abstracted by the s3 module in dmutils, so now we import from there
instead.


samuelhwilliams:
https://github.com/alphagov/digitalmarketplace-scripts/pull/206

PR #215, opened 2018-04-20
Add script for resetting supplier declaration

samuelhwilliams:
## Summary
If suppliers were affected by not being able to copy their answers from
G-Cloud 9, we need to reset their declaration so that they have to click
'Start declaration' again, at which point they will get the opportunity
to re-use answers from G9.

### DIFF
--- a/scripts/oneoff/reset-supplier-declaration.py
+++ b/scripts/oneoff/reset-supplier-declaration.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python3
+
+"""
+A one-off script for resetting a supplier's declarations after a temporary bug affected the G-Cloud 10 application
+process, preventing suppliers from copying over answers from their G-Cloud 9 declaration.
+
+You can reset a declaration either by providing the email address of a user associated with that supplier, or by
+directly using the supplier's id.
+
+Syntax: ./scripts/oneoff/reset-supplier-declaration.py --stage preview --email 12345@user.marketplace.team
+        ./scripts/oneoff/reset-supplier-declaration.py --stage preview --supplier-id 123456
+"""
+
+import getpass
+import sys
+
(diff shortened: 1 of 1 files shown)


scripts/oneoff/reset-supplier-declaration.py
```diff
@@ -0,0 +1,83 @@
...
+            exit(1)
+
+        user_supplier_id = user['users']['supplier']['supplierId']
+        if user_supplier_id and supplier_id and user_supplier_id != supplier_id:
+            print(f'Email address provided does not match with supplier provided. Email address `{email}` is '
+                  f'associated with supplierId `{supplier_id}`. Script was called with supplierId `{supplier_id}`.')
+            exit(2)
+
+    try:
+        supplier_framework = data_api_client.get_supplier_framework_info(supplier_id=supplier_id,
```
Wynndow: If they pass in an email but not a supplier_id, this will fail. You need to set `supplier_id`

scripts/oneoff/reset-supplier-declaration.py
```diff
@@ -0,0 +1,85 @@
...
+        exit(3)
+
+    if not supplier_framework:
+        print(f'No supplier framework/interest record found for supplierId `{supplier_id}` on framework '
+              f'`{framework_slug}`.')
+        exit(4)
+
+    data_api_client.set_supplier_declaration(supplier_id=supplier_id, framework_slug=framework_slug, declaration={},
+                                             user=f'{getpass.getuser()} - {reason}')
+    data_api_client.set_supplier_framework_prefill_declaration(supplier_id=supplier_id, framework_slug=framework_slug,
```
Wynndow: Should catch these in case of HTTPError?

PR #216, opened 2018-04-24
Inject framework dates

samuelhwilliams:
## Summary
We are going to retire the `dates.yml` file from the
digitalmarketplace-frameworks repository in favour of storing this
information in the Framework model on the API. This script will inject
all the appropriate datetime information into existing framework
records.

 ## Ticket
https://trello.com/c/BVPgZRZc/65-store-datetimes-relating-to-the-framework-lifecycle-on-the-api

### DIFF
--- a/scripts/oneoff/inject-framework-dates.py
+++ b/scripts/oneoff/inject-framework-dates.py
@@ -112,7 +112,7 @@ def inject_framework_dates(stage):
                                                   f'oneoff/inject-framework-dates.py')
 
         except Exception as e:
-            print(f'Failed with {e}')
+            print(f'Failed with {e} on {framework_slug}. Data: {framework_data}')
 
 
 if __name__ == '__main__':
--- a/scripts/oneoff/inject-framework-dates.py
+++ b/scripts/oneoff/inject-framework-dates.py
@@ -26,6 +26,7 @@ FRAMEWORKS_AND_DATES = {
         'applicationsCloseAtUTC': '2016-01-19T15:00:00Z',
         'intentionToAwardAtUTC': '2016-02-18T12:00:00Z',
         'frameworkLiveAtUTC': '2016-03-03T12:00:00Z',
+        'frameworkExpiresAtUTC': '2017-01-27T12:00:00Z',
     },
     'digital-outcomes-and-specialists-2': {
         'clarificationsCloseAtUTC': '2016-12-01T17:00:00Z',
@@ -33,6 +34,7 @@ FRAMEWORKS_AND_DATES = {
         'applicationsCloseAtUTC': '2016-12-14T17:00:00Z',
         'intentionToAwardAtUTC': '2017-01-16T12:00:00Z',
         'frameworkLiveAtUTC': '2017-01-27T12:00:00Z',
+        'frameworkExpiresAtUTC': '2018-01-29T12:00:00Z',
     },
     'digital-outcomes-and-specialists-3': {
         'clarificationsCloseAtUTC': '2525-01-01T12:00:00Z',
@@ -40,6 +42,7 @@ FRAMEWORKS_AND_DATES = {
         'applicationsCloseAtUTC': '2525-01-01T12:00:00Z',
         'intentionToAwardAtUTC': '2525-01-01T12:00:00Z',
         'frameworkLiveAtUTC': '2525-01-01T12:00:00Z',
+        'frameworkExpiresAtUTC': '2525-01-01T12:00:00Z',
     },
 
     # Begin Unix epoch time polyfill.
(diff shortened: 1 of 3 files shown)


scripts/oneoff/inject-framework-dates.py
```diff
@@ -0,0 +1,125 @@
...
+    data_api_token = get_auth_token('api', stage) if stage != 'development' else 'myToken'
+    data_api_client = DataAPIClient(get_api_url('api', stage), data_api_token)
+
+    for framework_slug, framework_data in FRAMEWORKS_AND_DATES.items():
+        print(f'Injecting dates for {framework_slug}: {framework_data}')
+        try:
+            data_api_client.update_framework(framework_slug=framework_slug,
+                                             data=framework_data,
+                                             user=f'{getpass.getuser()} - '
+                                                  f'digitalmarketplace-scripts/scripts/'
```
Wynndow: Do these need to be f strings? Not saying you should change, just interested if they need to be the same to concatenate them. I could probably just check for myself....

scripts/oneoff/inject-framework-dates.py
```diff
@@ -0,0 +1,125 @@
...
+                                                  f'oneoff/inject-framework-dates.py')
+
+        except Exception as e:
+            print(f'Failed with {e}')
```
Wynndow: Logging out the frameworks_slug might be useful.

scripts/oneoff/inject-framework-dates.py
samuelhwilliams: They don't, but PyCharm made the second string automatically and f-string based on the first line being one. /shrug

scripts/oneoff/inject-framework-dates.py
samuelhwilliams: 👍

scripts/oneoff/inject-framework-dates.py
```diff
@@ -26,20 +26,23 @@
         'applicationsCloseAtUTC': '2016-01-19T15:00:00Z',
         'intentionToAwardAtUTC': '2016-02-18T12:00:00Z',
         'frameworkLiveAtUTC': '2016-03-03T12:00:00Z',
+        'frameworkExpiresAtUTC': '1970-01-01T00:00:00Z',
```
Wynndow: Should `'frameworkExpiresAtUTC'` for DOS1 be `'2017-01-27T12:00:00Z'`? We turned DOS1 off and DOS2 on at the same time. I know it's not in the dates yml file but neither is live date.

scripts/oneoff/inject-framework-dates.py
```diff
@@ -26,20 +26,23 @@
...
     },
     'digital-outcomes-and-specialists-2': {
         'clarificationsCloseAtUTC': '2016-12-01T17:00:00Z',
         'clarificationsPublishAtUTC': '2016-12-07T17:00:00Z',
         'applicationsCloseAtUTC': '2016-12-14T17:00:00Z',
         'intentionToAwardAtUTC': '2017-01-16T12:00:00Z',
         'frameworkLiveAtUTC': '2017-01-27T12:00:00Z',
+        'frameworkExpiresAtUTC': '2018-01-29T12:00:00Z',
```
Wynndow: DOS2 already expired?

scripts/oneoff/inject-framework-dates.py
```diff
@@ -26,20 +26,23 @@
...
     },
     'digital-outcomes-and-specialists-3': {
         'clarificationsCloseAtUTC': '2525-01-01T12:00:00Z',
         'clarificationsPublishAtUTC': '2525-01-01T12:00:00Z',
         'applicationsCloseAtUTC': '2525-01-01T12:00:00Z',
         'intentionToAwardAtUTC': '2525-01-01T12:00:00Z',
         'frameworkLiveAtUTC': '2525-01-01T12:00:00Z',
+        'frameworkExpiresAtUTC': '1970-01-01T00:00:00Z',
```
Wynndow: If we're doing dummy dates for now it might make sense to match with the other dummy dates here?

scripts/oneoff/inject-framework-dates.py
samuelhwilliams: This is the date currently stored on the DOS2 framework.frameworkAgreementDetails.frameworkEndDate (i.e. what was put on the original framework agreements). It may be that we should update it, but I don't know what the 'correct' expiry date is. So this feels OK for the moment.

scripts/oneoff/inject-framework-dates.py
samuelhwilliams: Yep, sounds reasonable.

scripts/oneoff/inject-framework-dates.py
Wynndow: Ah ok. Did DOS2 get extended? If so maybe this was the original date. If that's the case maybe this should be updated (not now, just in general if this happens).

samuelhwilliams: Yeah it got extended. We should try to find out what the current expected end date is.

scripts/oneoff/inject-framework-dates.py
samuelhwilliams: Sounds good to me.

PR #218, opened 2018-05-02
Bump utils to 36.7.0 and notify client to 4.8.2

Wynndow:
The new utils has a fix for a bug in the DM Mailchimp client. It'll
catch and handle errors if a response has malformed or missing json
data. See here:
https://github.com/alphagov/digitalmarketplace-utils/pull/390

The Notify client has a fix for issues installing with pip v10.
Interestingly their internal tagging script has an issue which means
that installing 4.8.2 directly from Github won't work. We can instead
install from PyPI though.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,10 +3,9 @@
 # with package version changes made in requirements-app.txt
 
 git+https://github.com/madzak/python-json-logger.git@v0.1.5#egg=python-json-logger==v0.1.5
-git+https://github.com/alphagov/digitalmarketplace-utils.git@36.0.0#egg=digitalmarketplace-utils==36.0.0
+git+https://github.com/alphagov/digitalmarketplace-utils.git@36.7.0#egg=digitalmarketplace-utils==36.7.0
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -24,32 +24,31 @@ yq==2.4.1
...
 cffi==1.11.5
 chardet==3.0.4
-click==6.7
 colorama==0.3.7
 contextlib2==0.4.0
 cryptography==1.9
 docopt==0.6.2
 docutils==0.14
-Flask==0.12.2
+Flask==0.10.1
```
galund: A downgrade feels weird - intentional?

katstevens: We're pinning Flask to 0.10.1 in all the FE apps - I wonder why scripts was higher to begin with?

Wynndow: Yeah, I hadn't noticed that TBH. It's the pin in utils which is dictating the version here.

galund: Oh, I think it'll have been when @samuelhwilliams fixed it from not being pinned at all, he probably picked the latest non-1.0 version? I could have that completely wrong. Consistency with the frontends makes sense, as long as nothing now breaks as a result :-)

PR #227, opened 2018-05-24
Virus scanner - only load timestamps if !!'since'

Wynndow:
We only need timestamps if we're scanning files modified from a certain
date. Loading timestamps can take some time.

### DIFF
--- a/scripts/virus-scan-s3-bucket.py
+++ b/scripts/virus-scan-s3-bucket.py
@@ -32,12 +32,13 @@ def virus_scan_bucket(s3, clam, prefix, since, dry_run=True):
     # s3.list can block for a significant amount of time against large buckets when loading with timestamps. During
     # this time there will be no output.
     logger.info('Gathering objects. This may take some time depending on the size of the bucket ...')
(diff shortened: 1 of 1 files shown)


scripts/virus-scan-s3-bucket.py
```diff
@@ -27,12 +27,13 @@
 
 
 def virus_scan_bucket(s3, clam, prefix, since, dry_run=True):
+    import pdb; pdb.set_trace()
```
samuelhwilliams: :trollface:

Wynndow: It's a feature. Not a bug....

samuelhwilliams:
On the line further down, `logger.info(f'Processing file from {file_meta["last_modified"]}: {file_meta["path"]}')`, does `last_modified` definitely exist if we don't load timestamps?

Wynndow:
hmm....no. Good point.

Wynndow:
@samuelhwilliams Have updated

samuelhwilliams:
Yep, just testing it out

PR #228, opened 2018-05-25
Update generation scripts with lot slugs

samuelhwilliams:
## Summary
We broke the interface into `get_csv_rows` to add a required parameter
of `framework_lot_slugs`. Some other scripts depended on this interface
and so need updating.


samuelhwilliams:
Thanks 😄

PR #230, opened 2018-06-01
Add an entrypoint to forward API ports from Docker

lfdebrux:
Ticket: [Trello](https://trello.com/c/R5tAKXEA/301-testing-dm-scripts-against-local-environment-within-the-docker-container-is-broken)

This is useful for macOS development environments, if services are
running on the host machine they are not normally accessible to the
container.

This commit adds a script `docker_for_mac_entrypoint.sh`, that
forwards the ports within the container transparently to the Docker
host, so that the API services can be accessed as if they were
running within the container.

By default the entrypoint is not used; it is copied into PATH in the
container but must be enabled when `docker run` is called. An
explanation of how to do this has been added to the readme file.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,7 @@
 *
 
 !dmscripts/
+!docker_for_mac_entrypoint.sh
 !email_templates/
 !requirements.txt
 !schemas/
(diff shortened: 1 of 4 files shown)


galund:
I don't know the background here - why can't we use `-p` for port mapping?

lfdebrux:
Would help if I attached the Trello ticket...

lfdebrux:
Regarding the `-p`, it seems to only work one way, so we can access services running inside the container, but not services running outside the container

PR #231, opened 2018-06-01
Use new key for supplier name in `upload_counterpart_agreements`

Wynndow:
For [this trello ticket.](https://trello.com/c/gDAB5w9c)

The `nameOfOrganisation` key is no longer used in the declaration.
Instead, the registered_name stored in the supplier table is injected
into the declaration and uses `supplierRegisteredName`.


samuelhwilliams:
Looks fine to me, although I see some other scripts in this repo (DOS exports) that seem to use `nameOfOrganisation` as well. Do these not need updating?

PR #232, opened 2018-06-07
Redirect Notify emails in scripts

samuelhwilliams:
## Summary
We send a lot of fake emails with scripts that run against preview.
Doing this regularly and repeatedly will degrade Notify's reputation
with its email provider, so we need to stop sending emails to addresses
that we know are bad. We should make a concerted effort to stop doing
this wherever possible.

This is an alternative to centralising the email redirects in utils as
put forward in
https://github.com/alphagov/digitalmarketplace-utils/pull/405.

 ## Ticket
https://trello.com/c/YIJ25eiP/479

### DIFF
--- a/dmscripts/helpers/email_helpers.py
+++ b/dmscripts/helpers/email_helpers.py
@@ -1,4 +1,16 @@
+from functools import partial
+
 from mandrill import Mandrill
+
+from dmutils.email import DMNotifyClient
+
+
+def scripts_notify_client(notify_api_key, logger):
+    return partial(DMNotifyClient, govuk_notify_api_key=notify_api_key, logger=logger, redirect_domains_to_address={
+        "example.com": "success@simulator.amazonses.com",
+        "example.gov.uk": "success@simulator.amazonses.com",
+        "user.marketplace.team": "success@simulator.amazonses.com",
+    })
 
 
 def get_sent_emails(mandrill_api_key, tags, date_from=None):
--- a/tests/test_notify_buyers_to_award_closed_briefs.py
+++ b/tests/test_notify_buyers_to_award_closed_briefs.py
@@ -135,7 +135,7 @@ class TestSendEmailToBriefUserViaNotify:
         notify_client.assert_not_called()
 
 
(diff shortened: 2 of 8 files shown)


risicle:
Hmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm...

Not possible to at least centralize these for this *repo*?

risicle:
Really I don't know which approach is better - I'm not going to block either, just noting that neither are perfect.

samuelhwilliams:
Of course. I was being a little purposefully obtuse 😅

dmscripts/helpers/email_helpers.py
```diff
@@ -1,5 +1,15 @@
 from mandrill import Mandrill
 
+from dmutils.email import DMNotifyClient
+
+
+def scripts_notify_client(notify_api_key, logger):
```
risicle: This is becoming a bit bikesheddy, but why not just use `partial`?

samuelhwilliams: "a bit"

samuelhwilliams: Partial'd.

risicle:
Cool let me run the tests

tests/test_notify_buyers_to_award_closed_briefs.py
```diff
@@ -135,7 +135,7 @@ def test_send_email_to_brief_user_via_notify_logs_instead_of_sending_for_dry_run
         notify_client.assert_not_called()
 
 
-@mock.patch('dmscripts.notify_buyers_to_award_closed_briefs.DMNotifyClient', autospec=True)
+@mock.patch('dmscripts.notify_buyers_to_award_closed_briefs.scripts_notify_client')
```
katstevens: Is this really the only test where we patch the `DMNotifyClient`? 🤔 I guess all the other ones are top-level scripts, huh.

samuelhwilliams: Indeed

PR #235, opened 2018-06-12
Generate framework agreement pages

samuelhwilliams:
## Summary
We have a separate `digitalmarketplace-agreements` repository that holds
templates used to generate the signature page for framework agreements
that successful suppliers for G-Cloud and Digital Outcomes and
Specialists need to sign before providing services on the framework.
This repository used to have a separate template for each framework
iteration, but now we have a unified template for the signature page
that all frameworks (on the Digital Marketplace, at least) can use.
This repository is pointed at by a script in this
(digitalmarketplace-scripts) repository, so it introduces a layer of
separation between two related things. Given that we are only
maintaining a single template now, effectively, it feels better to bring
the template into the scripts repository directly to minimise the number
of repos we need to juggle and to bring the two related things closer
together so that it's easier to find and maintain context.

This commit effectively deprecates the digitalmarketplace-agrements
repository. It creates a new top-level folder here, called 'templates',
which contains a 'framework_signature_page' to hold the template. A
separate top-level folder, 'email_templates', has been consolidated
under the new 'templates' top-level directory.

The framework agreement template has been updated slightly to update
variable names based on how we now store company details in the supplier
declaration, and how we have removed some duplicate keys from the
framework.frameworkAgreementDetails data blob.


dmscripts/generate_framework_agreement_signature_pages.py
```diff
@@ -21,21 +21,21 @@ def save_page(html, supplier_id, output_dir, descriptive_filename_part):
         htmlfile.write(html)
 
 
-def render_html_for_successful_suppliers(rows, framework_kwargs, template_dir, output_dir):
+def render_html_for_successful_suppliers(rows, framework, template_dir, output_dir, data_api_client):
```
Wynndow: Passing the api client in but never using it here.

scripts/generate-framework-agreement-signature-pages.py
```diff
@@ -61,19 +63,17 @@
...
-    framework_kwargs['frameworkName'] = framework['name']
 
     supplier_id_file = args['<supplier_id_file>']
     supplier_ids = get_supplier_ids_from_file(supplier_id_file)
     html_dir = tempfile.mkdtemp()
 
     records = find_suppliers_with_details_and_draft_service_counts(client, framework_slug, supplier_ids)
     headers, rows = get_csv_rows(records, framework_slug, framework_lot_slugs, count_statuses=("submitted",))
-    render_html_for_successful_suppliers(rows, framework_kwargs, args['<template_folder>'], html_dir)
+    render_html_for_successful_suppliers(rows, framework, 'templates/framework_signature_page', html_dir, client)
```
Wynndow: Feels a bit weird having the template dir hardcoded here. Had a _quick_ look and couldn't see this method being used anywhere other than `dmscripts/generate_framework_agreement_signature_pages.py`. Could just hardcode it there?

Wynndow: Also, the client being passed in here is never used.

samuelhwilliams: This script is the entrypoint for the real python functions that do the work - so it's basically configuration and setup. This feels like the higher level place to hardcode a directory path, if anywhere, rather than deep inside the script.

Yeah the client is a good spot. Was originally going to add an extra call to get the supplier_framework inside the script but it wasn't needed. Have removed.

PR #241, opened 2018-06-15
Check for correct keys in declaration

samuelhwilliams:
## Summary
We need to check the declaration for the presence of two keys in order
to support old and new framework iterations: supplierRegisteredName (G10
and onwards) and nameOfOrganisation (DOS2 and older).

Also updates the upload-counterpart-signatures.py script to derive the
api token from the environment.


lfdebrux:
Can we break this into two commits?

Otherwise looks good to me.

samuelhwilliams:
Done 👍

PR #243, opened 2018-06-18
Convert country codes to names

samuelhwilliams:
## Summary
We store supplier country details in the database as country codes that
relate to a Country or Territory Register. These are not intended for
human consumption, so we should convert these back to the proper
representation of the country (e.g. `country:GB` -> `United Kingdom`),
otherwise we need to do this manually before sharing the data with our
partners (e.g. CCS).

 ## Ticket
https://trello.com/c/bx6MXS0V/173

### DIFF
--- a/dmscripts/helpers/supplier_data_helpers.py
+++ b/dmscripts/helpers/supplier_data_helpers.py
@@ -2,7 +2,12 @@
 """Helper classes for fetching supplier data given a client."""
 from collections import OrderedDict
 from datetime import date
+from functools import lru_cache
 from itertools import groupby
+import json
+
+import backoff
+import requests
 
 from dmutils.formats import DISPLAY_DATE_FORMAT
 
@@ -155,3 +160,16 @@ def get_supplier_ids_from_file(supplier_id_file):
         return None
     with open(supplier_id_file, 'r') as f:
         return list(map(int, [_f for _f in [l.strip() for l in f.readlines()] if _f]))
+
+
+@lru_cache()
+@backoff.on_exception(backoff.expo, requests.exceptions.RequestException, max_tries=3)
+def country_code_to_name(country_code):
+    register, code = country_code.split(':')
+
+    register_response = requests.get(f'https://{register}.register.gov.uk/records/{code}.json')
+    if register_response.status_code == 200:
+        record = json.loads(register_response.text)
+        return record[code]['item'][0]['name']
+
+    raise requests.exceptions.RequestException(register_response)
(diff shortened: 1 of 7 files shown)


tests/conftest.py
```diff
@@ -32,3 +33,21 @@ def mock_data_client():
     with open(os.path.join(FIXTURES_DIR, 'test_supplier_frameworks_response.json')) as supplier_frameworks_response:
         mock_data_client.find_framework_suppliers.return_value = json.loads(supplier_frameworks_response.read())
     return mock_data_client
+
+
+class MockResponse:
```
katstevens: I like this simple class, but is there a reason we're not using `requests-mock` (like we do in the API client?)

samuelhwilliams: None! I just didn't know we used that. I'll change over to it.

scripts/export-framework-applicant-details.py
```diff
@@ -33,11 +34,10 @@
     arguments = docopt(__doc__)
 
     STAGE = arguments['<stage>']
-    API_TOKEN = arguments['<api_token>']
     FRAMEWORK = arguments['<framework_slug>']
     OUTPUT_DIR = arguments['<output_dir>']
 
-    client = DataAPIClient(get_api_endpoint_from_stage(STAGE), API_TOKEN)
+    client = DataAPIClient(get_api_endpoint_from_stage(STAGE), get_auth_token('api', STAGE))
```
lfdebrux: Can we split removing the api_token from the script into a separate commit? Sorry.

samuelhwilliams: Going to remove that from this PR altogether https://github.com/alphagov/digitalmarketplace-scripts/pull/244)

lfdebrux: That makes sense

dmscripts/helpers/supplier_data_helpers.py
```diff
@@ -155,3 +158,15 @@ def get_supplier_ids_from_file(supplier_id_file):
...
     with open(supplier_id_file, 'r') as f:
         return list(map(int, [_f for _f in [l.strip() for l in f.readlines()] if _f]))
+
+
+@lru_cache()
+def country_code_to_name(country_code):
+    register, code = country_code.split(':')
+
+    register_response = requests.get(f'https://{register}.register.gov.uk/records/{code}.json')
+    if register_response.status_code == 200:
```
Wynndow: Would we not want to raise an exception here if the status code isn't 200? Could sliently end up with a mix of country codes and names in the output from the script?

Could maybe with adding backoff to retry in case of none 200s?

samuelhwilliams: Yep, done.

PR #244, opened 2018-06-18
[WIP] `get_auth_token`

samuelhwilliams:
## Summary
Use `get_auth_token` to retrieve API token from environment rather than
pass in on command line.

 ## Warning
Will need to update all Jenkins jobs using these scripts as soon as this gets merged.

PR #245, opened 2018-06-19
208 delete users code that takes the user and strips all the personal data out

benvand:
** Requires alphagov/digitalmarketplace-apiclient#142 **
** Requires https://github.com/alphagov/digitalmarketplace-api/pull/802 **

* Brings in new `apiclient` version
* Adds `scripts/data-retention-remove-user-data.py`
  * If the users last login was greater than 3 * 365 days ago then remove the personal data using the `apiclient` methods that use the new `api` views
  * verbosity and dryrun options
* Adds `scripts/data-retention-remove-contact-information-data.py`
  * If all users of a supplier have had their personal data removed then remove the personal data from that supplier

### DIFF
--- a/scripts/data-retention-remove-user-data.py
+++ b/scripts/data-retention-remove-user-data.py
@@ -1,0 +1,69 @@
+#!/usr/bin/env python
+"""
+Our data retention policy is that users have accounts deactivated and personal data stripped after 3 years without
+a login.
+
+This script is very simple and has not been upgraded to accept any arguments to prevent the possibility of accidental
+deletion. If you are in doubt use the dry run option.
+
+Usage: data-retention-remove-user-data.py <stage> [--dry-run] [--verbose]
+
+Options:
+    --stage=<stage>                                       Stage to target
+
+    --dry-run                                             List account that would have data stripped
+    --verbose
+    -h, --help                                            Show this screen
+
+Examples:
+    ./scripts/data-retention-remove-user-data.py preview
+    ./scripts/data-retention-remove-user-data.py preview --dry-run --verbose
+
+"""
+import logging
+import sys
+from datetime import datetime, timedelta
+from docopt import docopt
+
+from dmapiclient import DataAPIClient
+
+sys.path.insert(0, '.')
+
+from dmscripts.helpers.auth_helpers import get_auth_token
+from dmscripts.helpers.env_helpers import get_api_endpoint_from_stage
+from dmscripts.helpers import logging_helpers
+
+
+if __name__ == "__main__":
+    arguments = docopt(__doc__)
+
+    # Get script arguments
+    stage = arguments['<stage>']
+
+    dry_run = arguments['--dry-run']
+    verbose = arguments['--verbose']
+
+    # Set defaults, instantiate clients
+    prefix = '[DRY RUN]: ' if dry_run else ''
+    logger = logging_helpers.configure_logger(
+        {"dmapiclient": logging.INFO} if verbose else {"dmapiclient": logging.WARN}
+    )
+    data_api_client = DataAPIClient(
+        base_url=get_api_endpoint_from_stage(stage),
+        auth_token=get_auth_token('api', stage)
+    )
+    cutoff_date = datetime.now() - timedelta(days=365 * 3)
+
+    all_users = list(data_api_client.find_users_iter(personal_data_removed=False))
+
+    for user in all_users:
+        last_logged_in_at = datetime.strptime(user['loggedInAt'], '%Y-%m-%dT%H:%M:%S.%fZ')
+        if last_logged_in_at < cutoff_date and not user['personalDataRemoved']:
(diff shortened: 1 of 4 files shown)


scripts/data-retention-remove-user-data.py
```diff
@@ -0,0 +1,64 @@
...
+    cutoff_date = datetime.now() - timedelta(days=365 * 3)
+    data_api_client = DataAPIClient(base_url=get_api_endpoint_from_stage(stage), auth_token=api_token)
+
+    all_users = list(data_api_client.find_users_iter())
+
+    for user in all_users:
+        last_logged_in_at = datetime.strptime(user['loggedInAt'], '%Y-%m-%dT%H:%M:%S.%fZ')
+        if last_logged_in_at < cutoff_date:
+            logger.info(
+                f"{prefix}Removing personal data for user: {user['id']} with email {user['emailAddress']}"
```
katstevens: Don't we normally hash emails before logging them? Let's log this user that we want to forget 😹 Or is this so we have a fighting chance of reinstating the user if this script goes wrong?

scripts/data-retention-remove-user-data.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # Set defaults, instantiate clients
+    prefix = '[DRY RUN]: ' if dry_run else ''
+    logger = logging_helpers.configure_logger(
+        {"dmapiclient": logging.INFO} if verbose else {"dmapiclient": logging.WARN}
+    )
+    cutoff_date = datetime.now() - timedelta(days=365 * 3)
+    data_api_client = DataAPIClient(base_url=get_api_endpoint_from_stage(stage), auth_token=api_token)
+
+    all_users = list(data_api_client.find_users_iter())
```
katstevens: Is there any benefit filtering out users who have already had personal data removed? We wouldn't need to handle the inevitable error...

scripts/data-retention-remove-user-data.py
```diff
@@ -0,0 +1,64 @@
...
+
+sys.path.insert(0, '.')
+
+
+if __name__ == "__main__":
+    arguments = docopt(__doc__)
+
+    # Get script arguments
+    stage = arguments['<stage>']
+    api_token = arguments['<api_token>']
```
katstevens: @samuelhwilliams is changing how our scripts use API tokens - see https://github.com/alphagov/digitalmarketplace-scripts/pull/244. Let's use the helper instead.

scripts/data-retention-remove-user-data.py
benvand: Done 👍 Removed this, was over cautious

scripts/data-retention-remove-user-data.py
benvand: Yup, added to both scripts. We only want to find users who _have not_ had their data removed in the users script.

We only want to find users who _have_ in the contact details one

benvand: Done 👍

scripts/data-retention-remove-user-data.py
benvand: Niiiiiiicccceee!! Done 👍

scripts/data-retention-remove-user-data.py
```diff
@@ -0,0 +1,69 @@
...
+    data_api_client = DataAPIClient(
+        base_url=get_api_endpoint_from_stage(stage),
+        auth_token=get_auth_token('api', stage)
+    )
+    cutoff_date = datetime.now() - timedelta(days=365 * 3)
+
+    all_users = list(data_api_client.find_users_iter(personal_data_removed=False))
+
+    for user in all_users:
+        last_logged_in_at = datetime.strptime(user['loggedInAt'], '%Y-%m-%dT%H:%M:%S.%fZ')
```
katstevens: Looking at the `user['loggedInAt']` field - this could potentially also be the 'createdAt' date (if the user has never logged in). I don't think it will make a difference in practice - they still haven't logged in within 3 years! - but it might be good to explicitly mention this somewhere.

benvand: I'm not that worried about this one because the account creation still logs the user in. We want to delete them after 3 years regardless of the number of times they've logged in (ie even if they've only logged in once)

scripts/data-retention-remove-user-data.py
```diff
@@ -0,0 +1,69 @@
...
+Options:
+    --stage=<stage>                                       Stage to target
+
+    --dry-run                                             List account that would have data stripped
+    --verbose
+    -h, --help                                            Show this screen
+
+Examples:
+    ./scripts/data-retention-remove-user-data.py preview myToken
+    ./scripts/data-retention-remove-user-data.py preview myToken --dry-run --verbose
```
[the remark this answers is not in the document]
benvand: Done 👍

PR #248, opened 2018-06-20
Add G10 service keys to bad words script

Wynndow:
The list is now an amalgamation of all keys checked for G7 through G10.

I did want to remove this list completely and just scan any value in the
service data that was of string type. Unfortunately we have list
questions, and checkbox questions, which both appears as lists of
strings in the data.

The strings from checkbox questions are defined by us, and they may
include words defined in our list of bad words (words given to us by
CCS, not naughty words). This means we could get some false positives.

### DIFF
--- a/scripts/scan-g-cloud-services-for-bad-words.py
+++ b/scripts/scan-g-cloud-services-for-bad-words.py
@@ -14,11 +14,7 @@ from collections import Counter
 
 sys.path.insert(0, '.')
 import os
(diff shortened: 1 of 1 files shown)


Wynndow:
Yeah, I guess that it's not that necessary. It could probably just be lists for each framework family. Alternatively, could do away with the list of keys all together and just scan any value in the service that is of string type?

PR #249, opened 2018-06-21
Acknowledge draft service updates

samuelhwilliams:
## Summary
When moving G-Cloud 10 services from drafts to 'published', we generate
a lot of `update_service` audit events. We have an admin page that
allows CCS to review and acknowledge `update_service` audit events, but
this is only intended for services which are currently live (i.e. the
framework is open). Service changes from the script which promotes the
drafts, `publish-g-cloud-draft-services.py` (and
`publish-dos-draft-services.py`) should not need to be acknowledged by
CCS.

This script will go through and acknowledge all audit events created by
the publishing scripts.

### DIFF
--- a/scripts/oneoff/acknowledge_publish_draft_service_updates.py
+++ b/scripts/oneoff/acknowledge_publish_draft_service_updates.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python
+
+import argparse
+import backoff
+import multiprocessing as mp
+from multiprocessing.pool import ThreadPool
+import sys
+
+from dmapiclient import HTTPError
+from dmapiclient.audit import AuditTypes
+from dmapiclient.data import DataAPIClient
+
+sys.path.insert(0, '.')  # noqa
+
+from dmscripts.helpers.auth_helpers import get_api_url, get_auth_token
+
+
+ACKNOWLEDGE_USER = 'scripts/oneoff/acknowledge_publish_draft_service_updates.py'
+
+
+@backoff.on_exception(backoff.expo, HTTPError, max_tries=5)
+def acknowledge_update(audit_event):
+    response = data.acknowledge_audit_event(audit_event['id'], user=ACKNOWLEDGE_USER)['auditEvents']
+    print(f'Acknowledged {audit_event["id"]} on service id #{audit_event["data"]["serviceId"]} by '
+          f'supplier `{audit_event["data"]["supplierName"]}`: {response["acknowledged"]}')
+
+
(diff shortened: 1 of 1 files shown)


scripts/oneoff/acknowledge_publish_draft_service_updates.py
```diff
@@ -0,0 +1,37 @@
...
+import sys
+
+from dmapiclient.audit import AuditTypes
+from dmapiclient.data import DataAPIClient
+
+sys.path.insert(0, '.')  # noqa
+
+from dmscripts.helpers.auth_helpers import get_api_url, get_auth_token
+
+def acknowledge_update(audit_event):
```
Wynndow: worth adding backoff? There's gonna be a whole load of audit events to work through.

scripts/oneoff/acknowledge_publish_draft_service_updates.py
```diff
@@ -0,0 +1,37 @@
...
+    response = data.acknowledge_audit_event(audit_event['id'], user='(dm developers) scripts/publish-g-cloud-draft-services.py')['auditEvents']
```
Wynndow: maybe user should be `scripts/acknowledge_publish_draft_service_updates.py`?

scripts/oneoff/acknowledge_publish_draft_service_updates.py
```diff
@@ -0,0 +1,37 @@
...
+    print(f'Acknowledged {audit_event["id"]} on service id #{audit_event["data"]["serviceId"]} by supplier `{audit_event["data"]["supplierName"]}`: {response["acknowledged"]}')
```
Wynndow: Maybe add a check for a 200?

Wynndow:
Bunch of flake8 stuff for line length and whitespace

scripts/oneoff/acknowledge_publish_draft_service_updates.py
samuelhwilliams: API client doesn't return a status code, it raises errors on non-200.

scripts/oneoff/acknowledge_publish_draft_service_updates.py
samuelhwilliams: Yeah sure

scripts/oneoff/acknowledge_publish_draft_service_updates.py
Wynndow: cool

PR #254, opened 2018-06-22
Give published documents 'public-read' acl

Wynndow:
Since we last published G-Cloud documents we've changed our S3 module
and documents now get a default acl of 'private'. We need to override
this.

Also adds some info to the docstring about object owners.


Wynndow:
No, I don't think so. But the next time this script is used, the person running it will have to think about how to run this with the correct credentials. It possibly means adding aws-auth to the Jenkins box, so the Jenkins can assume a different role to run the script.

Wynndow:
@benvand Nudge nudge.

Wynndow:
You might have to merge this Ben, I don't have permissions anymore!

PR #255, opened 2018-06-22
Add SQL script to fix accidentally published services

Wynndow:
Some services where published accidentally for G10 that belonged to
suppliers who didn't make it on to the framework.

This reverts the mistake and adds audit events manually as a record.

### DIFF
--- a/scripts/oneoff/fix_accidentially_published_g10_services.sql
+++ b/scripts/oneoff/fix_accidentially_published_g10_services.sql
@@ -1,0 +1,85 @@
+/*
+This script is to fix the results of accidentally publishing some draft services for some suppliers who were not
+awarded `on_framework` for G10. The issue was with the publishing script, and the fix to prevent this happening again is
+here: https://github.com/alphagov/digitalmarketplace-scripts/pull/256
+
+To apply this script run the below command
+Targe the correct environment with `cf target -s <desired environment>`
+`cf conduit digitalmarketplace_api_db -- psql < ./scripts/oneoff/fix_accidentially_published_g10_services.sql`
+*/
+
+START transaction;
+SELECT
+   services.service_id,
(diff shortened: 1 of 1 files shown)


scripts/oneoff/fix_accidentially_published_g10_services.sql
```diff
@@ -0,0 +1,84 @@
...
+   AND sf.framework_id = (SELECT id FROM frameworks WHERE slug = 'g-cloud-10')
+;
+
+UPDATE
+   draft_services
+SET
+   service_id = NULL
+WHERE
+   framework_id = (SELECT id FROM frameworks WHERE slug = 'g-cloud-10')
+   AND supplier_id IN (SELECT supplier_id FROM naughty_services)
```
galund: It'd be more consistent to test `service_id` here rather than `supplier_id`?

scripts/oneoff/fix_accidentially_published_g10_services.sql
```diff
@@ -0,0 +1,84 @@
+/*
+To apply this script run the below command
```
galund: Would perhaps be good to link here to the PR that fixes the broken thing that cause this to be necessary?

Wynndow: Yep, will do when it's up!

scripts/oneoff/fix_accidentially_published_g10_services.sql
Wynndow: I think i've made it even simpler than than! Let me know if you think its ok.

scripts/oneoff/fix_accidentially_published_g10_services.sql
Wynndow: It's done!

PR #260, opened 2018-07-11
send-stats-to-performance-platform: allow configurable pp "service" name

risicle:
Trello https://trello.com/c/GE3VaURm/95-get-performance-platform-dashboard-for-dos3

Decided to make this an optional parameter defaulting to "gcloud" for backwards compatibility - should be mergeable without a synchronized jenkins update.

### DIFF
--- a/dmscripts/send_stats_to_performance_platform.py
+++ b/dmscripts/send_stats_to_performance_platform.py
@@ -10,14 +10,14 @@ from dmscripts.helpers.logging_helpers import logging
 
 HOURLY_TIME_FORMAT = '%Y-%m-%dT%H:00:00+00:00'  # On the hour exactly
 DAILY_TIME_FORMAT = '%Y-%m-%dT00:00:00+00:00'  # Midnight
-PERFORMANCE_PLATFORM_URLS = {
+PERFORMANCE_PLATFORM_URL_TEMPLATES = {
     "day": {
-        "stage": "https://www.performance.service.gov.uk/data/gcloud/applications-by-stage",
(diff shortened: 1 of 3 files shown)


lfdebrux:
So we already have a framework slug argument, would it be better for the performance platform service name to be derived from this?

Otherwise you have the situation where if you have the framework slug as dos3 the stats will go to the gcloud service by default, which seems like surprising behaviour to me.

risicle:
The issue is that we're not *necessarily* the ones that get to decide what the pp service name ends up being. It's generally something that we get assigned. Deriving it from the fw slug would also bind us to that derivation scheme and if we needed to e.g. temporarily point it at a different pp service for testing something we wouldn't be able to. Having this as an opaque input string seems the most unassuming thing here.

risicle:
But having a map would mean we would necessarily have to make a PR to this repo every time we had a new framework - it would be putting what is basically configuration in this repo.

To clarify - the intention is to remove the default after jenkins is updated and make the arg mandatory.

...though actually come to think of it, nothing is actually running this script at the moment, so merging a breaking change requiring the new argument would actually not break anything. Might just do that,

lfdebrux:
As I'm about to merge a change to almost all of the scripts (#244) which requires changing Jenkins and rebuilding all the scripts (digitalmarketplace-jenkins#100) now is a good time to change the arguments.

risicle:
Doing

risicle:
Done

lfdebrux:
Cool cool. Do you have a change for Jenkins as well?

PR #261, opened 2018-07-13
Remove code that relies on sops/credentials repo

lfdebrux:
When running scripts in Docker, the credentials repo and SOPS are not available.

If the user forgets to expose the api token environment variable (like I did, today 😱), they will get confusing errors about SOPS not being available, when really the error should be that the api token is not available in the environment.

As we are going to be running scripts in Docker from now on, this sops code is dead, and so I have replaced it with a helpful error message.

### DIFF
--- a/dmscripts/helpers/auth_helpers.py
+++ b/dmscripts/helpers/auth_helpers.py
@@ -1,6 +1,4 @@
 import os
-import subprocess
-import yaml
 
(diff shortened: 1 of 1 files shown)


risicle:
Got your hand bitten off with that one.

lfdebrux:
Rub salt in the wound why don't you 😭 😭  😭

PR #262, opened 2018-07-17
Check ALL of a suppliers users for personal data removed rather than …

benvand:
…just ones that _do_ have personal data removed

### DIFF
--- a/scripts/data-retention-remove-contact-information-data.py
+++ b/scripts/data-retention-remove-contact-information-data.py
@@ -53,10 +53,7 @@ if __name__ == "__main__":
         auth_token=get_auth_token('api', stage)
     )
 
-    users = sorted(
-        data_api_client.find_users_iter(role='supplier', personal_data_removed=True),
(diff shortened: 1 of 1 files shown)


scripts/data-retention-remove-contact-information-data.py
```diff
@@ -53,10 +53,7 @@
         auth_token=get_auth_token('api', stage)
     )
 
-    users = sorted(
-        data_api_client.find_users_iter(role='supplier', personal_data_removed=True),
-        key=lambda i: i['supplier']['supplierId']
-    )
+    users = sorted(data_api_client.find_users_iter(role='supplier'), key=lambda i: i['supplier']['supplierId'])
```
Wynndow: Why sort them here when they're grouped by supplier id in the next line?

benvand: One of my least favourite Python things. Groupby doesn't group unless they're already in groups 🤔

PR #273, opened 2018-08-30
Add tests for indexing script

katstevens:
Prep for https://trello.com/c/KzDlJAcH/125-prevent-indexing-script-using-wrong-mapping-file

Our main indexing script is untested 🙀. This PR should have no functional changes at all. I've run it locally for briefs and services and it seems fine, but not against Preview yet. Perhaps you, the reviewer, would like to give it a go...

- Moves logic out of top level script
- Adds tests for `do_index` (the main function) and the major class methods
- Some weird patching of the two API clients was needed as *I didn't want to change the code or the docopt stuff at all*. Normally in scripts we initialise the clients at the top level then pass the instance through to the script function, but here we initialise within the script function itself. So instead of just using a mock object as the client instance, we have to patch the class and use `data_api_client.return_value`. Hopefully this is only temporary as this script is going to get refactored to be like the others soon... right?

### DIFF
--- a/dmscripts/index_to_search_service.py
+++ b/dmscripts/index_to_search_service.py
@@ -1,0 +1,116 @@
+from datetime import datetime
+from multiprocessing.pool import ThreadPool
+
+import backoff
+import dmapiclient
+from six.moves import map
+
+from dmscripts.helpers import logging_helpers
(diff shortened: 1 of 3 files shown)


katstevens:
Ran this against preview, all seems fine.

PR #275, opened 2018-09-05
Handle multiple email addresses in Notify suppliers of awarded briefs script

lfdebrux:
In some cases users have multiple email addresses in their respondToEmailAddress field.

This causes errors in the "Notify suppliers of awarded briefs" script (see [this ticket](https://trello.com/c/DHDgBuns)).

This PR adds a limited ability to handle strings where there are multiple email addresses separated by a character such as `;` or `/`.

See alphagov/digitalmarketplace-utils#450 for more details on the helper functions used.

This PR only modifies the one script, as that is the only one I've seen having this error. We can extend it to other scripts with a little effort; unfortunately all the scripts seem to be written in slightly different ways without much reuse.

Additionally, I removed Flask-FeatureFlags from the requirements file as I upgraded to `dmutils 43.x.x`. I also upgraded Flask to the latest version as used in the frontend apps. I'm not clear on exactly why Flask is pinned in this repo, as it isn't used anywhere, but as @katstevens is away I've left it in until I can ask her about it (she added it in).

### DIFF
--- a/dmscripts/notify_suppliers_of_awarded_briefs.py
+++ b/dmscripts/notify_suppliers_of_awarded_briefs.py
@@ -1,7 +1,7 @@
 from datetime import datetime, timedelta, date
 
 from dmutils.email.exceptions import EmailError
-from dmutils.email.helpers import hash_string
+from dmutils.email.helpers import get_email_addresses, hash_string, validate_email_address
 from dmutils.formats import DATE_FORMAT
 from dmutils.env_helpers import get_web_url_from_stage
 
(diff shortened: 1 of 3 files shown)


dmscripts/notify_suppliers_of_awarded_briefs.py
```diff
@@ -55,34 +55,49 @@ def _build_and_send_emails(brief_responses, mail_client, stage, dry_run, templat
...
-                    'email_address': hash_string(email_address),
-                }
-            )
-        except EmailError:
-            # Log individual failures in more detail
+        invalid_email_addresses = [a for a in email_addresses if not validate_email_address(a)]
+        if invalid_email_addresses:
             logger.error(
-                "Email sending failed for BriefResponse {brief_response_id} (Brief ID {brief_id})",
+                "Invalid email address(es) for BriefResponse {brief_response_id} (Brief ID {brief_id})",
```
Wynndow: Do we want to skip the entire brief response if only one email address is invalid? Would it be better to log the invalid email and continue with the valid?

PR #277, opened 2018-09-11
Export suppliers/users to CSV scripts

katstevens:
Trello: https://trello.com/c/FtV7sVMU/138-admin-download-all-suppliers-for-framework-endpoint-locks-up-an-api-process

Fetches supplier/user data from the API, creates a CSV (saving to a local `data` folder by default, as per the `get-model-export.py` script) then uploads that file to S3. 

There are 3 different CSV files that could be generated by this script:

1. Official supplier details for a framework
2. Supplier user contact details for a framework (all active users)
3. Supplier user contact details for a framework (user research opted-in only)

Examples:

```
./scripts/generate-supplier-user-csv.py preview suppliers g-cloud-10
./scripts/generate-supplier-user-csv.py preview users g-cloud-10 --dry-run
./scripts/generate-supplier-user-csv.py preview users g-cloud-10 --user-research-opted-in
```

The code for the API calls and report generation is adapted from the Admin FE views (the CSV generation is simplified, as we don't have to return it as a streamed response): 
https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/main/views/users.py#L74
https://github.com/alphagov/digitalmarketplace-admin-frontend/blob/master/app/main/views/users.py#L119

The code for the script initialisation and the S3 upload is adapted from `scripts/upload-counterpart-agreements.py`.

Ready for review, however the new buckets still need to be created.

### DIFF
--- a/dmscripts/generate_supplier_user_csv.py
+++ b/dmscripts/generate_supplier_user_csv.py
@@ -1,0 +1,154 @@
+import csv
+from dmutils.s3 import S3ResponseError
+
+from dmscripts.models.writecsv import csv_path
+
+
+def generate_supplier_csv(framework_slug, data_api_client, logger):
+    framework = data_api_client.get_framework(framework_slug).get("frameworks")
+
+    supplier_rows = data_api_client.export_suppliers(framework_slug).get('suppliers', [])
+    if not supplier_rows:
+        logger.info('No supplier data found for framework {}'.format(framework_slug))
+        return [], [], None
+
+    supplier_and_framework_headers = [
+        "supplier_id",
+        "supplier_name",
+        "supplier_organisation_size",
+        "duns_number",
+        "companies_house_number",
+        "registered_name",
+        "declaration_status",
+        "application_status",
+        "application_result",
+        "framework_agreement",
+        "variations_agreed",
+    ]
+    service_count_headers = [lot['slug'] for lot in framework['lots']]
+    contact_info_headers = [
+        'contact_name',
+        'contact_email',
+        'contact_phone_number',
+        'address_first_line',
+        'address_city',
+        'address_postcode',
+        'address_country',
+    ]
+
+    filename = "official-details-for-suppliers-{}".format(framework_slug)
+
+    supplier_headers = supplier_and_framework_headers + ["total_number_of_services"] + \
+        ["service-count-{}".format(h) for h in service_count_headers] + contact_info_headers
+
+    formatted_rows = []
+    for row in supplier_rows:
+        # Include an extra column with the total number of services across all lots
+        service_counts = [row['published_services_count'][heading] for heading in service_count_headers]
+        total_number_of_services = sum(service_counts)
+
+        formatted_rows.append(
+            [row[heading] for heading in supplier_and_framework_headers] +
+            [total_number_of_services] +
+            service_counts +
+            [row['contact_information'][heading] for heading in contact_info_headers]
+        )
+
+    return supplier_headers, formatted_rows, filename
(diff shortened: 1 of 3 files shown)


lfdebrux:
👍 for tests

dmscripts/generate_supplier_user_csv.py
```diff
@@ -0,0 +1,152 @@
...
+        'contact_name',
+        'contact_email',
+        'contact_phone_number',
+        'address_first_line',
+        'address_city',
+        'address_postcode',
+        'address_country',
+    ]
+
+    filename = "official-details-for-suppliers-{}".format(framework_slug)
```
lfdebrux: Is it worth putting a datestamp in the filename?

dmscripts/generate_supplier_user_csv.py
```diff
@@ -0,0 +1,152 @@
...
+        "variations_agreed",
+        "published_service_count"
+    ]
+    user_rows = data_api_client.export_users(framework_slug).get('users', [])
+    if not user_rows:
+        logger.info('No user data found for framework {}'.format(framework_slug))
+        return [], [], None
+
+    if user_research_opted_in:
+        user_rows = [row for row in user_rows if row['user_research_opted_in']]
```
lfdebrux: We can turn this into a generator expression by wrapping in `(` instead of `[` and maybe gain a little performance

dmscripts/generate_supplier_user_csv.py
```diff
@@ -0,0 +1,152 @@
...
+        filename = "user-research-suppliers-on-{}".format(framework_slug)
+    else:
+
+        filename = "all-email-accounts-for-suppliers-{}".format(framework_slug)
+
+    formatted_rows = [
+        [row[heading] for heading in user_headers] for row in user_rows
```
lfdebrux: I think the square brackets here are superfluous

dmscripts/generate_supplier_user_csv.py
katstevens: I'm a bit torn on that. The current report doesn't have a datestamp, but it's generated on demand so the user should 'know' its date. Adding it might be useful to flag up when the data has changed. On the other hand, it will make finding the file in the bucket and generating the URL more annoying...

scripts/generate-supplier-user-csv.py
```diff
@@ -0,0 +1,95 @@
...
+              variable to reference the right credentials before running the script.
+              Alternatively, if this script is being run from Jenkins, do not provide any credentials and boto will use
+              the Jenkins IAM role. It should have the required permissions for the bucket.
+
+This will:
+ * call the export_<suppliers|users>_for_framework endpoint from the API
+ * create a CSV file from the results and save to saved to `<output-dir>/<filename>.csv` (see get-model-data.py)
+ * upload the file to the S3 admin reports bucket for <stage> with the file path:
+    <framework_slug>/suppliers/<YYYY-MM-DD>-suppliers-export.csv OR
+    <framework_slug>/suppliers/<YYYY-MM-DD>-supplier-users-export.csv
```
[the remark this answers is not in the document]
katstevens: Ooops yes this was copied over from another script, will update.

dmscripts/generate_supplier_user_csv.py
katstevens: We can iterate through it just once anyway with a bit of rejigging. I'll push a commit.

katstevens:
Going to merge this so I can try it out on Jenkins (runs fine locally).

PR #279, opened 2018-09-13
Allow mark_definite_framework_results to take supplier id list

Wynndow:
If you're reassessing, you may only want to assess certain suppliers.

### DIFF
--- a/dmscripts/mark_definite_framework_results.py
+++ b/dmscripts/mark_definite_framework_results.py
@@ -82,8 +82,11 @@ def mark_definite_framework_results(
     reassess_failed=False,
     reassess_failed_draft_services=False,
     logger=logging.getLogger("script"),
+    supplier_ids=None,
 ):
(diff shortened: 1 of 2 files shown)


dmscripts/mark_definite_framework_results.py
```diff
@@ -82,8 +82,11 @@ def mark_definite_framework_results(
     reassess_failed=False,
     reassess_failed_draft_services=False,
     logger=logging.getLogger("script"),
+    supplier_ids=None,
 ):
-    interested_supplier_ids = client.get_interested_suppliers(framework_slug).get('interestedSuppliers', ())
+    interested_supplier_ids = supplier_ids or client.get_interested_suppliers(
+        framework_slug
+    ).get('interestedSuppliers', ())
```
risicle: So `supplier_ids` don't *also* get checked for being "interested"?

Wynndow: mmm....no. I guess it's left up to the users discretion.

PR #282, opened 2018-09-25
Include reports folder in S3 upload path

katstevens:
Trello: https://trello.com/c/FtV7sVMU/138-admin-download-all-suppliers-for-framework-endpoint-locks-up-an-api-process

In order to serve the S3 file in the admin (via the `DM_ASSETS_URL`), the router needs to proxy the url to the correct bucket. To do that, it uses a regex to match the path supplied (see https://github.com/alphagov/digitalmarketplace-router/pull/30/files). The other asset buckets include their type in the path structure to assist with this (e.g. `/g-cloud-10/communications/filename.csv`), and the reports bucket should do the same.

TL;DR - this PR sticks a `.../reports/...` in the CSV upload path.

Tested against preview already, works as expected.


katstevens:
Thanks for the speedy review 😸

PR #283, opened 2018-09-25
Remove hard coded framework from send_dos_opportunities

Wynndow:
For [this trello ticket.](https://trello.com/c/FMcdroGN)

Content has been checked with Tracy and Constance.

The script to send suppliers information about newly released dos
opportunities had the framework name hardcoded into the subject line for
the mailchimp campaign. This was an issue as there is a chance that
we'll have new briefs from two different iterations when we transition
from DOS2 to DOS3.

This grabs the iteration from the live briefs found and formats the
subject line. It also means we don't have to worry about updating the
script after DOS3 goes live.

### DIFF
--- a/tests/test_send_dos_opportunities_email.py
+++ b/tests/test_send_dos_opportunities_email.py
@@ -1,7 +1,8 @@
 import mock
-from freezegun import freeze_time
-
 from datetime import date
+
+import pytest
+from freezegun import freeze_time
 from lxml import html
 
 from dmscripts.send_dos_opportunities_email import (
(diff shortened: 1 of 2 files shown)


tests/test_send_dos_opportunities_email.py
```diff
@@ -155,7 +166,11 @@ def test_get_campaign_data():
 def test_main_creates_campaign_sets_content_and_sends_campaign(
     get_campaign_data, get_live_briefs_between_two_dates, get_html_content
 ):
-    get_live_briefs_between_two_dates.return_value = [{"brief": "yaytest"}]
+    live_briefs = [
+        {"brief": "yaytest", "frameworkName": "Digital Outcomes and Specialists 2"},
+        {"brief": "bootest", "frameworkName": "Digital Outcomes and Specialists 1"},
```
katstevens: Would `Digital Outcomes and Specialists 50` give the right suffix here? I'm not sure we really need to support it - DOS 10, can you imagine 🙀 - but having the `50` test scenario above seems a bit misleading so I'd ditch that.

Wynndow: I think I was just trying to show that it's not limited to just "1" and "2". Happy to remove though.

Wynndow: It's gone!

PR #284, opened 2018-09-27
Update scripts that manage and send DOS3 opportunity emails

Wynndow:
For [this trello ticket](https://trello.com/c/y3Rv3kGC).

This will require updates to the Jenkins jobs which run these scripts. Coming next.

### Add DOS3 list IDs to upload_dos_opportunities_email_list  …
We're going to need a new list for DOS3 suppliers. Being able to select
framework specific list ids will make the transition from dos2 to dos3
much easier.

The list IDs for DOS3 have already been created in Mailchimp.

### Add DOS3 mailchimp list IDs to email sending script  …
And pass framework_slug to main.

We need to send separate emails to separate lists based on which DOS
opportunities are available.

### DIFF
--- a/dmscripts/send_dos_opportunities_email.py
+++ b/dmscripts/send_dos_opportunities_email.py
@@ -9,24 +9,26 @@ from dmutils.formats import DATETIME_FORMAT, DISPLAY_DATE_FORMAT
 logger = logging_helpers.configure_logger({'dmapiclient': logging.INFO})
 
 
-def get_live_briefs_between_two_dates(data_api_client, lot_slug, start_date, end_date):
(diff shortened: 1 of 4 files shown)


katstevens:
I'm curious why we need a new mailing list for DOS 3. Would the supplier notice/care if there was a mixture? I'd probably prefer one email to two...

Wynndow:
My thinking is that there's only going to be the potential for one day for suppliers to receive two emails - and that probably won't actually happen.

Going forward though, we shouldn't be sending emails to all the DOS2 suppliers about DOS3 opportunities, as they can't apply. Sure there will be a large overlap, but for suppliers who are on DOS2 and not DOS3 it seems a bit off.

katstevens:
OK, that makes sense.

PR #286, opened 2018-10-01
Add DOS3 to get-model-data-script

Wynndow:
The get-model-data script takes a nightly data extract from the database and exports it for use in analytics. Now DOS3 is live we want to make sure we're capturing that data too.


katstevens:
Did we add new specialist roles for DOS3? There's a hardcoded list of them in this script...

Wynndow:
Ah, nice catch. I'll add them.

PR #290, opened 2018-10-09
Update dev requirements

lfdebrux:
This PR updates the packages we use for testing scripts


risicle:
Unlocking the versions? Are you sure this is wise? We'll all be running our tests with different versions of e.g. pytest - will bring whole new meaning to "works for me".

lfdebrux:
@risicle I've pinned the versions for the packages where it makes sense; if there are future issues we can always be more restrictive then.

PR #292, opened 2018-10-09
Add LICENCE

katstevens:
Trello: https://trello.com/c/ECCRn0Lp/189-licenses-for-public-repos

### DIFF
--- a/LICENCE
+++ b/LICENCE
@@ -1,0 +1,20 @@
+The MIT License (MIT)
+Copyright (C) 2015 Crown Copyright (Government Digital Service)
+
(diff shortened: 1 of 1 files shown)


LICENCE
```diff
@@ -0,0 +1,20 @@
+The MIT License (MIT)
+Copyright (C) 2015 Crown Copyright (Government Digital Service)
```
lfdebrux: Any reason the date is 2015?

katstevens: First commit was October 2015. We could put a range but we don't in our other licences.

PR #293, opened 2018-10-10
Antivirus scanning script

risicle:

This ports the script `virus-scan-s3-bucket.py` to use the antivirus api for actual scanning of objects. It will allow us to scan & tag all existing files on a bucket and will also be good to use as a "catch up" script which runs periodically and scans any file versions which missed SNS-triggered scanning for some reason.

There are a number of things that should(/could?) be done to this before we say it's done:

 - Add tests, obviously. This may need some preparatory work as the test will presumably use `moto` which isn't yet a dependency of this repo. And because this will be the third place I'd be copypasta-ing basically the same `moto`-based pytest fixtures, it feels like time to stick them together - possibly in `-test-utils`?
- Multithread the loop. Not particularly hard - we have multiple `antivirus-api` instances and multiple webserver and `clamd` processes per instances - the service should be able to handle a number of requests in parallel with no problem and this should reduce the execution time when performing batch scans. Maybe I should run a timing test to give people a better picture how "slow" a full-bucket scan could be. Incidentally, this is why I designed the log messages in a way that would make sense if the request and response for each file got interleaved with another...
- Options like "force re-scan"? This would also require additions to the api endpoint, but I definitely foresee us needing to do this at some point in the future.
 - Actually run this on production to reassure ourselves we're not already hosting any nasties.

Thoughts on what order they'd like these done in?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,11 +6,11 @@ virtualenv:
 	[ -z $$VIRTUAL_ENV ] && [ ! -d venv ] && python3 -m venv venv || true
 
 .PHONY: requirements
-requirements: virtualenv requirements.txt
+requirements: virtualenv test-requirements requirements.txt
 	${VIRTUALENV_ROOT}/bin/pip install -r requirements.txt
 
 .PHONY: requirements-dev
-requirements-dev: virtualenv requirements-dev.txt
+requirements-dev: virtualenv test-requirements requirements-dev.txt
 	${VIRTUALENV_ROOT}/bin/pip install -r requirements-dev.txt
 
 .PHONY: freeze-requirements
--- a/default.nix
+++ b/default.nix
@@ -38,6 +38,9 @@ in (with args; {
     VIRTUAL_ENV_DISABLE_PROMPT = "1";
     SOURCE_DATE_EPOCH = "315532800";
 
+    # pip's builds will never be pure - allow gcc to include impure paths
+    NIX_ENFORCE_PURITY=0;
+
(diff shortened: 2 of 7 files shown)


katstevens:
I think `moto` wouldn't be essential for tests if we really want to avoid the dependency - we're passing in an initialised `s3_client` to the `dmscripts` function so we can just use a normal mock with JSON return values? I did something similar for the bucket in `test_generate_supplier_user_csv.py`. 

Hmm, after a bit of digging, I'm not sure `moto` would be much help for this particular script anyway: https://github.com/spulec/moto/blob/master/moto/ecr/responses.py#L140

My verdict: a simple happy path test for scanning a bucket containing three files (1 pass, 1 fail and 1 'already tagged') would be useful here before we start doing multi threading (or other tweaks). 

Force-scan: I'm not sure about the scenarios where we need this. Can we chat about it?

risicle:
Force-scan: I'm considering e.g. 6 months, a year down the line when we will probably want to re-scan a whole bucket with an updated virus definition.

risicle:
The issue with just mocking out the s3 client with responses you "expect" to get back means that you don't assert that you've got the right idea about the how the api *actually* works. e.g. there are a couple of ins & outs of the boto `paginator` interface that I didn't expect. Using moto in the av-api itself instead of just mocking what I expected caught several bugs.

risicle:
Ok, (basic) tests implemented.

risicle:
Think that will have been because I initially pushed up a slightly mis-rebased version.

PR #295, opened 2018-10-15
virus_scan_s3_bucket: don't try to be so clever in counting

risicle:

@lfdebrux alerted me that integer increment in python is indeed *not* atomic. So instead don't try to be such a smartass in the file counting and instead pass the counts in the map function results and accumulate in the loop...

### DIFF
--- a/dmscripts/virus_scan_s3_bucket.py
+++ b/dmscripts/virus_scan_s3_bucket.py
@@ -14,17 +14,12 @@ def virus_scan_bucket(
     dry_run=True,
     map_callable=map,
 ):
-    candidate_count, pass_count, fail_count, already_tagged_count = 0, 0, 0, 0
-
     def handle_version(version):
-        # integer incrementing is "atomic" in python so it's ok to share these counters across threads. the reason we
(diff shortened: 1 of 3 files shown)


risicle:
Think we might have the first case of "tests work for me, but..."

risicle:
No, I was just looking through the meaning of the word **non**local

lfdebrux:
While I appreciate that you have already changed the counting logic a lot, I think some of the legacy of the original form is now obscuring the intent.

I think it would be easier to read if the `handle_version` function were to return a status string such as `'pass'`, '`already_tagged`', etc, and then the counting could take place purely outside of the `handle_version` function.

Let me know if you disagree.

risicle:
That's also a possibility, was just trying to avoid having a the sort of function that can have unknown or unexpected responses. Also that forces the function to classify a result as one category when the caller might want to count different properties of the response. Ultimately it still bears the mark of the fact that it's really just a broken-out loop body, but I'm ok with that - it's just a little script.

risicle:
Right - I think this method just trades one set of future difficulties or for an equal alternative and isn't any more readable. But I really don't think it makes much difference either way.

PR #297, opened 2018-10-18
virus_scan_s3_bucket: add support for scanning multiple buckets in one call

risicle:

This is because we'll probably want to scan all an environment's buckets in a single jenkins job and writing the loop in python is nicer than writing it in bash.

Tests getting more complex though.

### DIFF
--- a/tests/test_virus_scan_s3_bucket.py
+++ b/tests/test_virus_scan_s3_bucket.py
@@ -2,7 +2,7 @@ from collections import Counter
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
 from datetime import datetime
-from itertools import groupby
+from itertools import chain, groupby
 
 import mock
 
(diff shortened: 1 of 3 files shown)


tests/test_virus_scan_s3_bucket.py
```diff
@@ -187,18 +223,60 @@ def test_unfiltered(self, versions_page_size, dry_run, concurrency):
...
+                dry_run=dry_run,
+                map_callable=map_callable,
+            )
+
+        assert s3_client.mock_calls == [
+            mock.call.get_paginator("list_object_versions"),
+            mock.call.get_paginator().paginate(Bucket="spade", Prefix=""),
+            mock.call.get_paginator("list_object_versions"),
+            mock.call.get_paginator().paginate(Bucket="martello", Prefix=""),
+        ]
```
katstevens: Do these calls need to be sorted for threading as well?

risicle: _Shouldn't_ - the iterator is only ever advanced by the main thread.

katstevens: Ah yes - that's called before the map_callable is applied, isn't it. Cool.

PR #299, opened 2018-10-23
Update app requirements file

lfdebrux:
Whilst working with dmrunner on my laptop I noticed a bunch of red from `make requirements` due to conflicting versions in digitalmarketplace-scripts.

This PR should make those errors go away ;)

PR #301, opened 2018-10-24
Update requirements file (take 2)

lfdebrux:
This is a do-over for #299, last time it broke everything :sweat_smile:

The main change is pinning our version of docopt, this conflicts with mandrill (which requires an old version of docopt, which is what broken everything) but as we don't use mandrill pinning docopt is preferable.

I've tested the index script which failed on my machine with the same `--create-mapping=` argument which caused issues last time and it worked for me :+1:


risicle:
Unless you know that each of the inbetween states is a valid state for which the tests will pass I'd be careful putting this in as so many separate commits. If all you know is that the state before is valid and the state after is valid I'd consider it all part of an atomic change which shouldn't be split.

lfdebrux:
@risicle I've updated the commit to put docopt at the front of the queue

PR #306, opened 2018-11-01
Delete old declarations

jonodrew:
# What
This PR creates a new class of methods for getting rid of declarations from supplier frameworks. It also extends an existing script to remove declarations older than three years.

# Why
It's part of our GDPR work. We shouldn't keep data any longer than we need it.

[Ticket](https://trello.com/c/aIsqTYgp/42-deleting-declaration-of-suppliers-after-3-years)


dmscripts/supplier_framework.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        This functions calls the api endpoint and returns a list of the supplier_ids of suppliers that applied to be on
+        the framework specified when the class was created but failed
+        :return: A list of ints that are supplier_id
+        :rtype: List[int]
+        """
+        return [
+            framework_supplier['supplierId']
+            for framework_supplier in self.api_client.find_framework_suppliers_iter(self.framework_slug)
+            if framework_supplier['onFramework'] == 'False'
```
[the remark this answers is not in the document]
jonodrew: Sausages. You're right, the actual output is boolean. Thanks for the catch!

tests/test_supplier_framework.py
```diff
@@ -0,0 +1,19 @@
...
+    def _supplier_framework_response(self):
+        with open("tests/fixtures/test_supplier_frameworks_response.json", 'r') as response_file:
+            return json.load(response_file)['supplierFrameworks']
+
+    def test_suppliers_application_failed_to_framework(self):
+        mocked_api_client = create_autospec(DataAPIClient)
+        mocked_api_client.find_framework_suppliers_iter.side_effect = lambda *args, **kwargs: \
+            iter(self._supplier_framework_response())
+        supplier_framework = SupplierFrameworkMethods(mocked_api_client, 'g-cloud-8')
+        assert supplier_framework.suppliers_application_failed_to_framework() == [12345, 23456]
```
risicle: You can probably test more here - in fact you can probably do proper testing of `remove_declaration_from_suppliers()` by doing some more setup of the mock api client's returns and then asserting the calls made afterwards. I'll show you if you like.

jonodrew: Yes please

scripts/data-retention-remove-user-data.py
```diff
@@ -52,6 +53,11 @@
         base_url=get_api_endpoint_from_stage(stage),
         auth_token=get_auth_token('api', stage)
     )
+    supplier_frameworks = SupplierFrameworkMethods(
+        api_client=data_api_client,
+        logger=logger,
+        dry_run=dry_run
+    )
```
katstevens: I'm not sure about this class/module name.

While our scripts repo isn't the tidiest, the file name structure is actually relatively consistent:
```
scripts/action-that-script-is-carrying-out.py    # top level script, parsing args, initialising api clients etc
dmscripts/action_that_script_is_carrying_out.py    # logic, email sending, api calls, other testable stuff - same name, so we can match up the top level script with its logic
dmscripts/helpers/thing_helpers.py   # any shared functions, classes etc that might apply to more than one script
```

I don't have a problem with class/method based scripts per se, but I'm not sure we should be instantiating it  here in the top level script. Also it feels like for `SupplierFrameworkMethods`, there's some crossover with `dmscripts/helpers/supplier_data_helpers.py` (perhaps we can extend the `SupplierFrameworkData` helper class instead?) and that the call to `supplier_frameworks.remove_declaration_from_suppliers(cutoff_date)` should go in `dmscripts` so it's testable. This is a bit of a rambly comment - happy to chat offline?

scripts/data-retention-remove-user-data.py
```diff
@@ -53,17 +53,5 @@
...
-        if last_logged_in_at < cutoff_date and not user['personalDataRemoved']:
-            logger.warn(
-                f"{prefix}Removing personal data for user: {user['id']}"
-            )
-            if not dry_run:
-                data_api_client.remove_user_personal_data(
-                    user['id'],
-                    'Data Retention Script {}'.format(datetime.now().isoformat())
-                )
+    remove_supplier_data(data_api_client, logger, dry_run, cutoff_date)
```
katstevens: I am starting to have second thoughts about doing the declaration deletion in this script.

The current script needs to run nightly, as each day might have a new set of users that have logged on >3 years ago. But frameworks expire much more rarely than this - twice a year at most. So the declaration deletions could be run as a one-off script (covering both failed suppliers and for any frameworks expired >3 years ago). This would also reduce the risk of accidentally deleting 'live' declarations. Let's discuss when you get back @jonodrew.

PR #308, opened 2018-11-13
Delete expired supplier declarations

jonodrew:
This is a new branch since I made a bit of a fudge of the previous one.

[Here's the old branch with discussion](https://github.com/alphagov/digitalmarketplace-scripts/pull/306) and here's [the ticket that started it all off](https://trello.com/c/aIsqTYgp/42-deleting-declaration-of-suppliers-after-3-years).


scripts/data-retention-remove-supplier-declarations.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python
+"""
+Our data retention policy is that users have accounts deactivated and personal data stripped after 3 years without
+a login.
```
katstevens: A minor tweak needed to this, something like: `Our data retention policy is that supplier declarations are removed  3 years after the framework's expiry date`. We can extend this when we do the failed applications bit.

scripts/data-retention-remove-supplier-declarations.py
```diff
@@ -0,0 +1,56 @@
...
+Options:
+    --stage=<stage>                                       Stage to target
+
+    --dry-run                                             List account that would have data stripped
+    --verbose
+    -h, --help                                            Show this screen
+
+Examples:
+    ./scripts/data-retention-remove-user-data.py preview
+    ./scripts/data-retention-remove-user-data.py preview --dry-run --verbose
```
katstevens: The examples and the usage will need updating with the script name, otherwise docopt will complain 😸

tests/helpers/test_supplier_data_helpers.py
```diff
@@ -2,6 +2,32 @@
...
+    def _supplier_framework_response(self):
+        with open("tests/fixtures/test_supplier_frameworks_response.json", 'r') as response_file:
+            return json.load(response_file)['supplierFrameworks']
+
+    def test_suppliers_application_failed_to_framework(self):
+        mocked_api_client = mock.create_autospec(DataAPIClient)
+        mocked_api_client.find_framework_suppliers_iter.side_effect = lambda *args, **kwargs: \
+            iter(self._supplier_framework_response())
+        supplier_framework = SupplierFrameworkDeclarations(mocked_api_client, mock.ANY, 'g-cloud-8', dry_run=False)
+        assert supplier_framework.suppliers_application_failed_to_framework() == [12345, 23456]
```
katstevens: One more assert on the `mocked_api_client. remove_supplier_declaration ` call would be ideal here, but we can add that in when we do the failed applications scenario.

tests/helpers/test_supplier_data_helpers.py
```diff
@@ -2,6 +2,32 @@
...
+
+    def test_remove_declaration_from_suppliers(self):
+        mocked_api_client = mock.create_autospec(DataAPIClient)
+        mocked_api_client.remove_supplier_declaration.return_value = {'declaration': {}}
+        sfd = SupplierFrameworkDeclarations(mocked_api_client, mock.ANY, 'g-cloud-8', dry_run=False)
+        assert sfd.remove_declaration(1, 'g-cloud-8')['declaration'] == {}
+        mocked_api_client.remove_supplier_declaration.assert_called_with(1, 'g-cloud-8', 'user')
```
katstevens: It'd be great if we tested the `sfd. remove_supplier_declaration_for_expired_frameworks(...)` method, as well as the lower level `sfd.remove_declaration`. Then we'd have confidence about the dates we're using in the API query (we want to make sure it's 3 years, right 😸), the other API calls we're making and the other helper methods on the class (e.g. `_frameworks_older_than_date`).

benvand:
This looks great, nice one!

PR #319, opened 2018-12-28
threading: make all uses of multithreaded map do so through dependency injection

risicle:
Stumbled upon this while trying to get the test suite to pass for https://github.com/alphagov/digitalmarketplace-scripts/pull/320 instead of deadlocking on unrelated tests.

This way different implementations can be supplied on invocation and e.g. we can call each in single threaded mode by supplying the builtin `map`, which is also used as the default argument everywhere so these functions can still be called from anywhere without worrying about the map implementation and still get a safe result.

This is particularly useful for tests, where running the code in multithreaded mode complicates things and can even cause deadlocks if an instance of `mock.Mock` is shared among several threads.

Something I'm a little nervous of with this is that we obviously don't have test coverage of the actual script entry points (i.e. `scripts/*`) which I've changed to now inject the thread pool. I *think* I've got them right though.

### DIFF
--- a/dmscripts/export_framework_applicant_details.py
+++ b/dmscripts/export_framework_applicant_details.py
@@ -189,7 +189,7 @@ def _create_row(framework_slug, record, count_statuses, framework_lot_slugs):
     ))
 
 
-def export_supplier_details(data_api_client, framework_slug, filename, framework_lot_slugs):
-    records = find_suppliers_with_details_and_draft_service_counts(data_api_client, framework_slug)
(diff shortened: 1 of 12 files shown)


risicle:
BTW it may look that I'm just screwing around with this PR, but without this patch the scripts repo tests currently hang on a deadlock for me in `tests/test_export_framework_results_reasons.py`

katstevens:
I see threading on `index_to_search_service.do_index` isn't included in this PR - is that one OK as is?

katstevens:
Really! I don't remember that 😹 OK - rest of it looks fine to me, will run the tests locally then 👍

PR #320, opened 2018-12-28
data-retention-remove-user-data: make this also check mailchimp mailing list membership

risicle:
More thoughts here https://trello.com/c/QROOUpoX/272-users-whose-personal-data-is-removed-are-not-removed-from-mailchimp-mailing-lists

Also, refactor into separate implementation function for easy testing, add simple tests.

Depends on https://github.com/alphagov/digitalmarketplace-utils/pull/483

### DIFF
--- a/dmscripts/data_retention_remove_user_data.py
+++ b/dmscripts/data_retention_remove_user_data.py
@@ -1,0 +1,56 @@
+from datetime import timedelta, datetime
+
+from dmutils.formats import DATETIME_FORMAT
+
+
+def data_retention_remove_user_data(
+    data_api_client,
+    logger,
+    dm_mailchimp_client=None,
+    dry_run=True,
+):
+    cutoff_date = datetime.now() - timedelta(days=365 * 3)
+    prefix = '[DRY RUN]: ' if dry_run else ''
+    all_users = data_api_client.find_users_iter()
+
+    for user in all_users:
+        last_logged_in_at = datetime.strptime(user['loggedInAt'], DATETIME_FORMAT)
+        if last_logged_in_at < cutoff_date:
+            if not user['personalDataRemoved']:
+                logger.warn(
+                    f"{prefix}Removing personal data in API for user: {user['id']}"
+                )
+                if not dry_run:
+                    data_api_client.remove_user_personal_data(
+                        user['id'],
+                        'Data Retention Script {}'.format(datetime.now().isoformat())
+                    )
+
+            if dm_mailchimp_client is not None:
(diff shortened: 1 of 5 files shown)


scripts/data-retention-remove-user-data.py
```diff
@@ -41,29 +44,34 @@
...
     data_api_client = DataAPIClient(
         base_url=get_api_endpoint_from_stage(stage),
-        auth_token=get_auth_token('api', stage)
+        auth_token=get_auth_token('api', stage),
+    )
+
+    if bool(arguments["<mailchimp_username>"]) != bool(arguments["<mailchimp_api_key>"]):
+        raise TypeError(
+            "Either both of '--mailchimp-api-key' and '--mailchimp-username' need to be specified or neither"
+        )
```
katstevens: 🏆 I wish our other mailchimp scripts did this check...

dmscripts/data_retention_remove_user_data.py
```diff
@@ -0,0 +1,56 @@
...
+                            prefix,
+                            email_hash,
+                            mailing_list["list_id"],
+                            mailing_list["name"],
+                        )
+                        if not dry_run:
+                            dm_mailchimp_client.permanently_remove_email_from_list(
+                                email_address=user["emailAddress"],
+                                list_id=mailing_list["list_id"],
+                            )
```
katstevens: I mentioned this on the utils PR - can we handle a False return value here, so that the script output can show that the deletion failed?

risicle: I think it does? The method itself should output an `error` of `Mailchimp failed to permanently remove user ({hashed_email}) from list ({list_id})` if this happens.

scripts/data-retention-remove-user-data.py
```diff
@@ -1,15 +1,17 @@
 #!/usr/bin/env python
 """
 Our data retention policy is that users have accounts deactivated and personal data stripped after 3 years without
-a login.
+a login. This script not only does that but also removes any such users' email addresses from any of our mailing lists.
 
-This script is very simple and has not been upgraded to accept any arguments to prevent the possibility of accidental
-deletion. If you are in doubt use the dry run option.
+This script is simple and has deliberately not been designed without many arguments to reduce the possibility of
+accidental deletion. If you are in doubt use the dry run option.
```
katstevens: Just spotted this - I'd remove the "not" from "deliberately not designed"...

risicle: Bloody hell

risicle:
Ok - about to merge, let's watch the mailchimp jobs...

PR #321, opened 2019-01-03
Fixup audit event

jonodrew:
This is to fixup something @benvand spotted. I've changed over the record for the audit event so that it's got more detail, in case we need to go back and check what happened when.

### DIFF
--- a/dmscripts/helpers/supplier_data_helpers.py
+++ b/dmscripts/helpers/supplier_data_helpers.py
@@ -202,7 +202,7 @@ class SupplierFrameworkDeclarations:
             return self.api_client.remove_supplier_declaration(
                 supplier_id,
                 framework_slug,
-                self.user
+                user=f'{self.user} {datetime.now().isoformat()}'
             )
 
     def remove_declaration_from_failed_applicants(self, framework_slug):
--- a/dmscripts/helpers/supplier_data_helpers.py
+++ b/dmscripts/helpers/supplier_data_helpers.py
@@ -160,10 +160,11 @@ class SupplierFrameworkDeclarations:
     Methods for manipulating supplier declarations
     """
 
-    def __init__(self, api_client, logger, dry_run):
+    def __init__(self, api_client, logger, dry_run, user):
         self.api_client = api_client
         self.dry_run = dry_run
         self.logger = logger
+        self.user = user  # passed through by the script that calls this
 
     def suppliers_application_failed_to_framework(self, framework_slug):
         """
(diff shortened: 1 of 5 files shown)


dmscripts/helpers/supplier_data_helpers.py
```diff
@@ -198,7 +198,11 @@ def remove_declaration(self, supplier_id, framework_slug):
                 f"Declaration of supplier {supplier_id} "
                 f"that applied to framework {framework_slug} removed"
             )
-            return self.api_client.remove_supplier_declaration(supplier_id, framework_slug, 'user')
+            return self.api_client.remove_supplier_declaration(
+                supplier_id,
+                framework_slug,
+                'Data Retention Script {}'.format(datetime.now().isoformat())
```
benvand: This assumes that the only script to ever use this method will want to sign their audit event as `Data Retention Script *`.
It removes the generic element of this class and arguably its reason to belong in helpers.

Passing the actor/ user through from the script would open this up to be more generic again.

jonodrew: @benvand I think this is fixed now

scripts/data-retention-remove-supplier-declarations.py
```diff
@@ -43,6 +44,7 @@
 
     dry_run = arguments['--dry-run']
     verbose = arguments['--verbose']
+    user = arguments['<user>']
```
katstevens: Let's set a default with `getpass` in case the arg is missing (see https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/insert-framework-results.py#L35 for an example).

jonodrew: Ugh. And I did this on the other one as well. Good spot, thank you!

tests/helpers/test_supplier_data_helpers.py
```diff
@@ -45,18 +46,37 @@ def _supplier_framework_response(self):
...
+                mocked_api_client,
+                mock.MagicMock(),
+                dry_run=False,
+                user='Data Retention Script {}'.format(datetime.now().isoformat())
+            )
+            assert sfd.remove_declaration(1, 'g-cloud-8')['declaration'] == {}
+            mocked_api_client.remove_supplier_declaration.assert_called_with(
+                1,
+                'g-cloud-8',
+                'Data Retention Script 2019-01-01T12:00:00')
```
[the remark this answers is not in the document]
jonodrew: I assumed it would be passed through as the user, but if that's not the case then I'll need to add it to `SupplierFrameworkDeclarations.remove_declaration`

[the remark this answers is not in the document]
katstevens: Well, it depends if we need it or not. If not, we can simplify these tests a bit.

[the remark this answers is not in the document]
jonodrew: I think we should - @benvand pointed out that it's the format we use elsewhere: https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/data-retention-remove-contact-information-data.py#L74

PR #327, opened 2019-01-10
data_retention_remove_user_data: don't attempt to remove users from mailchimp if already personalDataRemoved

risicle:
Another stupidity I've realized related to my implementation of https://trello.com/c/QROOUpoX

If `user["personalDataRemoved"]`, then the `emailAddress` we're given will already be garbage and looking it up in mailchimp is meaningless.

### DIFF
--- a/dmscripts/data_retention_remove_user_data.py
+++ b/dmscripts/data_retention_remove_user_data.py
@@ -20,40 +20,40 @@ def data_retention_remove_user_data(
     for user in all_users:
         last_logged_in_at = datetime.strptime(user['loggedInAt'], DATETIME_FORMAT)
         if last_logged_in_at < cutoff_date:
-            if dm_mailchimp_client is not None:
-                email_hash = dm_mailchimp_client.get_email_hash(user["emailAddress"])
-                logger.info(
-                    "Checking mailing list membership for email with hash %s (user %s)",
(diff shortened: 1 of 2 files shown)


dmscripts/data_retention_remove_user_data.py
```diff
@@ -20,40 +20,40 @@ def data_retention_remove_user_data(
...
-                            rm_result = dm_mailchimp_client.permanently_remove_email_from_list(
-                                email_address=user["emailAddress"],
-                                list_id=mailing_list["list_id"],
-                            )
-                            if not rm_result:
-                                raise MailchimpRemovalFailed(
-                                    "Mailchimp failure trying to permanently_remove_email_from_list"
-                                )
-                else:
+            if not user['personalDataRemoved']:
```
katstevens: We can filter our users by this on the API I think - `data_api_client.find_users_iter(personal_data_removed=False)`?

risicle: Done

katstevens: Er.. did you push the commit?

risicle: The question is - did I push the _right_ commit?

PR #328, opened 2019-01-11
Rename send-dos-opportunities to notify-suppliers-of-new-dos-opportunities

lfdebrux:
Renamed script to fit with naming pattern of other email scripts.

Although the script doesn't actually use Notify, I think this is an implementation detail ;)

If this PR does get merged then we will also need to merge a PR to change the Jenkins job (this is pending).

This is part of https://trello.com/c/aRpNbH4o/309-rename-send-dos-opportunities-emailpy-to-notify-suppliers-something-somethingpy

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ The `scripts` folder in this repository contains scripts that interact with the 
 
   Checks all free-text fields in G-Cloud services for "bad words" and generates a CSV report of any bad words found.
 
-* `send-dos-opportunities-email.py`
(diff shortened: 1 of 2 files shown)


katstevens:
Can we hold off renaming this one until we actually switch it? I definitely have a mental shorthand of seeing 'notify' in the script name and assuming that script uses Notify. Sorry!

lfdebrux:
That makes sense!

PR #329, opened 2019-01-11
Fix duplicate logging output from send_dos_opportunitues_emails

lfdebrux:
We were getting duplicated log output when running the `send-dos-opportunities-emails.py` script.

This commit fixes this by removing the logging configuration from the `send_dos_opportunities_emails.py` module, which was causing multiple logs with the same contents to exist.

Closes https://trello.com/c/u18CsU00/313-statements-are-logged-twice-on-jenkins-during-dos-opportunities-mailchimp-job

### DIFF
--- a/dmscripts/send_dos_opportunities_email.py
+++ b/dmscripts/send_dos_opportunities_email.py
@@ -1,12 +1,11 @@
 from datetime import datetime, date, timedelta
 
 from dmscripts.helpers.html_helpers import render_html
-from dmscripts.helpers import logging_helpers
 from dmscripts.helpers.logging_helpers import logging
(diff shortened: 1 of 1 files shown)


lfdebrux:
I've tested this locally and it seemed to do the trick

katstevens:
The parent script doesn't set logging for `dmapiclient` - are we bothered about this?

lfdebrux:
> The parent script doesn't set logging for `dmapiclient` - are we bothered about this?

The documentation for the `configure_logging` function says that `dmapiclient` is set to `INFO` by default, so I didn't bother changing the script, but it might be better to be more explicit. What do you think?

katstevens:
Ah, perfect 👍

PR #330, opened 2019-01-11
add export-outcomes-to-csv script

risicle:
https://trello.com/c/t2N1EqPk

I can't think of much interesting to say about this.

### DIFF
--- a/dmscripts/export_outcomes_to_csv.py
+++ b/dmscripts/export_outcomes_to_csv.py
@@ -1,0 +1,42 @@
+import csv
+
+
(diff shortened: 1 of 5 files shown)


katstevens:
Just to check, are we going to handle the file upload within the Jenkins job as per the other export script?

Also I wonder if it's worth including some basic logging, maybe a count of the number of outcomes?

Otherwise looks great! Neat stuff on the nested value fetching.

risicle:
Uploads? I ... quite honestly hadn't thought that far ahead yet.

Logging - perhaps I should just hook up the logging output from the apiclient.

PR #332, opened 2019-01-15
get-model-data: add "completed outcomes" export

risicle:
https://trello.com/c/t2N1EqPk/

This required changing the way column headers are generated from "nested fields". Previously the innermost field name of a chain would be picked, but this is vulnerable to clashes, as will happen when multiple nested dictionaries with `id` attributes are desired. The change is to `.`-concatenate the field names. The only effect this should have on existing exports is changing one header in `briefs.csv` from `emailAddress` to `users.0.emailAddress`. Ash reckons this shouldn't break any of his stuff but we're going to make sure with a dry run from a local machine before merging.


risicle:
Ash is apparently happy

PR #335, opened 2019-01-21
Tidy signature page generation script

lfdebrux:
https://trello.com/c/jEZzVo2u/326-tidy-signature-page-generation-script

Refactoring the signature page generation script to be more verbose and more user-friendly.

Now with added functional tests! Functional tests require https://github.com/alphagov/digitalmarketplace-test-utils/pull/20, Travis won't pass until this is merged.

### DIFF
--- a/functional-tests/generate-framework-agreement-signature-pages/setup.sh
+++ b/functional-tests/generate-framework-agreement-signature-pages/setup.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+. "$TESTDIR"/../setup.sh
+
+# generate-framework-agreement-signature-pages generally requires the
+# agreements repo
+[ -n "$DM_AGREEMENTS_REPO" ] || DM_AGREEMENTS_REPO="$DM_SCRIPTS_REPO/../digitalmarketplace-agreements"
+
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,3 +9,5 @@ lxml
 mock
 pytest
 requests-mock
+
+git+https://github.com/alphagov/digitalmarketplace-test-utils.git@2.3.0#egg=digitalmarketplace-test-utils==2.3.0
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,5 @@
 -r requirements.txt
+cram
 flake8==3.5.0
 flake8-per-file-ignores
 freezegun
--- a/scripts/generate-framework-agreement-signature-pages.py
+++ b/scripts/generate-framework-agreement-signature-pages.py
@@ -47,11 +47,7 @@ sys.path.insert(0, ".")
 
 from docopt import docopt
 
-from dmscripts.export_framework_applicant_details import get_csv_rows
 from dmscripts.helpers.auth_helpers import get_auth_token
-from dmscripts.helpers.framework_helpers import (
-    find_suppliers_with_details_and_draft_service_counts
-)
 from dmscripts.helpers.logging_helpers import (
     configure_logger,
     logging,
@@ -62,6 +58,7 @@ from dmapiclient import DataAPIClient
 from dmutils.env_helpers import get_api_endpoint_from_stage
 
 from dmscripts.generate_framework_agreement_signature_pages import (
+    find_suppliers,
     render_html_for_successful_suppliers,
     render_pdf_for_each_html_page,
 )
@@ -107,31 +104,15 @@ if __name__ == "__main__":
 
     logger.debug(f"fetching lots for framework '{framework_slug}'")
     framework = client.get_framework(framework_slug)["frameworks"]
-    framework_lot_slugs = tuple([lot["slug"] for lot in client.get_framework(framework_slug)["frameworks"]["lots"]])
 
-    # get supplier details (returns a lazy generator)
-    logger.debug(f"fetching records for {len(supplier_ids) if supplier_ids else 'all'} suppliers")
-    records = find_suppliers_with_details_and_draft_service_counts(
-        client,
-        framework_slug,
-        supplier_ids,
-        map_impl=map_impl,
-    )
-    # we reuse code from another script to filter and flatten our supplier details
-    _, rows = get_csv_rows(
-        records,
-        framework_slug,
-        framework_lot_slugs,
-        count_statuses=("submitted",),
-        dry_run=dry_run,
-    )
+    suppliers = find_suppliers(client, framework, supplier_ids, map_impl, dry_run)
 
     # create a temporary directory for the HTML files
     with tempfile.TemporaryDirectory() as html_dir:
         # create signature pages in HTML using Jinja templates from agreements repo
         logger.debug(f"generating HTML signature pages")
         render_html_for_successful_suppliers(
-            rows, framework, agreements_dir, html_dir, dry_run)
+            suppliers, framework, agreements_dir, html_dir, dry_run)
 
         # convert HTML to PDF (this uses wkhtmltopdf under-the-hood)
         logger.debug(f"generating PDF signature pages from HTML")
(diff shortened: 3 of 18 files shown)


katstevens:
Would a `--dry-run` flag be useful? We're only churning out PDFs but 1) all our other scripts have it 2) it might be quicker for debugging purposes...

lfdebrux:
> Would a `--dry-run` flag be useful? We're only churning out PDFs but 1) all our other scripts have it 2) it might be quicker for debugging purposes...

Hmmm, I although I agree that it would be useful to have a --dry-run flag, I can't really think of what the script would do in that case... Just print out suppliers it would generate pdfs for?

katstevens:
I think that's a sign we should add some logging to this script! How about printing out the supplier ID when it's a single supplier, and the total number of suppliers if it's from a file?

scripts/generate-framework-agreement-signature-pages.py
```diff
@@ -63,28 +66,44 @@
...
+        supplier_ids = args['--supplier-id']
+    else:
+        supplier_ids = None
+
     framework_slug = args['<framework_slug>']
-    client = DataAPIClient(get_api_endpoint_from_stage(args['<stage>']), get_auth_token('api', args['<stage>']))
-    framework = client.get_framework(framework_slug)['frameworks']
-    framework_lot_slugs = tuple([lot['slug'] for lot in client.get_framework(framework_slug)['frameworks']['lots']])
+    agreements_repo = pathlib.Path(args['<path_to_agreements_repo>'])
+    agreements_dir = agreements_repo / 'documents' / framework_slug
```
katstevens: Whooooo fancy Python 3 path syntax! I love it.

scripts/generate-framework-agreement-signature-pages.py
```diff
@@ -63,28 +66,44 @@
 if __name__ == '__main__':
     args = docopt(__doc__)
 
+    if args['<supplier_id_file>']:
+        supplier_ids = get_supplier_ids_from_file(args['<supplier_id_file>'])
+    elif args['--supplier-id']:
+        supplier_ids = args['--supplier-id']
+    else:
+        supplier_ids = None
```
katstevens: I think we should fail noisily if there are no suppliers provided - if we're running this script then we should have some! At the very least let's log that nothing's going to happen.

scripts/generate-framework-agreement-signature-pages.py
```diff
@@ -36,13 +36,16 @@
 If supplied, <supplier_id_file> is a text file with one supplier ID per line; framework agreement pages will only be
 generated for these suppliers.
 
+Alternatively you can supply one or more supplier IDs at the command line using the `--supplier-id` flag.
+
 Usage:
     scripts/generate-framework-agreement-signature-pages.py <stage> <framework_slug> <output_folder>
-    <path_to_agreements_repo> [<supplier_id_file>]
+    <path_to_agreements_repo> [(--supplier-id=<id>... | <supplier_id_file>)]
```
katstevens: I can't get this to run locally with the ID or the file 🤔 I might be doing something wrong?

`./scripts/generate-framework-agreement-signature-pages.py local digital-outcomes-and-specialists-3 output ../digitalmarketplace-agreements --supplier-id=710702`

scripts/generate-framework-agreement-signature-pages.py
lfdebrux: Ah, if you don't supply a supplier ID then it gets all suppliers; that should be documented

lfdebrux: This is now documented, see https://github.com/alphagov/digitalmarketplace-scripts/pull/335/files#diff-7df3cea3a04365ea67135cd2d5e6d21aR26

lfdebrux:
I've added some fairly intrusive logging, anyone want to try our the new version of the script?

scripts/generate-framework-agreement-signature-pages.py
```diff
@@ -1,16 +1,43 @@
 #!/usr/bin/env python
 """
+Generate framework agreement signature pages from supplier "about you"
+information for suppliers who successfully applied to a framework.
 
-PREREQUISITE: You'll need wkhtmltopdf installed for this to work (http://wkhtmltopdf.org/)
+Usage:
+    scripts/generate-framework-agreement-signature-pages.py [-n] [-v...]
+        <stage> <framework> <output_dir> <agreements_repo>
+        [--supplier-id=<id>... | --supplier-ids-from=<file>]
```
katstevens: The script below is still looking for `<supplier_id_file>`. But changing it below to `<file>` doesn't work for me, neither does changing this line to `<supplier_id_file>`!  Docopt strikes again 😿

lfdebrux: Good spot, I'll fix that now

lfdebrux:
Okay, I think I'm happy with the dry run functionality and level of output for this script now.

Next is more tests.

requirements-dev.txt
```diff
@@ -8,3 +9,5 @@ lxml
 mock
 pytest
 requests-mock
+
+git+https://github.com/alphagov/digitalmarketplace-test-utils.git@2.2.0#egg=digitalmarketplace-test-utilts==2.2.0
```
katstevens: Just trying this out locally - I think there's a typo here: `test-utilts`

lfdebrux: Nice spot

requirements-dev.txt
```diff
@@ -8,3 +9,5 @@ lxml
 mock
 pytest
 requests-mock
+
+git+https://github.com/alphagov/digitalmarketplace-test-utils.git@2.3.0#egg=digitalmarketplace-test-utils==2.3.0
```
jonodrew: I think this is a typo - I can't find a version 2.3.0?

lfdebrux: See https://github.com/alphagov/digitalmarketplace-test-utils/pull/20

functional-tests/generate-framework-agreement-signature-pages/setup.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+. "$TESTDIR"/../setup.sh
+
+# generate-framework-agreement-signature-pages generally requires the
+# agreements repo
+[ -n "$DM_AGREEMENTS_REPO" ] || DM_AGREEMENTS_REPO="$DM_SCRIPTS_REPO/../digitalmarketplace-agreements"
+
```
benvand: Is this file run automatically by cram?

benvand:
Changes look positive to me. Nice one!

Couple points but not blockers:
* Not 100% sure about the `setup.sh` files. Are they a `cram` thing? Looks like most of it could be Make commands instead 🤷‍♂️ 
* Generally it's recommended to use implicit line continuation rather than backslash line continuation https://github.com/google/styleguide/blob/gh-pages/pyguide.md#32-line-length

functional-tests/generate-framework-agreement-signature-pages/setup.sh
lfdebrux: No the test script that wants to use it can source it. See for instance `functional-tests/generate-framework-agreement-signature-pages/signature-pages.t`.

lfdebrux:
> Changes look positive to me. Nice one!

Thanks! Thanks for wading through them ;)

> Couple points but not blockers:
> 
> * Not 100% sure about the `setup.sh` files. Are they a `cram` thing? Looks like most of it could be Make commands instead :man_shrugging:

That's true, however I feel they are useful as it means you can if you wish run `cram` standalone, which I found useful.

> * Generally it's recommended to use implicit line continuation rather than backslash line continuation https://github.com/google/styleguide/blob/gh-pages/pyguide.md#32-line-length

Yeah I prefer the implicit line continuation normally, but for some reason putting a `lambda` in brackets made me all weirded out...

PR #337, opened 2019-02-06
Don't fetch binaries from s3

lfdebrux:
The Dockerfile for digitalmarketplace-scripts currently fetches some binary dependencies from an s3 bucket managed by us. This means we are responsible for keeping these binaries up-to-date, which is a big pain.

A simpler method is to fetch the binaries from the GitHub release page; although we are still responsible for keeping the packages up-to-date, at least this way we don't have to log in to AWS ;)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,16 @@
 FROM python:3.6-slim
 
-ENV APP_DIR /app
-
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends gcc g++ make git libffi-dev libssl-dev libc6-dev wget curl socat jq
+RUN wget --quiet -O /usr/local/bin/aws-auth https://raw.githubusercontent.com/alphagov/aws-auth/1741ad8b8454f54dd40fb730645fc2d6e3ed9ea9/aws-auth.sh \
+    && chmod 0755 /usr/local/bin/aws-auth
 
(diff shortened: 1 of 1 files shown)


jonodrew:
What's the thinking behind splitting up the `RUN` steps? I think that'll end up with bigger docker images and I'm not sure what the payoff is.

lfdebrux:
> What's the thinking behind splitting up the `RUN` steps? I think that'll end up with bigger docker images and I'm not sure what the payoff is.

Why will it result in larger Docker images?

jonodrew:
> > What's the thinking behind splitting up the `RUN` steps? I think that'll end up with bigger docker images and I'm not sure what the payoff is.
> 
> Why will it result in larger Docker images?
> 
> The main advantage is that it's easier to read, additionally the cache won't be invalidated for a change to one of the binaries (that probably doesn't matter that much).

My understanding was that each command in a Dockerfile represents a new layer, and therefore increases the size. I agree about readability though.

lfdebrux:
> > > What's the thinking behind splitting up the `RUN` steps? I think that'll end up with bigger docker images and I'm not sure what the payoff is.
> > 
> > 
> > Why will it result in larger Docker images?
> > The main advantage is that it's easier to read, additionally the cache won't be invalidated for a change to one of the binaries (that probably doesn't matter that much).
> 
> My understanding was that each command in a Dockerfile represents a new layer, and therefore increases the size. I agree about readability though.

In this case I think it's worth it for the extra readability, plus the size overhead probably doesn't make much difference as we're adding some fairly large binaries to the image.

PR #347, opened 2019-03-05
Add basic script for generating categories/reseller report

jonodrew:
This adds a script requested by our performance analyst. It gives them data about which of our suppliers are resellers and which are selling original products. It also lists the categories each service has registered under. The script should be run against a local dump, as it hammers the API fairly hard. It outputs to TSV, because working with commas is hard and this was the path of least resistance.

[Ticket](https://trello.com/c/M5EBzmbK/398-export-g-cloud-service-data-for-benchmarking)

### DIFF
--- a/scripts/oneoff/categories-baseline.py
+++ b/scripts/oneoff/categories-baseline.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python
+"""Usage: categories-baseline.py <g-cloud-version> [--output-folder=OUTPUT] [--stage=STAGE]
+
+This very basic script is run for our performance analyst. It gives them an idea of what categories services are being
+offered under. It also shows who is a reseller, and who is selling original products. As this script hammers the API to
+get DUNS numbers, I strongly recommend it's only run locally against a recent database dump. So I've written the script
+to default to 'local' and users will need to specifically ask it to run against a different stage.
+
+The script generates three files, one for each lot. By default it drops them into the local folder,
+ie digitalmarketplace-scripts/
+If you'd like them to go somewhere else, supply a folder path to --output-path
+
+The script doesn't output any logging, but if you run it with DMRunner you can watch as it zooms through the API. It
+takes about 15m to run through G-Cloud 10, which has about 25,000 items.
+
+Options:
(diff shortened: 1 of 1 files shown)


scripts/oneoff/categories-baseline.py
```diff
@@ -0,0 +1,72 @@
...
+    # Set defaults, instantiate clients
+    data_api_client = DataAPIClient(
+        base_url=get_api_endpoint_from_stage(stage),
+        auth_token=get_auth_token('api', stage)
+    )
+
+    hosting_services = []
+    software_services = []
+    support_services = []
+    for service in data_api_client.find_services_iter(framework=f'{version}', status='published'):
```
katstevens: I think `find_services_iter` has a `lot` param that could let the API do the hard work and speed this bit up. 

```
support_services = data_api_client.find_services_iter(framework=f'{version}', status='published', lot='cloud-support')
hosting_services = data_api_client.find_services_iter(framework=f'{version}', status='published', lot='cloud-hosting')
software_services = data_api_client.find_services_iter(framework=f'{version}', status='published', lot='cloud-software')
```

scripts/oneoff/categories-baseline.py
```diff
@@ -0,0 +1,72 @@
...
+            hosting_services.append(service)
+        elif service.get('lot') == 'cloud-software':
+            software_services.append(service)
+
+    lots = {
+        'hosting': hosting_services,
+        'software': software_services,
+        'support': support_services
+    }
+    for key, value in lots.items():
```
katstevens: How about naming this `services_for_lot` instead of `value`?

scripts/oneoff/categories-baseline.py
```diff
@@ -0,0 +1,72 @@
...
+        with open(f'{key}-categories-{version}.tsv', 'w') as f:
```
katstevens: I think one file per lot is a good idea, but can we 1) mention that this is the expected output in the top docstring 2) maybe offer an optional destination folder arg, so the files don't have to hang around in `scripts/oneoff`?

jonodrew:
Thanks for that feedback Kat. Fixed up

PR #353, opened 2019-03-20
Add script to update supplier attributes

jonodrew:
This is a one-off script that needs a CSV input. The column titles should be 'DMP record ID', 'Field Name for update', and 'New Value'. It'll go through and update the value given.

As it stands this has too many prints, but the data needs testing before we commit to running it.

[Ticket for context](https://trello.com/c/RMLrneG5/423-data-controller-batch-update)

### DIFF
--- a/scripts/oneoff/update-supplier-data.py
+++ b/scripts/oneoff/update-supplier-data.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python
+"""Usage: update-supplier-data.py <path-to-updates> [--stage=STAGE] [--attribute=ATTRIBUTE] [--dry-run]
+
+This script iterates through a CSV and updates a supplier with one of three values: DUNS number, company registration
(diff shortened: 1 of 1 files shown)


scripts/oneoff/update-supplier-data.py
```diff
@@ -0,0 +1,111 @@
...
+    field_to_change = row.get("Field Name for update")
+    try:
+        attribute_map.get(field_to_change)(supplier_id, new_value)
+    except HTTPError as e:
+        print(f"Error updating supplier {supplier_id} with error {e}")
+
+
+def update_duns(supplier_id, new_duns):
+    try:
+        data_api_client.update_supplier(supplier_id, {"dunsNumber": new_duns}, UPDATER)
```
katstevens: Can we have some log output on the happy path too - so we have an idea of what's happening when we run for real?

PR #355, opened 2019-03-27
[WIP] Allow bad words script to scan draft services

katstevens:
Trello: https://trello.com/c/Q0tQI9xL/439-spike-scan-g-cloud-services-for-bad-words-script-can-we-run-this-against-draft-services

- Adds an optional arg `--scan-drafts` that allows the script to scan services before converting them to 'real' services (i.e. we have more time to deal with any problem services before indexing them)
- Avoids hardcoding the fields to scan by looking in the frameworks repo instead (see https://github.com/alphagov/digitalmarketplace-frameworks/pull/559) This does mean that only one iteration's services can be scanned at once, but seeing as that's what we end up doing anyway...
- Some general reshuffling/tidying 
- TBC: ...tests? 🙀 

This PR also gives a sneak preview of my secret plan to move all the framework scripts into a `/scripts/framework-applications` folder. Behold, the new era of 'knowing where to find things when flipping frameworks'.


lfdebrux:
> This PR also gives a sneak preview of my secret plan to move all the framework scripts into a /scripts/framework-applications folder. Behold, the new era of 'knowing where to find things when flipping frameworks'.

😱

PR #357, opened 2019-04-03
Fix csv export so only aborts if no data found

lfdebrux:
We had an issue where CSVs with no rows were being created legitimately (because there were no entries in the database that fit the criteria), but the CSV export script was failing to upload this to s3 as it thought this was due to an error.

This PR changes the `generate_csv_and_upload_to_s3` procedure so it only aborts if it does not have a valid s3 filename.

This does mean that in future it is possible for CSVs with only a header row to be uploaded.


katstevens:
Looks good - we have test coverage for this script, worth adding in a test for the empty rows case?

lfdebrux:
> Looks good - we have test coverage for this script, worth adding in a test for the empty rows case?

Done

PR #361, opened 2019-04-15
Change so that no suppliers/users is a warning only

lfdebrux:
Previously generate_supplier/user_csv would return early if there was no
data from the server, effectively treating this as an error.

However it is very possible that there might be no data for whatever
reason; we should warn but then continue as normal.

### DIFF
--- a/dmscripts/generate_supplier_user_csv.py
+++ b/dmscripts/generate_supplier_user_csv.py
@@ -141,6 +141,10 @@ def generate_csv_and_upload_to_s3(
     else:
         headers, rows, download_filename = generate_supplier_csv(framework_slug, data_api_client, logger=logger)
 
+    # no need to create an empty csv
+    if not rows:
+        return True
+
     # Save CSV to output dir and upload to S3
     file_path = csv_path(output_dir, download_filename)
     _build_csv(file_path, headers, rows)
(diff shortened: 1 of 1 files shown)


dmscripts/generate_supplier_user_csv.py
```diff
@@ -10,7 +10,6 @@ def generate_supplier_csv(framework_slug, data_api_client, logger):
     supplier_rows = data_api_client.export_suppliers(framework_slug).get('suppliers', [])
     if not supplier_rows:
         logger.warn('No supplier data found for framework {}'.format(framework_slug))
-        return [], [], None
```
katstevens: Can we instead add a line above the `if not download_filename:` clause on L146. Something like

```
if not rows:
    logger.info("No data for {}".format(whatever_it_is))
    return True
```

We also have tests for this script - should probably update them with the new expected behaviour.

lfdebrux: Is the idea to not upload empty files? I think in that case it would make more sense to put the test in `generate_csv_and_upload_to_s3`?

lfdebrux: Added a commit with a short-circuit in `generate_csv_and_upload_to_s3` https://github.com/alphagov/digitalmarketplace-scripts/pull/361/commits/e33c190aeeb74545bba669b44aeac32ab178fe2e

dmscripts/generate_supplier_user_csv.py
```diff
@@ -146,6 +144,10 @@ def generate_csv_and_upload_to_s3(
     if not download_filename:
```
katstevens: We can get rid of this line now, then - `generate_user_csv` and `generate_supplier_csv` will always return a filename.

PR #364, opened 2019-04-23
requires apiclient 20.0.0

agalamatis:
Scripts using the new default with_declarations (which is now True, instead of None) are expected to slow down.


katstevens:
So if `find_framework_suppliers` now has `with_declarations=True` by default, then can we update our scripts with `client.find_framework_suppliers(with_declarations=False)` to get the same behaviour/speed?

PR #365, opened 2019-04-25
Script to swap DUNS numbers for two suppliers

katstevens:
Trello: https://trello.com/c/l5647BuW/733-duns-change-from-dave

Slightly tidied-up version of the script I've been using to swap DUNS numbers for two suppliers. It does the update in 3 steps, using a temporary dummy DUNS. It's verbose on purpose to make it clear what's happening at each step.

It performs some basic checks before attempting the swap:
- the suppliers should not already have the target DUNS numbers
- the existing DUNS numbers for the suppliers should match the target DUNS numbers (i.e. you can only set supplier 1's new DUNS to supplier 2's existing DUNS, and vice versa)
- the two target DUNS numbers should be different 
- the target DUNS numbers should be 9 digits

### DIFF
--- a/scripts/oneoff/swap-supplier-duns.py
+++ b/scripts/oneoff/swap-supplier-duns.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python
+"""
+This script attempts to swap the DUNS numbers for two suppliers. This involves temporarily setting one of the suppliers
+to a fake DUNS number to satisfy the database's uniqueness constraint.
+
+Usage: swap-supplier-duns.py <supplier_1> <supplier_2> <duns_1> <duns_2> [options]
+
+    <supplier_1>                                          Supplier ID #1
+    <supplier_2>                                          Supplier ID #2
+    <duns_1>                                              DUNS number #1
+    <duns_2>                                              DUNS number #2
+
+Options:
+    --stage=<stage>                                       Stage to target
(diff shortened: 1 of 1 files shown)


scripts/oneoff/swap-supplier-duns.py
```diff
@@ -0,0 +1,141 @@
...
+
+    <supplier_1>                                          Supplier ID #1
+    <supplier_2>                                          Supplier ID #2
+    <duns_1>                                              DUNS number #1
+    <duns_2>                                              DUNS number #2
+
+Options:
+    --stage=<stage>                                       Stage to target
+    --updated-by=<updated-by>                             Updater email address for audit trail
+    --dummy-duns=<dummy-duns>                             Dummy DUNS number
```
lfdebrux: You could specify the default dummy DUNS number here, which might make it easier to find

```suggestion
    --dummy-duns=<dummy-duns>                             Dummy DUNS number [default: 0000000002]
```

lfdebrux:
Can we simplify the usage so that the user doesn't need to supply the DUNS numbers? 🤔

katstevens:
@lfdebrux I'd rather not, to be honest. This script should be run rarely, and the developer should have had a specific request from CCS before running it, with details of the DUNS to be changed.

PR #366, opened 2019-04-25
Hide fake suppliers

agalamatis:
https://trello.com/c/Px4kcY6B/462-exclude-fake-suppliers-from-framework-results-script

New optional keyword parameters to exclude fake suppliers for:
- generate-framework-master
- mark-definite-framework-results
- export-framework-results-reasons

PS: Also bumped the apiclient and explicitly calling `find_framework_suppliers` with `with_declarations=None` which used to be the default.

### DIFF
--- a/dmscripts/mark_definite_framework_results.py
+++ b/dmscripts/mark_definite_framework_results.py
@@ -83,11 +83,20 @@ def mark_definite_framework_results(
     dry_run=True,
     supplier_ids=None,
     logger=logging.getLogger("script"),
+    excluded_supplier_ids=None,
 ):
     interested_supplier_ids = supplier_ids or client.get_interested_suppliers(
         framework_slug
     ).get('interestedSuppliers', ())
     total_interested_suppliers = len(interested_supplier_ids)
+
+    # exclude suppliers with IDs the executioner defined
+    if excluded_supplier_ids is not None:
+        for excluded_supplier in excluded_supplier_ids.split(','):
+            try:
+                interested_supplier_ids.pop(interested_supplier_ids.index(int(excluded_supplier)))
+            except IndexError:  # occurs when .index() fails
+                continue
 
     # Loop over suppliers breaking out if they pass or fail, if they make it to the end they get a discretionary pass
     for i, supplier_id in enumerate(interested_supplier_ids, start=1):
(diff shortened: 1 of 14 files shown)


dmscripts/mark_definite_framework_results.py
```diff
@@ -83,12 +83,21 @@ def mark_definite_framework_results(
...
     ).get('interestedSuppliers', ())
     total_interested_suppliers = len(interested_supplier_ids)
 
+    # exclude suppliers with IDs the executioner defined
+    if excluded_supplier_ids is not None:
+        for excluded_supplier in excluded_supplier_ids.split(','):
+            try:
+                interested_supplier_ids.pop(interested_supplier_ids.index(int(excluded_supplier)))
+            except IndexError:  # occurs when .index() fails
+                continue
```
risicle: Y'know we could just make both `interested_supplier_ids` and `excluded_supplier_ids` `set`s and do set operations here instead of faffy index lookups. It's not like we really care about the ordering is it?

scripts/mark-definite-framework-results.py
```diff
@@ -32,6 +32,7 @@
 --reassess-failed-draft-services  Don't skip draft_services with "failed" status
 -v, --verbose                     Produce more detailed console output
 --supplier-id-file=<path>         Path to file containing supplier ids to check. One ID per line.
+--excluded-supplier-ids=<esis>    Supplier IDs to be excluded.
```
risicle: Should probably mention that they're comma separated in case someone thinks there also an "id file" like the option above.

dmscripts/mark_definite_framework_results.py
agalamatis: Yes, `interested_supplier_ids = list(set(interested_supplier_ids) - set(excluded_supplier_ids))`, but this would require knowledge of sets.

scripts/mark-definite-framework-results.py
agalamatis: Indeed, I will update the README.md with the format required and the _secret_ location of the IDs.

dmscripts/mark_definite_framework_results.py
risicle: Well, we don't even need to convert them back to a list afterwards, and it's fine to use sets - we're all adults here. If we want to make it more explicit we can always use `set`'s named methods rather than the operators...

PR #370, opened 2019-04-30
Move get_bucket_name into helper module

katstevens:
No ticket - fix for broken export scripts.

- Moves `get_bucket_name` into a new helper module
- Adds `reports` to the list of approved bucket types

### DIFF
--- a/scripts/generate-supplier-user-csv.py
+++ b/scripts/generate-supplier-user-csv.py
@@ -32,10 +32,10 @@ import sys
 sys.path.insert(0, '.')
 
 from dmscripts.helpers.auth_helpers import get_auth_token
-from dmscripts.bulk_upload_documents import get_bucket_name
 from dmscripts.generate_supplier_user_csv import generate_csv_and_upload_to_s3
 from dmscripts.helpers import logging_helpers
 from dmscripts.helpers.logging_helpers import logging
+from dmscripts.helpers.s3_helpers import get_bucket_name
 from dmutils.env_helpers import get_api_endpoint_from_stage
 
 from docopt import docopt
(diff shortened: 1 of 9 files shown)


scripts/generate-supplier-user-csv.py
```diff
@@ -32,10 +32,10 @@
 sys.path.insert(0, '.')
 
 from dmscripts.helpers.auth_helpers import get_auth_token
-from dmscripts.bulk_upload_documents import get_bucket_name
 from dmscripts.generate_supplier_user_csv import generate_csv_and_upload_to_s3
 from dmscripts.helpers import logging_helpers
 from dmscripts.helpers.logging_helpers import logging
+from helpers.s3_helpers import get_bucket_name
```
lfdebrux: Should this be dmscripts.helpers?
```suggestion
from dmscripts.helpers.s3_helpers import get_bucket_name
```

scripts/upload-counterpart-agreements.py
```diff
@@ -42,10 +42,11 @@
 sys.path.insert(0, '.')
 
 from dmscripts.helpers.auth_helpers import get_auth_token
-from dmscripts.bulk_upload_documents import get_bucket_name, get_all_files_of_type
+from dmscripts.bulk_upload_documents import get_all_files_of_type
```
lfdebrux: While we're at it can we move this function as well?

katstevens: Why not. Will add a separate commit

scripts/generate-supplier-user-csv.py
katstevens: That's weird, Pycharm autosuggested it with no problems. Will fix

PR #371, opened 2019-05-01
add new generalized & refactored publish-draft-services.py script 

risicle:
~A draft because it's still missing tests, but script seems to work from my ad-hoc testing so take a look.~

https://trello.com/c/V2miZJ5G

Should be ready to review now, has a _reasonable_ level of testing - a few corner cases missed out though.

### DIFF
--- a/dmscripts/publish_draft_services.py
+++ b/dmscripts/publish_draft_services.py
@@ -1,0 +1,219 @@
+from itertools import chain
+import random
+import re
+from urllib.parse import urljoin
+
+import backoff
+
+import dmapiclient
+from dmscripts.helpers.framework_helpers import find_suppliers_on_framework, get_submitted_drafts
+from dmscripts.helpers.logging_helpers import get_logger
+from dmutils.s3 import S3ResponseError
+
+
+def _parse_document_url(url, framework_slug):
+    pattern = r'/{}/submissions/(\d+)/(\d+)-(.*)$'.format(re.escape(framework_slug))
+    match = re.search(pattern, url)
+    if not match:
+        raise ValueError(f"Could not parse document URL {url!r}")
+    return {
+        'supplier_id': match.group(1),
+        'draft_id': match.group(2),
+        'document_name': match.group(3),
+    }
+
+
+def _get_draft_document_path(parsed_document_url, framework_slug):
+    return '{framework_slug}/submissions/{supplier_id}/{draft_id}-{document_name}'.format(
+        framework_slug=framework_slug,
+        **parsed_document_url)
+
+
+def _get_live_document_path(parsed_document_url, framework_slug, service_id):
+    return '{framework_slug}/documents/{supplier_id}/{service_id}-{document_name}'.format(
+        framework_slug=framework_slug,
+        service_id=service_id,
+        **parsed_document_url)
+
+
+def _copy_document(draft_bucket, documents_bucket, draft_document_path, live_document_path, dry_run):
+    if not draft_bucket.path_exists(draft_document_path):
+        raise ValueError(f"Draft document {draft_document_path} does not exist in bucket {draft_bucket.bucket_name}")
+
+    message_suffix = f"{draft_bucket.bucket_name}:{draft_document_path} to " \
+        f"{documents_bucket.bucket_name}:{live_document_path}"
+
+    if dry_run:
+        get_logger().info("dry run: skipped copying %s", message_suffix)
+    else:
+        get_logger().info("copying %s", message_suffix)
+        documents_bucket.copy(
+            src_bucket=draft_bucket.bucket_name,
+            src_key=draft_document_path,
+            target_key=live_document_path,
+            acl='public-read',
+        )
+
+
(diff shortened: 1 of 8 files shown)


dmscripts/publish_draft_services.py
```diff
@@ -0,0 +1,219 @@
...
+        service_id = random.randint(55500000, 55599999)
+        get_logger().info(
+            "supplier %s: draft %s: dry run: generating random test service id: %s",
+            draft_service["supplierId"],
+            draft_service['id'],
+            service_id,
+        )
+    else:
+        try:
+            services = client.publish_draft_service(draft_service['id'], 'publish_draft_services.py')
```
katstevens: Should we make it clearer that `publish_draft_services.py` is the `updated_by` user? I know it's not a named kwarg on the client so maybe a comment, or sticking this in a suitably named variable...

dmscripts/publish_draft_services.py
```diff
@@ -0,0 +1,219 @@
...
+        documents_bucket.copy(
+            src_bucket=draft_bucket.bucket_name,
+            src_key=draft_document_path,
+            target_key=live_document_path,
+            acl='public-read',
+        )
+
+
+# this function exposed as importable - the intention is that a user of this module will wrap up a custom version
+# of this function and supply that to publish_draft_services as copy_documents_callable
```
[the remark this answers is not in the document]
risicle: In reality, probably adapted for the framework, but it's designed to be opaque to what it's being customized for.

scripts/publish-draft-services.py
```diff
@@ -0,0 +1,98 @@
...
+from dmscripts.helpers.auth_helpers import get_auth_token
+from dmscripts.helpers.logging_helpers import configure_logger
+from dmscripts.publish_draft_services import publish_draft_services, copy_draft_documents
+
+_document_keys_by_framework_family = {
+    "g-cloud": (
+        'pricingDocumentURL',
+        'serviceDefinitionDocumentURL',
+        'sfiaRateDocumentURL',
+        'termsAndConditionsDocumentURL',
```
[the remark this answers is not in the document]
risicle: But ... to where? And how will they be served?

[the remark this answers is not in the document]
katstevens: Exactly... needs clarification from product before we do anything

dmscripts/publish_draft_services.py
risicle: Sure.

risicle: ...but... hm... really I feel it should give space to put the *actual user* here which the script should probably take as an argument. But if it's being run from jenkins...? In theory we could try and extract the jenkins user name of the person that hit the button...

scripts/publish-draft-services.py
```diff
@@ -0,0 +1,98 @@
...
+    STAGE = arguments['<stage>']
+    api_url = get_api_endpoint_from_stage(STAGE)
+
+    client = dmapiclient.DataAPIClient(api_url, get_auth_token('api', STAGE))
+    DRAFT_BUCKET = arguments.get('<draft_bucket>') and S3(arguments['<draft_bucket>'])
+    DOCUMENTS_BUCKET = arguments.get('<documents_bucket>') and S3(arguments['<documents_bucket>'])
+    DRY_RUN = arguments['--dry-run']
+    FRAMEWORK_SLUG = arguments['<framework_slug>']
+    DRAFT_IDS_PATH = arguments['--draft-ids']
+    SKIP_DOCS_IF_PUBLISHED = arguments['--skip-docs-if-published']
```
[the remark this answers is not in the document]
risicle: We've got to leave *some* of the soul of kev in here...

scripts/publish-draft-services.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
risicle: Done

tests/test_publish_draft_services.py
```diff
@@ -0,0 +1,271 @@
...
+        mock_data_api_client,
+        "g-cloud-123",
+        DraftServiceStub(
+            id=379001,
+            framework_slug="g-cloud-123",
+            supplier_id="171400",
+            service_id="2229090909",
+            **{
+                "chinChopperURL": "https://gin.ger.br/ead/g-cloud-123/submissions/171400/379001-steak.pdf",
+                "beeoTeeTomURL": "https://gin.ger.br/ead/g-cloud-123/submissions/171400/379001-kidney.pdf",
```
katstevens: So this one should raise the ValueError, correct? Do we assert that?

risicle: Ooh well spotted - I realized this one was a tricky situation because the only side effect it has is a possible log message going out. _But_ the thing we really care about the script's behaviour in handling this exception is that it doesn't cause it to abort the run. So in essence the fact that the script continues and checks the following drafts _is_ the assertion.

katstevens: Ah I see - we raise if it's anything other than 'file is already here'. Cool 👍

PR #374, opened 2019-05-09
Get tokens from creds for performance platform stats script

katstevens:
Trello: https://trello.com/c/TZoo1F5B/480-performance-platform-jenkins-job-leaks-token

The token is currently retrieved from credentials when the Jenkins job is generated from the template and output in the Jenkins console (and the config!!). 

Rather than stick yet another var into the main Jenkins config (in this case there would be one for each framework!) we can get the script to decrypt it from either the Jenkins environment (if it's there) or the credentials repo, like we do for the API tokens.

For example:

```
docker run -e DM_DATA_API_TOKEN_PREVIEW digitalmarketplace/scripts scripts/my-amazing-script.py preview --notify-api-key=ABC1234
```

Can become (with a small tweak to the script):

```
docker run -e DM_DATA_API_TOKEN_PREVIEW -e NOTIFY_API_KEY digitalmarketplace/scripts scripts/my-amazing-script.py preview
```

I've also added a `--dry-run` flag to the performance stats script, so we can test it without messing up our G11 stats.

The Jenkins job needs updating to remove the token from the script call: https://github.com/alphagov/digitalmarketplace-jenkins/pull/215

### DIFF
--- a/dmscripts/helpers/auth_helpers.py
+++ b/dmscripts/helpers/auth_helpers.py
@@ -34,7 +34,12 @@ def get_auth_token(api, stage):
 
 
 def _get_nested_value(creds_json, env_var_path_list):
-    # Handle nested values in the credentials json
+    """
+    Handle nested values in the credentials json.
+    :param creds_json: e.g. {"performance_platform_bearer_tokens": {"g-cloud": "myToken"}}
+    :param env_var_path_list: e.g. ["performance_platform_bearer_tokens", "g-cloud"]
+    :return: e.g. "myToken"
+    """
     if len(env_var_path_list) > 1:
         return _get_nested_value(creds_json.get(env_var_path_list[0], {}), env_var_path_list[1:])
     return creds_json.get(env_var_path_list[0])
--- a/dmscripts/helpers/auth_helpers.py
+++ b/dmscripts/helpers/auth_helpers.py
@@ -5,6 +5,15 @@ import yaml
 DEV_ALIASES = ('dev', 'development', 'local')
 
 
+def _decrypt_yaml_file_with_sops(credentials_repo_path, credentials_filename):
+    creds_output = subprocess.check_output([
+        "{}/sops-wrapper".format(credentials_repo_path),
+        "-d",
+        "{}/{}".format(credentials_repo_path, credentials_filename)
+    ])
+    return yaml.safe_load(creds_output)
+
+
 def get_auth_token(api, stage):
     if stage.lower() in DEV_ALIASES:
         return 'myToken'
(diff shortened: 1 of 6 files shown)


dmscripts/helpers/auth_helpers.py
```diff
@@ -17,12 +26,43 @@ def get_auth_token(api, stage):
...
-        ])
-        auth_tokens = yaml.safe_load(creds)[api]['auth_tokens']
+        creds_json = _decrypt_yaml_file_with_sops(DM_CREDENTIALS_REPO, "vars/{}.yaml".format(stage))
+        auth_tokens = creds_json[api]['auth_tokens']
         auth_token = next(token for token in auth_tokens if token.startswith(token_prefix))
 
     return auth_token
+
+
+def _get_nested_value(creds_json, env_var_path_list):
```
benvand: Some tests or a docstring for this would give good impression of the sort of args it's expecting/ will work with.

katstevens: @benvand Pushed a commit 👍 (Also moved the top level script into the `framework-applications` folder, to live with the other framework lifecycle scripts)

lfdebrux: @katstevens I think the latest commit copied the file rather than moving it?

dmscripts/helpers/auth_helpers.py
```diff
@@ -17,12 +26,48 @@ def get_auth_token(api, stage):
...
+    """
+    Handle nested values in the credentials json.
+    :param creds_json: e.g. {"performance_platform_bearer_tokens": {"g-cloud": "myToken"}}
+    :param env_var_path_list: e.g. ["performance_platform_bearer_tokens", "g-cloud"]
+    :return: e.g. "myToken"
+    """
+    if len(env_var_path_list) > 1:
+        return _get_nested_value(creds_json.get(env_var_path_list[0], {}), env_var_path_list[1:])
```
risicle: We'll make a functional programmer of you yet... λ

dmscripts/send_stats_to_performance_platform.py
```diff
@@ -139,10 +139,14 @@ def send_by_stage_stats(stats, timestamp_string, period, pp_bearer, pp_service):
         "period": period
     } for stage in processed_stats]
 
+    if dry_run:
+        print(f"Would send the following stage stats to Performance Platform API:")
```
risicle: There _is_ a configured `logger` in this module, y'know...

risicle:
Hold on so now we have two scripts called `send-stats-to-performance-platform.py `? One in `scripts` and one in `scripts/framework-applications/`?

dmscripts/helpers/auth_helpers.py
katstevens: 🤦‍♀ Will sort that out

dmscripts/send_stats_to_performance_platform.py
katstevens: There is indeed! The other `print` statement was wrong anyway - have pushed a commit to fix.

dmscripts/helpers/auth_helpers.py
katstevens: Let's rewrite the entire marketplace in Haskell.

katstevens:
Happy to create another card for the relative creds path issue (agree it's not ideal).

PR #379, opened 2019-05-21
notify_suppliers_of_brief_withdrawal.py: use get_email_addresses from helpers to process emails

risicle:
To sort out https://trello.com/c/rNsjCmNh

This is a minimal fix allow support of splitting on multiple-address separators.

Going further to give this script better ergonomics I think we should consider combining it with `notify_suppliers_of_awarded_briefs.py` as elaborated on in https://trello.com/c/wDzrqj9n

### DIFF
--- a/dmscripts/notify_suppliers_of_brief_withdrawal.py
+++ b/dmscripts/notify_suppliers_of_brief_withdrawal.py
@@ -1,11 +1,16 @@
-from dmutils.email.helpers import hash_string
+from itertools import chain
+
+from dmutils.email.helpers import get_email_addresses, hash_string
 
 from dmutils.env_helpers import get_web_url_from_stage
 
 
 def get_brief_response_emails(data_api_client, brief_id):
     responses = data_api_client.find_brief_responses_iter(brief_id=brief_id, status="submitted")
-    return [response["respondToEmailAddress"] for response in responses]
+    return list(chain.from_iterable(
+        get_email_addresses(response["respondToEmailAddress"])
+        for response in responses
+    ))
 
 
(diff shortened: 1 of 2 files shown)


dmscripts/notify_suppliers_of_brief_withdrawal.py
```diff
@@ -1,11 +1,16 @@
...
+
+from dmutils.email.helpers import get_email_addresses, hash_string
 
 from dmutils.env_helpers import get_web_url_from_stage
 
 
 def get_brief_response_emails(data_api_client, brief_id):
     responses = data_api_client.find_brief_responses_iter(brief_id=brief_id, status="submitted")
-    return [response["respondToEmailAddress"] for response in responses]
+    return list(chain.from_iterable(
```
lfdebrux: what's the advantage of using chain here?

risicle: What's the alternative? Other than nested `for`s which I find unclear.

lfdebrux: Does `[]` not work?

lfdebrux: Oh wait it needs to be flattened I get it

risicle: Not if the contents are an iterable - to be clear, here I'm having to flatten an iterable of iterables, that's what the `chain.from_iterable` is about.

lfdebrux: I guess you could put the generator expression inside the call arguments

PR #387, opened 2019-06-04
Add script for validating AWS CloudTrail logs

lfdebrux:
Part of work to add a CloudTrail log validating job to Jenkins (ticket [here](https://trello.com/c/Ijzuk0qY/508-log-file-validation-job)). The script requires the CloudTrail ARNS to be added to the credentials repo alphagov/digitalmarketplace-credentials#221.

Rather than putting the script straight in a job I thought I would make separate PR for it here, so I can add tests and watnots.

### DIFF
--- a/scripts/validate-cloudtrail-logs.sh
+++ b/scripts/validate-cloudtrail-logs.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+#
+# Usage: validate-cloudtrail-logs.sh [options] ACCOUNT START_FROM [UP_TO]
+#
+# Examples:
+#   validate-cloudtrail-logs.sh main '1 hour ago'
+#   validate-cloudtrail-logs.sh production '2019-04-11T13:13'
+#   validate-cloudtrail-logs.sh --verbose development '3 days ago' '2 days ago'
+#
+# Arguments:
+#   ACCOUNT     The AWS account to validate logs for (e.g. main, development, backups).
+#   START_FROM  A date/time to start validating logs from, can be a date relative to now.
+#   UP_TO       A date/time to start validating logs from, can be a date relative to now. Defaults to now.
+#
+# Options:
+#   --config=FILE  Path to file containing details of trails (default: $DM_CREDENTIALS_REPO/jenkins-vars/cloudtrail.yaml).
+#   -n, --dry-run  Output information on command but do not run.
+#   -q, --quiet    Suppress output of log validation information.
+#   -v, --verbose  Print detailed script debug statements.
+#   -h, --help     Show this message.
+
+set -eo pipefail
+
+source docopts.sh
+
+help=$(docopt_get_help_string "$0")
+eval "$(docopts -h "$help" : "$@")"
+
+VERBOSE="${VERBOSE:-$verbose}"
+$VERBOSE && set -x
+
+[ -n "$config" ] && CONFIG_FILE="$config" || CONFIG_FILE="$DM_CREDENTIALS_REPO/jenkins-vars/cloudtrail.yaml"
+[ -f "$CONFIG_FILE" ] || { echo "error: config file '$CONFIG_FILE' does not exist"; exit 1; }
+
+[ -n "$UP_TO" ] || UP_TO="now"
+START_TIME="$(date --date="$START_FROM" --utc --iso-8601=minutes)" || exit 1
+END_TIME="$(date --date="$UP_TO" --utc --iso-8601=minutes)" || exit 1
+
+export ACCOUNT
+account="$("$DM_CREDENTIALS_REPO"/sops-wrapper -d --extract '["cloudtrail_validate_logs_roles"]' "$CONFIG_FILE" | yq 'map(select(.account == env.ACCOUNT))[0]')"
+
+AWS_PROFILE="$(echo "$account" | yq -r '.profile.name')"
+
+COMMAND="aws cloudtrail validate-logs"
+
+COMMAND+=" --trail-arn $(echo "$account" | yq -r '.trail_arn')"
+COMMAND+=" --start-time $START_TIME"
+[ -n "$END_TIME" ] && COMMAND+=" --end-time $END_TIME"
+
+if ! $QUIET
+then
+  COMMAND+=" --verbose"
+fi
+
+export AWS_PROFILE
+echo AWS_PROFILE="$AWS_PROFILE" $COMMAND
+$dry_run && exit
+
+# run command, echo output to terminal, exit with non-zero status if any logs are invalid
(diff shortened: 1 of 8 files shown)


scripts/validate-cloudtrail-logs.sh
```diff
@@ -0,0 +1,45 @@
...
+
+if $VERBOSE
+then
+  COMMAND+=" --verbose"
+fi
+
+export AWS_PROFILE
+echo AWS_PROFILE="$AWS_PROFILE" $COMMAND
+$dry_run && exit
+$COMMAND
```
agalamatis: Can we add a condition to also notify on in `$COMMAND | grep INVALID | wc -l`. If it is zero, we are good, if it is greater than zero echo the grep result to be posted to slack and exit with non-zero.

agalamatis:
Can we add a condition to also notify on in `$COMMAND | grep INVALID | wc -l`. If it is zero, we are good, if it is greater than zero echo the grep result to be posted to slack and exit with non-zero.

scripts/validate-cloudtrail-logs.sh
lfdebrux: Right, I foolishly assumed that the exit code would be non-zero if validation failed...

lfdebrux: You wouldn't happen to know if it's possible to cause a failed validation in someway?

agalamatis: My bash skills are very rusty, but let me try:
`IS_INVALID = $($COMMAND | grep INVALID)`
`if $(echo $IS_INVALID  | wc -l) > 0`
`then $echo $IS_INVALID | exit 1`
`fi`

agalamatis: The echo in then statement will give us the trail logs/digests that failed to validate, and the exit will trigger the error, which would send the alert to slack (if I understood the jenkins code correctly).

lfdebrux: I've made an attempt based on grep and tee, what do you think?

scripts/validate-cloudtrail-logs.sh
```diff
@@ -0,0 +1,47 @@
...
+##?
+##? Options:
+##?   --config=FILE  Path to file containing details of trails (default: $DM_CREDENTIALS_REPO/jenkins-vars/cloudtrail.yaml)
+##?   -v, --verbose
+##?   -n, --dry-run
+##?   -h, --help
+
+set -eo pipefail
+
+help=$(grep "^##?" "$0" | cut -c 5-)
```
benvand: Can't we use the pattern we've used elsewhere?

https://github.com/alphagov/digitalmarketplace-aws/blob/master/scripts/map-route.sh#L4-L3

scripts/validate-cloudtrail-logs.sh
```diff
@@ -0,0 +1,47 @@
...
+
+[ -n "$config" ] && CONFIG_FILE="$config" || CONFIG_FILE="$DM_CREDENTIALS_REPO/jenkins-vars/cloudtrail.yaml"
+[ -f "$CONFIG_FILE" ] || { echo "error: config file '$CONFIG_FILE' does not exist"; exit 64; }
+
+[ -n "$UP_TO" ] || UP_TO="now"
+START_TIME="$(date --date="$START_FROM" --utc --iso-8601=minutes)" || exit 64
+END_TIME="$(date --date="$UP_TO" --utc --iso-8601=minutes)" || exit 64
+
+export ACCOUNT
+account="$("$DM_CREDENTIALS_REPO"/sops-wrapper -d --extract '["cloudtrail_validate_logs_roles"]' "$CONFIG_FILE" | yq 'map(select(.account == env.ACCOUNT))[0]')"
```
benvand: So if we added these perms to the jenkins profile instance and the infrastructure/ user tolese we could forego all this sops decryption

scripts/validate-cloudtrail-logs.sh
```diff
@@ -0,0 +1,47 @@
...
+eval "$(docopts -h "$help" : "$@")"
+
+VERBOSE="${verbose:-$VERBOSE}"
+$VERBOSE && set -x
+
+[ -n "$config" ] && CONFIG_FILE="$config" || CONFIG_FILE="$DM_CREDENTIALS_REPO/jenkins-vars/cloudtrail.yaml"
+[ -f "$CONFIG_FILE" ] || { echo "error: config file '$CONFIG_FILE' does not exist"; exit 64; }
+
+[ -n "$UP_TO" ] || UP_TO="now"
+START_TIME="$(date --date="$START_FROM" --utc --iso-8601=minutes)" || exit 64
```
benvand: Don't know what an exit 64 is or why we're doing it. And I don't think it's common knowledge

benvand: (could probably use a comment)

benvand:
For me it _felt_ like it took me a bit too long to get to grips with all of what this script and the tests are doing considering how simple the command it needs to build and run is.

I'd suggest seeing if it can be simplified, seeing if we have existing patterns that could be used, or more heavily commenting each step. But you may disagree.

If we're sure all the additions here are worthwhile (`docopt`?) then I imagine there should be a recommendation to add it to all our scripts and change the way we do things everywhere rather than just introducing it in a single script. Otherwise the addition of what could be a useful tool, rather than making everyones lives more convenient, leads to higher complexity and more paradigms to learn before they can change the codebase.

scripts/validate-cloudtrail-logs.sh
lfdebrux: As well as the profile details [the file we're decrypting](https://github.com/alphagov/digitalmarketplace-credentials/pull/221/files#diff-1f8bec20b12553ad99d4b3ed1da377ce) contains the trail arn, which the aws cli needs. I guess we could forgo that file if we rely on the name scheme being consistent.

That being said, I don't know how it would be possible to have the permissions for validating logs from another account in the Jenkins instance profile.

scripts/validate-cloudtrail-logs.sh
lfdebrux: According to docopts it's linux convention https://man.cx/sysexits(3)

I didn't know about it before though so if you didn't either we should probably ignore it, rather than confuse people

scripts/validate-cloudtrail-logs.sh
lfdebrux: We could, but I found using docopts a lot easier, I think it has a lot of advantages over the alternatives, especially given that we use docopt in all our python scripts.

scripts/validate-cloudtrail-logs.sh
lfdebrux: Removed this

PR #395, opened 2019-06-12
Fix script to warn when framework is not open

lfdebrux:
Send a warning message to the log instead of crashing.

### DIFF
--- a/dmscripts/generate_supplier_user_csv.py
+++ b/dmscripts/generate_supplier_user_csv.py
@@ -1,5 +1,6 @@
 import csv
 from dmutils.s3 import S3ResponseError
+from requests.exceptions import HTTPError
 
 from dmscripts.models.writecsv import csv_path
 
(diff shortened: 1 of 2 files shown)


dmscripts/generate_supplier_user_csv.py
```diff
@@ -1,15 +1,24 @@
...
     framework = data_api_client.get_framework(framework_slug).get("frameworks")
 
-    supplier_rows = data_api_client.export_suppliers(framework_slug).get('suppliers', [])
-    if not supplier_rows:
-        logger.warn('No supplier data found for framework {}'.format(framework_slug))
+    try:
+        supplier_rows = data_api_client.export_suppliers(framework_slug).get("suppliers", [])
+    except HTTPError as e:
+        if e.response.status_code == 400:
+            logger.warn(f"Framework '{framework_slug}' is closed, no supplier data is available")
```
katstevens: The 400s are if it's an nonexistent framework slug or the framework isn't open yet. I'd go with `Framework '{framework_slug}' is not yet open, no supplier data is available`

lfdebrux: Oh yes that makes more sense

katstevens: ('Closed' implies that applications have closed)

PR #401, opened 2019-06-21
Refactor notify-successful-suppliers-for-framework script

lfdebrux:
I made a few changes based on what I would liked to have had when running this recently:

- Reorganise usage string
- Add dry-run flag
- Allow filtering supplier ids
- Add functional tests
- Hash email addresses in logs to prevent leaking PII

### DIFF
--- a/scripts/framework-applications/notify-successful-suppliers-for-framework.py
+++ b/scripts/framework-applications/notify-successful-suppliers-for-framework.py
@@ -1,17 +1,28 @@
 #!/usr/bin/env python
-"""Email suppliers who have at least one successful lot entry on the given framework.
-Uses the Notify API to inform suppliers of success result.
+"""
+Email suppliers who have at least one successful lot entry on the given framework.
+
+Uses the Notify API to inform suppliers of success result. This script *should not* resend emails.
 
 Usage:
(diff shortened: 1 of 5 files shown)


scripts/framework-applications/notify-successful-suppliers-for-framework.py
```diff
@@ -1,49 +1,76 @@
...
     context_data = context_helper.get_users_personalisations()
 
+    prefix = "[Dry Run] " if DRY_RUN else ""
+
     for user_email, personalisation in context_data.items():
-        logger.info(user_email)
-        mail_client.send_email(user_email, GOVUK_NOTIFY_TEMPLATE_ID, personalisation, allow_resend=False)
+        logger.info(f"{prefix}Sending email to supplier user '{hash_string(user_email)}'")
+
+        if DRY_RUN:
```
agalamatis: I'd rather nest the try/except blocks in `if not DRY_RUN` and not have a `continue`.

lfdebrux: I find it hard to read it that way which is why I usually prefer to do it this way. Do you feel very strongly about doing it they way you said?

PR #404, opened 2019-06-25
One week reminder

agalamatis:
trello: https://trello.com/c/0fG8ahIA/546-reminder-email-for-suppliers-with-incomplete-applications

Can I sit with someone opinionated about the structure of script, before I start patching and parametrising tests? and can give me an intro to notify templates?

### DIFF
--- a/dmscripts/notify_suppliers_with_incomplete_applications.py
+++ b/dmscripts/notify_suppliers_with_incomplete_applications.py
@@ -1,0 +1,71 @@
+from dmapiclient import DataAPIClient
+from dmscripts.helpers.auth_helpers import get_auth_token
+from dmscripts.helpers.email_helpers import scripts_notify_client
+from dmscripts.helpers.logging_helpers import configure_logger, logging
+from dmutils.email.exceptions import EmailError
+from dmutils.email.helpers import hash_string
+from dmutils.env_helpers import get_api_endpoint_from_stage
+
+
+NOTIFY_TEMPLATE_ID = '25c7c763-fe51-418f-8e51-130c391edc35'
+
+MESSAGES = [
+    '* confirm your company details\n',
+    '* finish your supplier declaration\n',
+    '* mark at least one of your services as complete\n',
+]
+
+
+def send_notification(mail_client, message, email, supplier_id, dry_run):
+    prefix = "[Dry Run] " if dry_run else ""
+    mail_client.logger.info(
+        f"{prefix}Sending email to supplier '{supplier_id}' "
+        f"user '{hash_string(email)}'"
+    )
+    if not dry_run:
+        try:
+            mail_client.send_email(
+                email, NOTIFY_TEMPLATE_ID, {'message': message}, allow_resend=False
+            )
+        except EmailError as e:
+            mail_client.logger.error(
+                f"Error sending email to supplier '{supplier_id}' "
+                f"user '{hash_string(email)}': {e}"
+            )
+            return 1
+    return 0
+
+
+def notify_suppliers_with_incomplete_applications(framework, stage, notify_api_key, dry_run):
+    logger = configure_logger({"dmapiclient": logging.INFO})
+    mail_client = scripts_notify_client(notify_api_key, logger=logger)
+    data_api_client = DataAPIClient(
+        base_url=get_api_endpoint_from_stage(stage), auth_token=get_auth_token('api', stage)
+    )
+    error_count = 0
+
+    for sf in data_api_client.find_framework_suppliers_iter(framework):
+        message = ''
+        if not sf.get('applicationCompanyDetailsConfirmed', None):
+            message += MESSAGES[0]
+        if not sf.get('declaration', {'status': None}).get('status', None) == 'complete':
+            message += MESSAGES[1]
+        has_submitted_at_least_one_service = False
+        for service in data_api_client.find_draft_services_iter(sf['supplierId'], framework=framework):
+            if service.get('status') == 'submitted':
+                has_submitted_at_least_one_service = True
+                break
+        if not has_submitted_at_least_one_service:
+            message += MESSAGES[2]
+
+        if message:
+            primary_email = sf.get('declaration', {'primaryContactEmail': None}).get('primaryContactEmail', None)
+            if primary_email:
+                error_count += send_notification(mail_client, message, primary_email, sf['supplierId'], dry_run)
+            for user in data_api_client.find_users(supplier_id=sf['supplierId']).get('users', []):
+                if user['active']:
+                    error_count += send_notification(
+                        mail_client, message, user['emailAddress'], sf['supplierId'], dry_run
+                    )
+
+    return error_count
--- a/tests/test_notify_suppliers_with_incomplete_applications.py
+++ b/tests/test_notify_suppliers_with_incomplete_applications.py
@@ -1,0 +1,136 @@
+from logging import Logger
+import mock
+import pytest
+
+from dmutils.email import DMNotifyClient
+from dmscripts.notify_suppliers_with_incomplete_applications import (
+    notify_suppliers_with_incomplete_applications,
+    MESSAGES,
+)
+
+
+FRAMEWORK_SUPPLIERS_TEST_CASES = [
+    [
+        # input
+        [{
+            'supplierId': 0,
+            'applicationCompanyDetailsConfirmed': True,
+            'declaration': {'status': 'complete', 'primaryContactEmail': 'abc@example.com'}
+        }],
+        ['abc@example.com', ],  # output, list of email addresses
+        '',  # output, expected message fragment
+    ],
+    [
+        # input
+        [{
+            'supplierId': 1,
+            'applicationCompanyDetailsConfirmed': True,
+            'declaration': {'status': 'started', 'primaryContactEmail': 'abc@example.com'}
+        }],
+        ['abc@example.com', ],  # output, list of email addresses
+        MESSAGES[1],  # output, expected message fragment
+    ],
+    [
+        # input
+        [{
+            'supplierId': 2,
+            'applicationCompanyDetailsConfirmed': True,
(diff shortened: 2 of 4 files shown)


dmscripts/notify_suppliers_with_incomplete_applications.py
```diff
@@ -0,0 +1,59 @@
...
+from dmscripts.helpers.supplier_data_helpers import AppliedToFrameworkSupplierContextForNotify
+from dmutils.email.exceptions import EmailError
+from dmutils.email.helpers import hash_string
+from dmutils.env_helpers import get_api_endpoint_from_stage
+
+
+def notify_suppliers_with_incomplete_applications(doc_opt_arguments):
+    dry_run = doc_opt_arguments['--dry-run']
+    framework = doc_opt_arguments['<framework>']
+    stage = doc_opt_arguments['<stage>']
```
risicle: We usually don't let details about the docopt information leak into the `dmscripts` modules. I think one of the reasons for the existence of the separate modules in the  `scripts/` dir is that _they_ are in charge of turning command line parameters into standard python args.

dmscripts/notify_suppliers_with_incomplete_applications.py
```diff
@@ -0,0 +1,59 @@
...
+
+    logger = logging_helpers.configure_logger({"dmapiclient": logging.INFO})
+    mail_client = scripts_notify_client(doc_opt_arguments['<notify_api_key>'], logger=logger)
```
risicle: Again with the `logger` and `mail_client` - these tend to get set up in the "script" and functions in `dmscripts` are intended to be a more programmatic interface which can just be handed a `logger` or `notify_client` object.

dmscripts/notify_suppliers_with_incomplete_applications.py
```diff
@@ -0,0 +1,56 @@
...
+        if not sf.get('applicationCompanyDetailsConfirmed', None):
+            message += 'confirm your company details\n'
+        if not (sf.get('declaration', {})):
+            message += 'finish your supplier declaration\n'
+        if not (data_api_client.find_draft_services(sf['supplierId'], framework=framework).get('services', [])):
+            message += 'mark at least one of your services as complete\n'
+
+        if len(message) > 0:
+            for user in data_api_client.find_users(supplier_id=sf['supplierId']).get('users', []):
+                if user['active'] and not dry_run:
```
katstevens: Do we need `and not dry_run` here if we're checking it below?

dmscripts/notify_suppliers_with_incomplete_applications.py
```diff
@@ -0,0 +1,56 @@
...
+    error_count = 0
+
+    for sf in data_api_client.find_framework_suppliers_iter(framework):
+        message = ''
+        if not sf.get('applicationCompanyDetailsConfirmed', None):
+            message += 'confirm your company details\n'
+        if not (sf.get('declaration', {})):
+            message += 'finish your supplier declaration\n'
+        if not (data_api_client.find_draft_services(sf['supplierId'], framework=framework).get('services', [])):
+            message += 'mark at least one of your services as complete\n'
```
risicle: We don't get to put markdown in Notify personalizations do we? If so, we could make these bullet points, otherwise it'll have to do.

dmscripts/notify_suppliers_with_incomplete_applications.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+    prefix = "[Dry Run] " if dry_run else ""
+    error_count = 0
+
+    for sf in data_api_client.find_framework_suppliers_iter(framework):
+        message = ''
+        if not sf.get('applicationCompanyDetailsConfirmed', None):
+            message += 'confirm your company details\n'
+        if not (sf.get('declaration', {})):
```
risicle: I think there you need to be checking for `declaration["status"] == "complete"`?

dmscripts/notify_suppliers_with_incomplete_applications.py
risicle: Yeah actually it'd be better not to have it so that we get the log message - presumably this is the point of the `prefix` variable.

dmscripts/notify_suppliers_with_incomplete_applications.py
```diff
@@ -0,0 +1,61 @@
...
+    )
+    error_count = 0
+
+    for sf in data_api_client.find_framework_suppliers_iter(framework):
+        message = ''
+        if not sf.get('applicationCompanyDetailsConfirmed', None):
+            message += '* confirm your company details\n'
+        if not sf.get('declaration', {'status': None}).get('status', None) == 'complete':
+            message += '* finish your supplier declaration\n'
+        if not data_api_client.find_draft_services(sf['supplierId'], framework=framework).get('services', []):
```
katstevens: We might need to be a bit stricter on this clause. We want to check if the supplier has at least 1 draft service with status `submitted` - if they don't, we should include the line below in the email.

dmscripts/notify_suppliers_with_incomplete_applications.py
```diff
@@ -0,0 +1,71 @@
...
+    error_count = 0
+
+    for sf in data_api_client.find_framework_suppliers_iter(framework):
+        message = ''
+        if not sf.get('applicationCompanyDetailsConfirmed', None):
+            message += MESSAGES[0]
+        if not sf.get('declaration', {'status': None}).get('status', None) == 'completed':
+            message += MESSAGES[1]
+        has_submitted_at_least_one_service = False
+        for service in data_api_client.find_draft_services(sf['supplierId'], framework=framework).get('services', []):
```
katstevens: We should use `find_draft_services_iter` in case there are more than 100 results. Also it'll mean we won't need the `.get('services', [])`. Something like this would do:

```
for draft_service in data_api_client.find_draft_services_iter(sf['supplierId'], framework=framework):
   if draft_service['status'] == 'submitted':
      has_submitted_at_least_one_service = True
      break
```

(I have no idea why the `list_draft_services` endpoint doesn't allow filtering by draft service status, it would make a bunch of scripts much quicker. Lost in the tech debt board, I guess...)

tests/test_notify_suppliers_with_incomplete_applications.py
```diff
@@ -0,0 +1,218 @@
...
+    draft_services_case, framework_supplier_case, users_case, expected_mails, expected_message
+):
+    """
+    Test if the correct number of email are sent and with the correct message.
+    """
+    mail_client_mock = mail_client_constructor_mock.return_value = mock.Mock(spec=DMNotifyClient)
+    mail_client_mock.logger = mock.Mock(spec=Logger)
+    data_api_client_mock().find_draft_services.return_value = draft_services_case
+    data_api_client_mock().find_framework_suppliers_iter.return_value = framework_supplier_case
+    data_api_client_mock().find_users.return_value = users_case
```
katstevens: Not essential for this script but we usually assert the `data_api_client_mock` calls as well as the notify client mocks (if we were posting/patching I'd say it was more important that we assert).

tests/test_notify_suppliers_with_incomplete_applications.py
```diff
@@ -0,0 +1,218 @@
...
+                    'emailAddress': 'abc@example.com'
+                },
+                {
+                    'active': True,
+                    'emailAddress': 'abc@example.com'
+                },
+            ]
+        },
+        4,
+        '',
```
katstevens: So to check my understanding, I take it this tuple will eventually have the following scenarios, where emails are sent to each supplier's primary contact, plus the one active user:

- supplier has done 0/3 steps 
- supplier has done 1/3 steps (confirmed company details)
- supplier has done 2/3 steps (confirmed company details and submitted declaration)
- supplier has done 2/3 steps (confirmed company details and completed a service)

And for a supplier who has done 3/3 steps, we will check that 0 emails have been sent?

tests/test_notify_suppliers_with_incomplete_applications.py
```diff
@@ -0,0 +1,218 @@
...
+        ],
+        {
+            'users': [
+                {
+                    'active': False,
+                    'emailAddress': 'abc@example.com'
+                },
+                {
+                    'active': True,
+                    'emailAddress': 'abc@example.com'
```
katstevens: Let's make the active and inactive users have different emails so we can assert it's been sent to the right one 😸

tests/test_notify_suppliers_with_incomplete_applications.py
agalamatis: That was a temporary insput to verify that mock/parametrize worked

tests/test_notify_suppliers_with_incomplete_applications.py
agalamatis: I will keep it in mind for the future.

agalamatis:
notify template here: https://www.notifications.service.gov.uk/services/95316ff0-e555-462d-a6e7-95d26fbfd091/templates/25c7c763-fe51-418f-8e51-130c391edc35

tests/test_notify_suppliers_with_incomplete_applications.py
```diff
@@ -0,0 +1,166 @@
...
+    [
+        # input
+        [{
+            'supplierId': 4,
+            'applicationCompanyDetailsConfirmed': None,
+            'declaration': {'status': 'complete', 'primaryContactEmail': 'abc@example.com'}
+        }],
+        ['abc@example.com', ],  # output, list of email addresses
+        MESSAGES[0],  # output, expected message fragment
+    ],
```
katstevens: We can probably remove these two cases with `'applicationCompanyDetailsConfirmed': None`  - suppliers cannot complete (or even start!) their declaration until they have confirmed their company details.

katstevens:
All right! This is looking pretty good (I'd have listed out the test scenarios differently, but this format is OK).  Squash the commits and I'll give it a 👍

PR #405, opened 2019-06-25
Use central supplier details for framework agreement PDFs

katstevens:
https://trello.com/c/skEefuVb/893-regenerate-fwas-for-suppliers

Spoiler: I already used this script to regenerate the PDFs needed for the support request (see ticket).

We don't seem to use Country in the PDF at all so I left that out.

No tests yet - comments on the general approach v welcome.

The corresponding PR for the agreements repo: https://github.com/alphagov/digitalmarketplace-agreements/pull/6

### DIFF
--- a/dmscripts/export_framework_applicant_details.py
+++ b/dmscripts/export_framework_applicant_details.py
@@ -146,6 +146,7 @@ def get_csv_rows(
     framework_lot_slugs,
     count_statuses=("submitted", "failed",),
     dry_run=False,
+    include_central_supplier_details=False,
 ):
     """
     :param count_statuses:  tuple of draft service statuses that should be
@@ -185,7 +186,9 @@ def get_csv_rows(
     logger.info(f"found {len(records)} supplier records to process")
 
     rows_iter = (
-        _create_row(framework_slug, record, count_statuses, framework_lot_slugs, dry_run)
+        _create_row(
+            framework_slug, record, count_statuses, framework_lot_slugs, dry_run, include_central_supplier_details
+        )
         for record in records
     )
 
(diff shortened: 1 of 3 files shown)


dmscripts/export_framework_applicant_details.py
```diff
@@ -212,8 +215,11 @@ def _format_field(field_name, field_value):
     return field_name, field_value
 
 
-def _create_row(framework_slug, record, count_statuses, framework_lot_slugs, dry_run=False):
+def _create_row(
+    framework_slug, record, count_statuses, framework_lot_slugs, dry_run=False, include_central_supplier_details=False
+):
     """
+    Fetch supplier data from central details, contact information and declaration
     :param dry_run:     if True return row without declaration information
```
lfdebrux: Could we have a line explaining what the new `include_central_supplier_details` param does?

dmscripts/export_framework_applicant_details.py
```diff
@@ -238,6 +244,23 @@ def _create_row(framework_slug, record, count_statuses, framework_lot_slugs, dry
...
+                "supplier_registration_number",
+                record["supplier"].get(
+                    'companiesHouseNumber',
+                    record["supplier"].get('otherCompanyRegistrationNumber', "")
+                )
+            ),
+            ("supplier_contact_address1", record["supplier"]["contactInformation"][0]['address1']),
+            ("supplier_contact_city", record["supplier"]["contactInformation"][0]['city']),
+            ("supplier_contact_postcode", record["supplier"]["contactInformation"][0]['postcode']),
+        ))
```
lfdebrux: I personally feel this would be more readable as a dict than as an tuple of tuples, feel free to disagree though.

katstevens: Broadly agree. Originally it was in the existing tuple above, and I moved it out. Happy to convert both to a dict...

lfdebrux: That would be 🔥

katstevens: Also realised that `get_csv_rows` also returns the field headers, which I... haven't updated. Going to add those now.

katstevens:
Going to merge this for now and add a tech-debt ticket for tests.

PR #406, opened 2019-06-26
556-include-other-registration-number-in-official-supplier-details-csv

benvand:
* Add field to csv export
* Add field to tests expecting this value

https://trello.com/c/schDS1vB/556-include-other-registration-number-in-official-supplier-details-csv

REQUIRES https://github.com/alphagov/digitalmarketplace-api/pull/945

### DIFF
--- a/dmscripts/generate_supplier_user_csv.py
+++ b/dmscripts/generate_supplier_user_csv.py
@@ -26,6 +26,7 @@ def generate_supplier_csv(framework_slug, data_api_client, logger):
         "supplier_organisation_size",
         "duns_number",
         "companies_house_number",
+        "other_company_registration_number",
         "registered_name",
         "declaration_status",
         "application_status",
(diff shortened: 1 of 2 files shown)


dmscripts/generate_supplier_user_csv.py
```diff
@@ -32,6 +32,7 @@ def generate_supplier_csv(framework_slug, data_api_client, logger):
         "application_result",
         "framework_agreement",
         "variations_agreed",
+        "other_company_registration_number",
```
katstevens: Could we put it next to `companies_house_number` so it's easy to see on the CSV which one the supplier has?

benvand: Yeah, I did think about this but was concerned that if Dave or Ash have macros working off this spreadsheet they might get misaligned, what do you think?

PR #411, opened 2019-07-03
Move script to framework-applications folder and include Supplier ID

katstevens:
The result of the script is no longer published publicly, but it's very useful when tallying stats of the framework just before it goes live. 

This also means we can include the Supplier ID which is a useful reference for any queries from CCS.

TODO: mention this script in the manual as a useful tool.

### DIFF
--- a/scripts/framework-applications/generate-successful-suppliers-for-framework-csv.py
+++ b/scripts/framework-applications/generate-successful-suppliers-for-framework-csv.py
@@ -2,7 +2,9 @@
 # -*- coding: utf-8 -*-
 """
 Generates a CSV listing all suppliers who have had services published to a framework, and the lots that they were
-successful on. This report is useful for tallying the results of the framework during standstill (i.e. just before the
+successful on, outputs to the current directory.
+
+This report is useful for tallying the results of the framework during standstill (i.e. just before the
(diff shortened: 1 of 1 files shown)


scripts/framework-applications/generate-successful-suppliers-for-framework-csv.py
```diff
@@ -1,15 +1,18 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 """
-Generate a CSV listing all suppliers who were successfully awarded onto a framework with the lots that they were
-successful on. This CSV is published as an update communication through the Digital Marketplace when the framework
-is awarded (i.e. when standstill ends).
+Generates a CSV listing all suppliers who have had services published to a framework, and the lots that they were
+successful on. This report is useful for tallying the results of the framework during standstill (i.e. just before the
+framework goes live).
+
```
lfdebrux: We should add a line to the docstring on where the data goes, or make it so that it outputs to standard out.

katstevens: Good idea. Don't want to faff with this script too much so have added info to the docstring.

PR #412, opened 2019-07-05
bad words script: hopefully make this a bit faster through reuse of compiled regexes

risicle:
Very short term improvement for https://trello.com/c/5IGQchTX/

The number of terms we use is probably greater than the re module's cache size, causing continual `re` re-compiling. 

No tests! :balloon:  But appears to work.

### DIFF
--- a/dmscripts/scan_g_cloud_services_for_bad_words.py
+++ b/dmscripts/scan_g_cloud_services_for_bad_words.py
@@ -68,9 +68,28 @@ def output_bad_words(
     writer.writerow(row)
 
 
+LOG_SUPPLIER_EVERY_N = LOG_SERVICE_EVERY_N = 10
+
+
 def check_services_with_bad_words(
     output_file_path, framework_slug, client, suppliers, bad_words, questions_to_check, logger, scan_drafts
 ):
+    supplier_count = 0
+
+    def count_supplier():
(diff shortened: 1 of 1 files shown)


dmscripts/scan_g_cloud_services_for_bad_words.py
```diff
@@ -122,7 +122,13 @@ def scan_services_for_bad_words(
...
     )
 
-    check_services_with_bad_words(
-        output_file_path, framework_slug, client, suppliers, bad_words, questions_to_check, logger, scan_drafts
+    return check_services_with_bad_words(
+        output_file_path,
+        framework_slug,
+        client,
+        suppliers,
+        tuple((bad_word, re.compile(rf"\b{re.escape(bad_word)}\b")) for bad_word in bad_words),
```
lfdebrux: Rather than changing the signature of the function can we compile the regexes inside it?

risicle: Only trying to give `scan_services_for_bad_words` more of a reason to exist.

risicle:
I've now also added logging of every 100th service or supplier to give the user some idea of progress.

risicle:
I hate you, flake8

PR #414, opened 2019-07-10
Add logging statements with hash of email address

lfdebrux:


dmscripts/upload_counterpart_agreements.py
```diff
@@ -83,9 +84,12 @@ def upload_counterpart_file(
                         "framework_name": framework["name"],
                         "supplier_name": supplier_name,
                     }, allow_resend=True)
+                    logger.info(f"NOTIFY: sent email to supplier '{supplier_id}' user {hash_string(notify_email)}")
```
katstevens: Just FYI there could potentially be a LOT of emails for the larger suppliers. It might be easier to spot errors if we just log the error path, not the happy path.

lfdebrux: Errors should be prefixed with ERROR by the logger; but I'm happy to take the line out

lfdebrux: I changed it to a debug statement so it will only come up if we set the script logging level to DEBUG

katstevens: Perfect 👍

PR #423, opened 2019-08-06
Improve get-model-data robustness

katstevens:
https://trello.com/c/WUPHdgSG/1007-export-data-failure-json-decode-error

One of the reports (brief-responses.csv) failed, stopping the entire script. This doesn't seem ideal.

This PR:
-  moves hardcoded lists of frameworks (that will soon need to be updated) to the top with the rest of the 'config'
- splits out the huge top-level script code into two functions
- adds a `try...except` around these functions to catch errors and continue on to the next 'config'
- updates the G-Cloud framework iterations we're exporting to slightly more interesting ones (G8 and G9 data is not changing any time soon!)

Out of scope for this PR:
- untangling the mess that is the testable functions for this script (i.e. moving my new functions into the testable modules and then testing them... maybe one day 😿 )
- figuring out why G-Cloud frameworks are passed through as string but DOS is a tuple (??)
- figuring out what 'configs' (i.e. individual reports) in this script are still needed or not

Tested against local env and everything seems to run OK.

### DIFF
--- a/scripts/get-model-data.py
+++ b/scripts/get-model-data.py
@@ -460,6 +460,86 @@ CONFIGS = [
     },
 ]
 
+
+def query_data_from_config(config, logger, limit, client, output_dir):
+    data = {}
+    if 'base_model' in config:
+        required_keys = list(config['keys']) + list(config.get('assign_json_subfields', {}).keys())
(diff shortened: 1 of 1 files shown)


katstevens:
@agalamatis Done! https://trello.com/c/DeH1vvTu/981-unravel-get-model-data-tests

PR #432, opened 2019-09-11
New DOS opportunities export script

katstevens:
https://trello.com/c/N3yUHHOe/92-make-dos-outcome-csv-framework-iteration-agnostic

- Splits out the public DOS opportunities CSV export from the mega `get-model-data` export (that exports a subset of model data to a private folder for internal use).
- Also adds tests & `--dry-run` functionality 🎉 

Currently, the Jenkins job doing the overnight export is also responsible for uploading this file to S3. This responsibility is now self contained within the script. (Updating the Jenkins job is next on my list - I won't merge this until that's ready).

This change should make the DOS opportunity export framework iteration agnostic - the file will be uploaded to the S3 bucket for latest live DOS framework. 

Having to fetch the brief responses for each brief via the API is a bit time consuming, but without changes to the API views I think it's the simplest way to get the data. This is a downside of splitting this out from the existing mega-export (which relies on `pandas` to join data from pre-existing csv files), but I think the benefits of visibility/maintainability make up for it. I could potentially add in `ThreadPool` (as per some of our other scripts) if people are keen, not sure how much it would help.

There are a few minor TODOs that aren't blockers.

### DIFF
--- a/dmscripts/export_dos_opportunities.py
+++ b/dmscripts/export_dos_opportunities.py
@@ -1,0 +1,91 @@
+import csv
+from dmscripts.models.process_rules import remove_username_from_email_address, format_datetime_string_as_date
+
+# This URL is framework agnostic
+PUBLIC_BRIEF_URL = "https://digitalmarketplace.service.gov.uk/digital-outcomes-and-specialists/opportunities/{}"
+
+DOS_OPPORTUNITY_HEADERS = [
+    "ID", "Opportunity", "Link", "Framework", "Category", "Specialist",
+    "Organisation Name", "Buyer Domain", "Location Of The Work",
+    "Published At", "Open For", "Expected Contract Length", "Applications from SMEs",
+    "Applications from Large Organisations", "Total Organisations", "Status", "Winning supplier",
+    "Size of supplier", "Contract amount", "Contract start date"
+]
+
+
+def _build_row(brief, brief_responses):
+    winner = None
+    applications_from_sme_suppliers = 0
+    applications_from_large_suppliers = 0
+
+    for brief_response in brief_responses:
+        if brief_response['supplierOrganisationSize'] == 'large':
+            applications_from_large_suppliers += 1
+        else:
+            applications_from_sme_suppliers += 1
+
+        if brief_response['status'] == 'awarded':
+            winner = brief_response
+
+    return [
+        brief['id'],
+        brief['title'],
+        PUBLIC_BRIEF_URL.format(brief['id']),
+        brief['frameworkSlug'],
+        brief['lotSlug'],
+        brief.get('specialist', ""),
+        brief['organisation'],
+        remove_username_from_email_address(brief['users'][0]['emailAddress']),
+        brief['location'],
+        format_datetime_string_as_date(brief['publishedAt']),
+        brief.get('requirementsLength', '2 weeks'),  # only briefs on the specialists lot include 'requirementsLength'
+        brief.get('contractLength', ''),
+        applications_from_sme_suppliers,
+        applications_from_large_suppliers,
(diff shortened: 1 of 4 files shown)


dmscripts/export_dos_opportunities.py
```diff
@@ -0,0 +1,90 @@
...
+def _build_row(brief, brief_responses):
+    winner = None
+    applications_from_sme_suppliers = 0
+    applications_from_large_suppliers = 0
+
+    for brief_response in brief_responses:
+        if brief_response['supplierOrganisationSize'] == 'large':
+            applications_from_large_suppliers += 1
+        else:
+            applications_from_sme_suppliers += 1
```
lfdebrux: This is super fussy but could we have a blank line after this one to make it clear the next conditional block is unrelated?

dmscripts/export_dos_opportunities.py
```diff
@@ -0,0 +1,90 @@
...
+        brief['title'],
+        PUBLIC_BRIEF_URL.format(brief['id']),
+        brief['frameworkSlug'],
+        brief['lotSlug'],
+        brief.get('specialist', ""),
+        brief['organisation'],
+        remove_username_from_email_address(brief['users'][0]['emailAddress']),
+        brief['location'],
+        format_datetime_string_as_date(brief['publishedAt']),
+        brief.get('requirementsLength', '2 weeks'),
```
lfdebrux: This seems a bit magical... is this documented anywhere? If not it might be good to.

dmscripts/export_dos_opportunities.py
```diff
@@ -0,0 +1,90 @@
...
+        if not dry_run:
+            # Save file
+            bucket.save(remote_key_name, source_file, acl='public-read', download_filename=download_name)
+
+        logger.info("{}UPLOAD: {} to {}::{}".format(
+            '[Dry-run]' if dry_run else '',
+            file_path,
+            bucket.bucket_name,
+            download_name
+        ))
```
lfdebrux: Can we have this log statement before the upload statement? I feel that would make the origin of any errors/tracebacks clearer

scripts/export-dos-opportunities.py
```diff
@@ -0,0 +1,77 @@
...
+    if not os.path.exists(OUTPUT_DIR):
+        logger.info("Creating {} directory".format(OUTPUT_DIR))
+        os.makedirs(OUTPUT_DIR)
+
+    client = DataAPIClient(get_api_endpoint_from_stage(STAGE), get_auth_token('api', STAGE))
+
+    latest_framework_slug = get_latest_dos_framework(client)
+
+    DOWNLOAD_FILE_NAME = 'opportunity-data.csv'
+    file_path = "{}/{}".format(OUTPUT_DIR, DOWNLOAD_FILE_NAME)
```
lfdebrux: Should probably use `os.path.join`... or pathlib ;)

scripts/export-dos-opportunities.py
```diff
@@ -0,0 +1,77 @@
...
+    bucket_name = 'digital-marketplace-{}-{}'.format(STAGE, STAGE)
+    key_name = '{}/communications/data/{}'.format(latest_framework_slug, DOWNLOAD_FILE_NAME)
```
lfdebrux: Just thought of something... this doesn't look right. I think the bucket should be `digitalmarketplace-communications-{STAGE}-{STAGE}`?

katstevens: Whoops, yes good spot 🕵

scripts/export-dos-opportunities.py
katstevens: I'll stick with `os.path` as it's used above 😸

PR #439, opened 2019-09-23
DOS4 opportunity alerts run for all frameworks with live briefs

katstevens:
https://trello.com/c/58q6lRKF/33-update-dos4-regular-email-jobs

Basically a total rewrite of the DOS opportunities email script, that we send via Mailchimp. Previously the script took a `framework` arg, and could only be run for that framework. That meant that on the day after go-live, we'd have to manually run a catchup script to 'catch' the previous iteration's briefs that had been created before the switchover. 

The docstring has a thorough description of what's meant to be happening but the tl;dr is that the script should get all live briefs, sort them into framework/lot lists, and create/send the campaigns for each list. 

Comments very welcome. 

Changes to the equivalent Jenkins job: https://github.com/alphagov/digitalmarketplace-jenkins/pull/268

### DIFF
--- a/dmscripts/send_dos_opportunities_email.py
+++ b/dmscripts/send_dos_opportunities_email.py
@@ -158,59 +158,51 @@ def main(data_api_client, mailchimp_client, number_of_days, framework_override, 
     # If specific framework script arg supplied, ignore other frameworks
     if framework_override:
         if not live_briefs_by_framework.get(framework_override):
-            logger.info(
-                "No new briefs found for {} in the last {} day(s)".format(framework_override, number_of_days)
-            )
+            logger.info(f"No new briefs found for {framework_override} in the last {number_of_days} day(s)")
             return True
         live_briefs_by_framework = {framework_override: live_briefs_by_framework.get(framework_override)}
 
     # If no briefs found, exit early
     if not live_briefs_by_framework:
-        logger.info(
-            "No new briefs found for DOS frameworks in the last {} day(s)".format(number_of_days),
-            extra={"number_of_days": number_of_days}
-        )
+        logger.info(f"No new briefs found for DOS frameworks in the last {number_of_days} day(s)")
         return True
 
     for framework_slug in live_briefs_by_framework.keys():
         for lot_slug, live_briefs in live_briefs_by_framework[framework_slug].items():
 
             if lot_slug_override and lot_slug != lot_slug_override:
-                logger.info("Skipping campaign for '{0}' lot on {1}".format(lot_slug, framework_slug))
+                logger.info(f"Skipping campaign for '{lot_slug}' lot on {framework_slug}")
                 continue
 
-            logger.info("{0} new briefs found for '{1}' lot on {2}".format(len(live_briefs), lot_slug, framework_slug))
+            logger.info(f"{len(live_briefs)} new briefs found for '{lot_slug}' lot on {framework_slug}")
 
             # Get the list_id for this lot/framework combination (or use list ID override if supplied as script arg)
             list_id = list_id_override or MAILCHIMP_LIST_IDS[framework_slug][lot_slug]
 
             # Create a new campaign for today's emails
-            logger.info("Creating campaign for '{0}' lot on {1}".format(lot_slug, framework_slug))
+            logger.info(f"Creating campaign for '{lot_slug}' lot on {framework_slug}")
             campaign_data = get_campaign_data(LOT_NAMES[lot_slug], list_id, live_briefs[0]['frameworkName'])
             campaign_id = mailchimp_client.create_campaign(campaign_data)
             if not campaign_id:
-                logger.warning("Unable to create campaign for '{0}' lot on {1}".format(lot_slug, framework_slug))
+                logger.warning(f"Unable to create campaign for '{lot_slug}' lot on {framework_slug}")
                 continue
 
             # Build the email content
             content_data = get_html_content(live_briefs, number_of_days)
             logger.info(
-                "Setting campaign data for '{0}' framework, '{1}' lot and '{2}' campaign id".format(
-                    framework_slug, lot_slug, campaign_id
-                )
+                f"Setting campaign data for '{framework_slug}' framework, "
+                f"'{lot_slug}' lot and '{campaign_id}' campaign id"
             )
             if not mailchimp_client.set_campaign_content(campaign_id, content_data):
-                logger.warning("Unable to set campaign data for campaign id '{0}'".format(campaign_id))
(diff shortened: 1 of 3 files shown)


dmscripts/send_dos_opportunities_email.py
```diff
@@ -8,12 +8,40 @@
...
+
+LOT_NAMES = {
+    "digital-specialists": "Digital specialists",
+    "digital-outcomes": "Digital outcomes",
+    "user-research-participants": "User research participants",
+}
+
+
+def get_live_briefs_between_two_dates(data_api_client, start_date, end_date):
     """Get all briefs for a lot which were published between 2 dates."""
```
lfdebrux: We should update the docstring to reflect the change in behaviour

```suggestion
    """Get all briefs which were published between 2 dates."""
```

dmscripts/send_dos_opportunities_email.py
```diff
@@ -67,47 +95,122 @@ def get_html_content(briefs, number_of_days):
...
+
+
+def main(data_api_client, mailchimp_client, number_of_days, framework_override, list_id_override, lot_slug_override):
+    """
+    Default behaviour is to fetch any briefs (regardless of framework/lot) published in the last N days.
+
+    The script groups briefs by framework/lot, and for each combination, looks up the Mailchimp list ID, and creates &
+    sends a new campaign.
+    - If a framework_override is supplied, just find briefs for that framework
+    - If a lot_override is supplied, just find briefs for that lot
```
lfdebrux: Should this be `lot_slug_override`?

```suggestion
    - If a lot_slug_override is supplied, just find briefs for that lot
```

dmscripts/send_dos_opportunities_email.py
```diff
@@ -67,47 +95,122 @@ def get_html_content(briefs, number_of_days):
...
+            else:
+                briefs_by_fw_and_lot[brief_fw_slug][brief_lot_slug] = [brief]
+        else:
+            briefs_by_fw_and_lot[brief_fw_slug] = {
+                brief_lot_slug: [brief]
+            }
+    return briefs_by_fw_and_lot
+
+
+def main(data_api_client, mailchimp_client, number_of_days, framework_override, list_id_override, lot_slug_override):
```
lfdebrux: I'd suggest making some of these arguments keyword only

```suggestion
def main(data_api_client, mailchimp_client, number_of_days, *, framework_override, list_id_override, lot_slug_override):
```

dmscripts/send_dos_opportunities_email.py
```diff
@@ -67,47 +95,122 @@ def get_html_content(briefs, number_of_days):
...
+    if framework_override:
+        if not live_briefs_by_framework.get(framework_override):
+            logger.info(
+                "No new briefs found for {} in the last {} day(s)".format(framework_override, number_of_days)
+            )
+            return True
+        live_briefs_by_framework = {framework_override: live_briefs_by_framework.get(framework_override)}
+
+    # If no briefs found, exit early
+    if not live_briefs_by_framework.keys():
```
lfdebrux: I think `not live_briefs_by_framework` would have the same value as `not live_briefs_by_framework.keys()`

```suggestion
    if not live_briefs_by_framework:
```

dmscripts/send_dos_opportunities_email.py
```diff
@@ -67,47 +95,122 @@ def get_html_content(briefs, number_of_days):
...
-        "Sending campaign for '{0}' lot and '{1}' campaign id".format(lot_data["lot_slug"], campaign_id)
-    )
-    if not mailchimp_client.send_campaign(campaign_id):
-        return False
+
+    for framework_slug in live_briefs_by_framework.keys():
+        for lot_slug, live_briefs in live_briefs_by_framework[framework_slug].items():
+
+            if lot_slug_override and lot_slug != lot_slug_override:
+                logger.warning("Skipping campaign for '{0}' lot on {1}".format(lot_slug, framework_slug))
```
lfdebrux: I'm not sure this needs to be a warning? Could be a `logger.debug`?

scripts/send-dos-opportunities-email.py
```diff
@@ -1,38 +1,44 @@
...
-    we have set up a testing list with ID "096e52cebb"
-
-    If you only need to send opportunities for one lot rather than all of them, you can also do this via the command
-    line. Note, this may come in useful if the script was to fail halfway and you wish to continue from the lot which
-    failed.
+Default behaviour is to fetch any briefs (regardless of framework/lot) published in the last N days.
+
+The script groups briefs by framework/lot, and for each combination, looks up the Mailchimp list ID, and creates &
+sends a new campaign.
+- If a framework_override is supplied, just find briefs for that framework
```
lfdebrux:
```suggestion
- If `--framework_slug` is supplied, just find briefs for that framework
```

scripts/send-dos-opportunities-email.py
```diff
@@ -1,38 +1,44 @@
...
+- If a lot_override is supplied, just find briefs for that lot
```
lfdebrux:
```suggestion
- If `--lot_slug` is supplied, just find briefs for that lot
```

scripts/send-dos-opportunities-email.py
```diff
@@ -1,38 +1,44 @@
...
+- If a list_id_override is supplied, only send emails to that list ID regardless of lot/framework
```
lfdebrux:
```suggestion
- If `--list_id` is supplied, only send emails to that list ID regardless of lot/framework
```

dmscripts/send_dos_opportunities_email.py
jonodrew: Why do these braces have numbers in while the braces on 162 do not?

dmscripts/send_dos_opportunities_email.py
katstevens: Good catch - copied and pasted from the other docstring. Will fix

dmscripts/send_dos_opportunities_email.py
katstevens: I'll go with `.info`, it's probably useful to see on the job output

katstevens: Haha, probably because I refactored them on different days 😹Will make them consistent

dmscripts/send_dos_opportunities_email.py
katstevens: Good call. I held off on that originally because I "didn't want to change too much" 🤦‍♀ Anyway,  will add a commit

PR #440, opened 2019-09-25
Enable autoescaping on html templates

jonodrew:
This was identified as a risk by a recent IT health check. The change ensures that user input is escaped, preventing any nasties from applying XSS attacks.

[Ticket](https://trello.com/c/H2ubPYqz/306-enable-autoescape-on-jinja-templates)

### DIFF
--- a/dmscripts/helpers/html_helpers.py
+++ b/dmscripts/helpers/html_helpers.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 import io
-from jinja2 import Environment, StrictUndefined
+from jinja2 import Environment, StrictUndefined, select_autoescape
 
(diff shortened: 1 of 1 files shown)


lfdebrux:
Could you also add some info to the commit message as to why we're enabling StrictUndefined?

jonodrew:
> Could you also add some info to the commit message as to why we're enabling StrictUndefined?

Not really? It was there before...

lfdebrux:
> > Could you also add some info to the commit message as to why we're enabling StrictUndefined?
> 
> Not really? It was there before...

🤦‍♂ that's what I get for reviewing at 5pm

PR #442, opened 2019-10-01
class to compare DMp with D&B data

agalamatis:
trello: https://trello.com/c/zAgmgNP8/234-generate-supplier-csv-report-on-results-from-db-api


dmscripts/compare_dmp_and_dnb_data.py
```diff
@@ -0,0 +1,324 @@
...
+import numpy as np
+import requests
+import unicodecsv
+
+from dmapiclient import DataAPIClient
+from dmscripts.helpers.auth_helpers import get_auth_token
+from dmutils.env_helpers import get_api_endpoint_from_stage
+
+
+def levenshtein_ratio_and_distance(s, t, ratio_calc=False):
```
risicle: are we really going to have our own implementation of this instead of adding e.g. `python-levenshtein` or even `fuzzywuzzy` to the deps?

dmscripts/compare_dmp_and_dnb_data.py
```diff
@@ -0,0 +1,324 @@
...
+    def get_dnb_org_data(self, duns_number):
+        "return a dictionary with all the information D&B has on the company with that DUNS number"
+        response = requests.get(
+            f'{self.root}/V6.0/organizations?match=true&MatchTypeText=Basic&DUNSNumber={str(duns_number)}',
+            headers={
+                'Authorization': self.dnb_auth_token,
+            }
+        )
+        if response.json()['MatchResponse']['TransactionResult']['ResultID'] == 'CM000':
+            return response.json()['MatchResponse']['MatchResponseDetail']['MatchCandidate'][0]
```
risicle: Could we have a comment explaining what this condition represents?

dmscripts/compare_dmp_and_dnb_data.py
```diff
@@ -0,0 +1,324 @@
...
+        self.text = text
+
+
+class DNBDirectAPIClient(object):
+    "A client utilising the D&B direct API with methods used by the DMp API."
+
+    def __init__(self, username, password, stage='production'):
+        "initialisation with D&B credentials and optionally DMp stage"
+        self.stage = stage
+        self.data_api_client = None
```
risicle: Are we supposed to just assign this `data_api_client` to the instance after construction? Would be nicer as an `__init__` arg...

risicle: Oh no I see it tries to autogenerate itself when needed...

dmscripts/compare_dmp_and_dnb_data.py
```diff
@@ -0,0 +1,324 @@
...
+                    if csv_filename is not None:
+                        counter += 1
+                        print(duns_number)
+                        writer.writerow(row)  # TODO: extend with D&B data
+                        self.duns_number_compared.add(duns_number)  # add DUNS to set
+                        if counter >= self.max_dnb_api_calls_per_method:
+                            return
+                    else:
+                        print(row)  # TODO: log maybe?
+        # close file resource if necessary
```
risicle: ideally this would be done with a `finally` block or context manager, but good enough for a prototype script...

dmscripts/compare_dmp_and_dnb_data.py
agalamatis: > Oh no I see it tries to autogenerate itself when needed...

Yes, and in addition the client has methods that independent of our API, only concerned with the D&B API.

dmscripts/compare_dmp_and_dnb_data.py
agalamatis: > are we really going to have our own implementation of this instead of adding e.g. `python-levenshtein` or even `fuzzywuzzy` to the deps?

I had problems importing `fuzzywuzzy` and the only thing I needed from it was the metric, so I just added it as a function to be able to work on everything else.

risicle: But everything "just works" with pipi doesn't it?

dmscripts/compare_dmp_and_dnb_data.py
agalamatis: > Could we have a comment explaining what this ~condition represents~ `CM000` thing is about?

Comment explaining it and reference to the D&B documentation.

dmscripts/compare_dmp_and_dnb_data.py
agalamatis: > ideally this would be done with a `finally` block or context manager, but good enough for a prototype script...

Agreed, but as you said it is overkill for a prototype.

dmscripts/compare_dmp_and_dnb_data.py
```diff
@@ -0,0 +1,339 @@
...
+        for framework in frameworks:
+            dmp_response = self.get_dmp_supplier_data(framework=framework, from_declaration=from_declaration)
+            if from_declaration:
+                dmp_response = dmp_response.get('supplierFrameworks', [])
+            else:
+                dmp_response = dmp_response.get('suppliers', [])
+            for dmp_supplier in dmp_response:
+                self.compare_suppliers(
+                    dmp_supplier, from_declaration, export_dnb_data, export_dmp_data, counter, writer
+                )
```
agalamatis: @risicle I pulled part of the function into a new one, because flake was complaining. Works the same as before. Can you have a look?

risicle: Ugh.. I expect you're referring to the "complexity checks". I'm very sceptical about them. TBH I think the result it's guided you towards is harder to understand - a _many_-arg'd function which adds the uncertainty over which conditions it's going to be called in. An inner block of a larger function has no such uncertainly - execution can only reach that point by following the function's prologue.

:man_shrugging: 

It is permitted to bump the check's limit, yknow.

PR #443, opened 2019-10-03
DOS lot export scripts improvements

katstevens:
https://trello.com/c/bNJt7mYX/324-dos-lot-exports-are-not-in-the-right-format-and-have-extra-fields

Addresses some problems with the DOS export scripts:
- Fixes bug where `question.fields` was being called for question types that didn't have that attribute, and adds a couple of tests to check this (a `checkbox` type question previously worked fine, but a `boolean` type question would have failed)
- Adds DOS service fixtures to support the tests above. Mocking the content loader is a bit of a nightmare, so I thought we might as well use the real one, but feed it a tiny subset of our content to keep things speedy.
- Makes logging consistent across the four scripts
- Allows the user to specify an `output-dir` rather than hardcoding to 'output'
- Exports the actual expected filenames for each lot (i.e. don't include framework slug) so we don't have to rename them before uploading to S3
- Adds 'last updated' row to the end of each file. Currently we add this manually to the top of each file, but I'm not sure how useful it is, given last year we updated the files precisely 0 times after initially generating them! Rather than prepend the date in a horribly inefficient way, I thought I'd stick it on the end for now. A dev can always manually insert it at the top as well if it's required.

Not done due to time/complexity:
- optionally filtering out suppliers that 'failed' (i.e. `onFramework = False`)
- re-ordering/relabelling columns
- general speed improvements
- combining the 4 scripts into one mega-script (the dream)

### DIFF
--- a/scripts/export-dos-outcomes.py
+++ b/scripts/export-dos-outcomes.py
@@ -101,5 +101,5 @@ if __name__ == '__main__':
     write_csv_with_make_row(
         suppliers,
         make_row(capabilities, locations),
-        os.path.join(OUTPUT_DIR, "{}-outcomes.csv".format(FRAMEWORK_SLUG))
+        os.path.join(OUTPUT_DIR, "digital-outcomes-suppliers.csv".format(FRAMEWORK_SLUG))
     )
--- a/scripts/export-dos-outcomes.py
+++ b/scripts/export-dos-outcomes.py
@@ -101,5 +101,6 @@ if __name__ == '__main__':
     write_csv_with_make_row(
         suppliers,
         make_row(capabilities, locations),
-        os.path.join(OUTPUT_DIR, "digital-outcomes-suppliers.csv".format(FRAMEWORK_SLUG))
+        os.path.join(OUTPUT_DIR, "digital-outcomes-suppliers.csv".format(FRAMEWORK_SLUG)),
+        include_last_updated=True
     )
--- a/scripts/export-dos-outcomes.py
+++ b/scripts/export-dos-outcomes.py
@@ -5,9 +5,14 @@ For a DOS-type framework this will export details of all "digital-outcomes" serv
 of outcome the supplier provides and the locations they can provide them in.
 
 Usage:
-    scripts/export-dos-outcomes.py <stage> <framework_slug> <content_path> [--verbose]
+    scripts/export-dos-outcomes.py <stage> <framework_slug> <content_path> [options]
+
+Options:
+    -v --verbose                Print INFO level messages.
+    --output-dir=<output_dir>   Directory to write csv files to [default: output]
 """
 from multiprocessing.pool import ThreadPool
+import os
 import sys
 sys.path.insert(0, '.')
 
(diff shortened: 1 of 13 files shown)


scripts/export-dos-outcomes.py
```diff
@@ -77,10 +94,13 @@ def get_outcomes_locations(content_manifest):
...
 
+    logger.info(f"Finding suppliers for Digital Outcomes on {FRAMEWORK_SLUG}")
     suppliers = find_all_outcomes(client, map_impl=pool.imap)
 
+    logger.info(f"Building CSV for {len(suppliers)} Digital Outcomes suppliers")
     write_csv_with_make_row(
         suppliers,
         make_row(capabilities, locations),
-        "output/{}-outcomes.csv".format(FRAMEWORK_SLUG)
+        os.path.join(OUTPUT_DIR, "digital-outcomes-suppliers.csv".format(FRAMEWORK_SLUG)),
```
lfdebrux: There isn't a site for the format method to insert the framework slug in this string?

scripts/export-dos-participants.py
```diff
@@ -68,10 +80,13 @@ def inner(record):
...
 
+    logger.info(f'Finding User Research Participants suppliers for {FRAMEWORK_SLUG}')
     records = find_all_participants(client, map_impl=pool.imap)
 
+    logger.info(f"Building CSV for {len(records)} User Research Participants suppliers")
     write_csv_with_make_row(
         records,
         make_row(content_manifest),
-        "output/{}-user-research-participants.csv".format(FRAMEWORK_SLUG)
+        os.path.join(OUTPUT_DIR, "user-research-participants-suppliers.csv".format(FRAMEWORK_SLUG)),
```
lfdebrux: There isn't a site for the format method to insert the framework slug in this string?

scripts/export-dos-specialists.py
```diff
@@ -72,10 +85,13 @@ def get_specialist_roles(section):
...
 
+    logger.info(f"Finding Digital Specialists suppliers for {FRAMEWORK_SLUG}")
     suppliers = find_all_specialists(client, map_impl=pool.imap)
 
+    logger.info(f"Building CSV for {len(suppliers)} Digital Specialists suppliers")
     write_csv_with_make_row(
         suppliers,
         make_row(content_manifest),
-        "output/{}-specialists.csv".format(FRAMEWORK_SLUG)
+        os.path.join(OUTPUT_DIR, "digital-specialists-suppliers.csv".format(FRAMEWORK_SLUG)),
```
lfdebrux: There isn't a site for the format method to insert the framework slug in this string?

tests/helpers/test_csv_helpers.py
```diff
@@ -26,3 +27,57 @@ def test_read_csv():
         ['Company Name', 'Supplier Name', 'pass'],
         ['678', 'Supplier Name', 'PasS'],
     ]
+
+
+class TestMakeFieldsFromContentQuestions:
+
+    def setup(self):
+        content_loader = ContentLoader('tests/fixtures/content')
```
lfdebrux: This is an excellent idea 🔥

scripts/export-dos-outcomes.py
katstevens: Good spot. I thought I'd removed all those but that got left behind 😸

PR #444, opened 2019-10-07
Improving the Export data script 

katstevens:
https://trello.com/c/iwI4osQ7/1137-investigate-export-data-job-failures

We've had a few annoying failures on the overnight runs for this script. To try and make it a bit more robust, I've split out some functions, added `try...except`s, some tests and logging, and removed some unused bits of logic.

The main change is to move a heap of logic into a testable module, and split one large function into two (one to handle the db/query actions on the `pandas` dataframe, one for post-processing). 

Also adds DOS4 to the list of frameworks we want to do service exports for. 

The test coverage I've added isn't exhaustive by any means, but we've got to start somewhere.


tests/models/test_process_rules.py
```diff
@@ -1,38 +1,214 @@
...
+        }
+        client = mock.Mock(autospec=DataAPIClient)
+        expected_query_result = pandas.DataFrame([{'id': '1', 'result': 'awarded'}])
+        base_model.side_effect = [expected_query_result]
+
+        assert query_data_from_config(
+            config, logger=mock.ANY, limit=100, client=client, output_dir="data"
+        ).equals(expected_query_result)
+
+        # What this test is really concerned with - we call queries.base_model rather than queries.model
```
risicle: Do any of the tests care about what methods were called on `client`?

katstevens: Those calls are buried in `ModelTrawler` - `test_modeltrawler.py` has coverage there.

risicle: Sounds like a joy.

PR #446, opened 2019-10-15
Script to export Notify templates to CSV

katstevens:
https://trello.com/c/sHKgjXRC/1063-spike-manage-notify-template-and-reply-to-ids-outside-app-config

The goal of the ticket above was for us to manage our Notify templates better - having oversight of what templates we actually have seems like a quick win towards that goal! I rustled up this simple script to export the templates, and thought I might as well commit it.

The script includes an option to include the multi-line `body` field - if you're importing the CSV into a spreadsheet it makes it much harder to get an at-a-glance view of the templates, but it might be useful for content designers.

### DIFF
--- a/scripts/oneoff/export-notify-templates-to-csv.py
+++ b/scripts/oneoff/export-notify-templates-to-csv.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python
+
+"""
+Script to fetch templates from the Notify API and export them to CSV.
+See Notify's documentation: https://docs.notifications.service.gov.uk/python.html#get-all-templates
+
+Usage:
+    export-notify-templates-to-csv.py <notify_api_key> [options]
+
+Options:
+    --include-message-body      # Include multi-line message body field (makes the csv harder to read at-a-glance)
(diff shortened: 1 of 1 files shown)


scripts/oneoff/export-notify-templates-to-csv.py
```diff
@@ -0,0 +1,68 @@
...
+    path = os.path.join(output_dir, filename)
+
+    # Fetch templates
+    resp = client.get_all_templates()
+
+    with open(path, 'w') as f:
+        writer = csv.writer(f, delimiter=',', quotechar='"')
+        writer.writerow(headers)
+        for r in resp['templates']:
+            writer.writerow([r[header] for header in headers])
```
lfdebrux: You might be able to use `csv.DictWriter` instead of explicitly writing out the list comprehension here

katstevens: That's an excellent idea 😸 Can you tell I copied and pasted this bit of code from a throwaway Python shell...

PR #447, opened 2019-10-16
index_to_search_service: split iteration of briefs to index into two separate calls

risicle:
https://trello.com/c/d1jpHXSp

If https://github.com/alphagov/digitalmarketplace-api/pull/979 is merged this would allow the bulk of our briefs to be fetched in compressed responses, improving reliability.

### DIFF
--- a/tests/test_index_to_search_service.py
+++ b/tests/test_index_to_search_service.py
@@ -70,11 +70,11 @@ class TestIndexers:
         ]
 
     def test_service_indexer_request_items_calls_data_api_client_with_frameworks(self):
-        self.data_api_client.return_value.find_services_iter.return_value = ['service1', 'service2']
+        self.data_api_client.return_value.find_services_iter.return_value = iter(('service1', 'service2',))
         indexer = ServiceIndexer(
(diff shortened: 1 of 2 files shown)


tests/test_index_to_search_service.py
```diff
@@ -50,27 +50,40 @@ def test_brief_indexer_create_index_raises_on_other_api_errors(self):
             assert e.message == 'disaster'
 
     def test_brief_indexer_request_items_calls_data_api_client_with_frameworks(self):
-        self.data_api_client.return_value.find_briefs_iter.return_value = ['brief1', 'brief2']
+        self.data_api_client.return_value.find_briefs_iter.side_effect = lambda *args, **kwargs: {
+            "live,cancelled,unsuccessful,awarded,closed": iter(('brief1', 'brief2',)),
+            "withdrawn": iter(('brief3',)),
+        }[kwargs.get("status")]
```
katstevens: This is neat 👍

risicle:
Yeah the only actual difference it makes is it's no longer sending drafts to the search api. I don't *think* that should be an issue...

PR #455, opened 2019-11-18
including the primary email in the success notification

agalamatis:
trello: https://trello.com/c/nmGTH8zd/111-1-notify-successful-suppliers-for-framework-script-doesnt-email-primary-contact

Tested on g-cloud and got 11937 instead of 10619 emails, I will keep the data sources and dumps for the reviewer(s).

### DIFF
--- a/dmscripts/helpers/supplier_data_helpers.py
+++ b/dmscripts/helpers/supplier_data_helpers.py
@@ -97,7 +97,9 @@ class SuccessfulSupplierContextForNotify(SupplierFrameworkData):
                 output.update(self.get_user_personalisation(user, supplier_framework))
             primary_email_address = supplier_framework.get('declaration', {}).get('primaryContactEmail')
             if primary_email_address and primary_email_address not in output:
-                supplier_framework['users'].append({'email address': primary_email_address})
+                output.update(
+                    self.get_user_personalisation({'email address': primary_email_address}, supplier_framework)
+                )
         return output
 
     def get_lot_dict(self, supplier_framework):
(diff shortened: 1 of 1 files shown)


dmscripts/helpers/supplier_data_helpers.py
```diff
@@ -93,7 +93,13 @@ def get_users_personalisations(self):
                 self.logger.info(
                     'Building user personalisations for supplier {}'.format(supplier_framework['supplierId'])
                 )
-            for user in supplier_framework['users']:
+            primary_email_address = supplier_framework.get('declaration', {}).get('primaryContactEmail')
+            if primary_email_address:  # if the primary email is also a user it will be overwritten in the output dict
+                if 'users' in supplier_framework:
```
katstevens: I think there will always be a non-empty `users` in supplier_framework (otherwise we've got a problem for that supplier!)

Would something like this be simpler instead?

```
for user in supplier_framework['users']:
    output.update(self.get_user_personalisation(user, supplier_framework))

primary_email_address = supplier_framework.get('declaration', {}).get('primaryContactEmail')
if primary_email_address:
    output.update(self.get_user_personalisation(primary_email_address, supplier_framework))
```
As you say, if the primary email address is already in `'users'` then the personalisation is the same, and it doesn't matter if we overwrite or not.

agalamatis: OK, I can move the primary email after the rest of the users, but I'd rather skip overwriting, so the condition would be:
`if primary_email_address and primary_email_address not in output:`
What do you think?

katstevens: Yep, sounds good 👍

PR #456, opened 2019-11-21
rotate-api-tokens.sh: add new commands change-ft-account-passwords and sync-ft-account-passwords

risicle:
https://trello.com/c/AOxxcCSy

This was far more involved than I expected it to be, but doing it manually is annoying and error prone, so I wouldn't be beat.

`change-ft-account-passwords` will "detect" email/password combinations in a set of FT variables by looking for matching pairs of settings ending in `_email` and `_password`. It will then generate a new password for each account, separate accounts being determined by unique email address. A
github PR will then be created to apply these changes.

`sync-ft-account-passwords` will use the API to set the account passwords for any accounts found in a set of FT variables (detected in the same way).

Doing it this way allows separate accounts to be used for the different test variants, new classes of accounts could be added, or some accounts could be omitted entirely for certain variants and the scripts should continue to operate robustly & agnostically.

Coming soon - a basic jenkins job for driving this.

### DIFF
--- a/scripts/rotate-api-tokens.sh
+++ b/scripts/rotate-api-tokens.sh
@@ -130,10 +130,92 @@ function remove_old_callback_token() {
   echo "Done."
 }
 
+function change_ft_account_passwords() {
+  create_credentials_update_branch
+
+  echo "Changing Functional Test DMP account passwords for ${STAGE}."
+
+  admin_manager_pass_path=pass/digitalmarketplace.service.gov.uk/admin-manager
+  export STAGE_LOWER
+
+  if [[ $STAGE_LOWER = "production" ]] ; then
+    # if we're changing the admin manager password, we're also going to have to update it in $admin_manager_pass_path,
+    # so note the email address to look out for
+    admin_manager_email=$(./sops-wrapper -d $admin_manager_pass_path | sed -En -e "s/^login:\s+//p")
+  fi
+
+  # find all values of fields named _email which have an equivalent _password variable
+  for ACCOUNT_EMAIL in $(./sops-wrapper -d jenkins-vars/jenkins.yaml|yq -crM '[[(.smoulder_test_variables, .smoke_test_variables, .functional_test_variables)[env.STAGE_LOWER]|values|to_entries]|flatten|.[]|select(.key|endswith("_email")).value]|unique|.[]') ; do
+    export ACCOUNT_EMAIL
+    # passwords for a particular email address need to be common across test variants becuase they
+    # share the same instance
+    new_password=$(generate_api_token)
+
+    for VARIANT in smoulder_test_variables smoke_test_variables functional_test_variables ; do
+      export VARIANT
+      # for each field ending in _password whose equivalent _email-ending field has the value $ACCOUNT_EMAIL
+      for password_field in  $(./sops-wrapper -d jenkins-vars/jenkins.yaml|yq -crM '.[env.VARIANT][env.STAGE_LOWER] as $vars|$vars|to_entries|.[]|select((.key|endswith("_email")) and .value == env.ACCOUNT_EMAIL).key|sub("_email$"; "_password")|select($vars[.])') ; do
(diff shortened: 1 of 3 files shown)


scripts/set-dm-password-by-email.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
```
katstevens: Can we call this script something else? It kind of implies we are setting a new password by sending an email 😸 

How about `set-user-password-via-apiclient.py`?

scripts/set-dm-password-by-email.py
```diff
@@ -0,0 +1,41 @@
...
+    user = data_api_client.get_user(email_address=email_address)
+    if not user:
+        print(f"User {email_address!r} not found")
+        exit(2)
+
+    if not data_api_client.update_user_password(
+        user["users"]["id"],
+        environ["ACCOUNT_PASSWORD"],
+        "set-dm-password-by-email.py",
+    ):
```
katstevens: Is it worth adding any sort of warnings etc when running this in production? It's possible to set an arbitrary password for any user, not just our smoke/smoulder accounts.

scripts/set-dm-password-by-email.py
```diff
@@ -0,0 +1,41 @@
...
+sys.path.insert(0, '.')
+
+from os import environ
+from sys import exit
+
+from dmapiclient import DataAPIClient
+from dmscripts.helpers.auth_helpers import get_auth_token
+from dmutils.env_helpers import get_api_endpoint_from_stage
+
+if __name__ == "__main__":
```
katstevens: Would be super-nice to pull this `__main__` stuff into a function so we can move it & add tests later, but not a blocker

scripts/rotate-api-tokens.sh
```diff
@@ -130,10 +130,78 @@ function remove_old_callback_token() {
...
+    for VARIANT in smoulder_test_variables smoke_test_variables functional_test_variables ; do
+      export VARIANT
+      # for each field ending in _password whose equivalent _email-ending field has the value $ACCOUNT_EMAIL
+      for password_field in  $(./sops-wrapper -d jenkins-vars/jenkins.yaml|yq -crM '.[env.VARIANT][env.STAGE_LOWER] as $vars|$vars|to_entries|.[]|select((.key|endswith("_email")) and .value == env.ACCOUNT_EMAIL).key|sub("_email$"; "_password")|select($vars[.])') ; do
+        ./sops-wrapper --set "[\"${VARIANT}\"][\"${STAGE_LOWER}\"][\"${password_field}\"] \"${new_password}\"" jenkins-vars/jenkins.yaml
+      done
+    done
+  done
+
+  commit_and_create_github_pr "Change functional test DMP account passwords" "## Summary\r\nChanges all functional test/smoke test/smoulder test DMP account passwords for ${STAGE}."
```
katstevens: A thought: if we want to reduce the amount of time between merging the new creds and syncing them, should we include this in the PR description? "Be ready to do the next step of the pipeline (syncing) when you merge this PR"

risicle: Quite possibly yes

scripts/set-dm-password-by-email.py
risicle: oh... yeah I guess it does. But "via apiclient" is a bit obvious. How else could it? What if it were `set-user-password-by-email-address.py`?

scripts/set-dm-password-by-email.py
risicle: Yeeeah I kinda wanted this script to not exist at all really - originally tried to include the source in the bash script as a one-liner.

risicle:
I've now done a full run of this script via jenkins for preview and it worked.

risicle:
I've run this for all 3 environments now and it works.

risicle:
Yeah I guess so, we already have some limited wrappers for sops in `-utils`

PR #459, opened 2019-12-30
add notify_suppliers_of_new_fw_cq_answers script

risicle:
Originally https://trello.com/c/xbqxyfNV/ but has rather morphed into a separate script.

We can start off running this script when asked by CCS. It could be extended to run as a periodic job if this was found to be too much hassle, possibly detecting when to fire based on the etag of the most recent clarification question answers document found in the S3 bucket.

### DIFF
--- a/dmscripts/notify_suppliers_of_new_fw_cq_answers.py
+++ b/dmscripts/notify_suppliers_of_new_fw_cq_answers.py
@@ -1,0 +1,83 @@
+from logging import Logger
+from typing import Optional
+from uuid import UUID, uuid4
+
+from dmapiclient import DataAPIClient
+
+from dmutils.email import DMNotifyClient
+from dmutils.email.exceptions import EmailError
+from dmutils.email.helpers import hash_string
+from dmutils.env_helpers import get_web_url_from_stage
+
+
+def notify_suppliers_of_new_fw_cq_answers(
+    data_api_client: DataAPIClient,
+    notify_client: DMNotifyClient,
+    notify_template_id: str,
+    framework_slug: str,
+    stage: str,
+    dry_run: bool,
+    logger: Logger,
+    run_id: Optional[UUID] = None,
+) -> int:
+    run_is_new = bool(run_id)
+    run_id = run_id or uuid4()
+    logger.info(f"{'Starting' if run_is_new else 'Resuming'} run id {{run_id}}", extra={"run_id": str(run_id)})
+
+    framework = data_api_client.get_framework(framework_slug)["frameworks"]
+
+    if framework["status"] != "open":
+        raise ValueError(f"Framework {framework_slug!r} is not open (status is {framework['status']!r})")
+
+    failure_count = 0
+
+    for supplier_framework in data_api_client.find_framework_suppliers_iter(framework_slug):
+        for user in data_api_client.find_users_iter(supplier_id=supplier_framework["supplierId"]):
+            if user["active"]:
+                # generating ref separately so we can exclude certain parameters from the personalization dict
+                notify_ref = notify_client.get_reference(
+                    user["emailAddress"],
+                    notify_template_id,
+                    {
+                        "framework_slug": framework["slug"],
+                        "run_id": str(run_id),
+                    },
+                )
+                if dry_run:
+                    if notify_client.has_been_sent(notify_ref):
+                        logger.debug(
+                            "[DRY RUN] Would NOT send notification to {email_hash} (already sent)",
+                            extra={"email_hash": hash_string(user["emailAddress"])},
+                        )
+                    else:
+                        logger.info(
+                            "[DRY RUN] Would send notification to {email_hash}",
+                            extra={"email_hash": hash_string(user["emailAddress"])},
+                        )
+                else:
+                    try:
+                        notify_client.send_email(
+                            user["emailAddress"],
+                            notify_template_id,
+                            {
+                                "framework_name": framework["name"],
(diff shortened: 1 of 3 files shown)


dmscripts/notify_suppliers_of_new_fw_cq_answers.py
```diff
@@ -0,0 +1,81 @@
...
+    framework = data_api_client.get_framework(framework_slug)["frameworks"]
+
+    if framework["status"] != "open":
+        raise ValueError(f"Framework {framework_slug!r} is not open (status is {framework['status']!r})")
+
+    failure_count = 0
+
+    for supplier_framework in data_api_client.find_framework_suppliers_iter(framework_slug):
+        for user in data_api_client.find_users_iter(supplier_id=supplier_framework["supplierId"]):
+            if user["active"]:
```
katstevens: I think I am going to create a tech debt card to allow filtering users by `active` status in the API - it seems strange that we don't have this already?

dmscripts/notify_suppliers_of_new_fw_cq_answers.py
```diff
@@ -0,0 +1,81 @@
...
+                notify_ref = notify_client.get_reference(
+                    user["emailAddress"],
+                    notify_template_id,
+                    {
+                        "framework_slug": framework["slug"],
+                        "run_id": str(run_id),
+                    },
+                )
+                if dry_run:
+                    if notify_client.has_been_sent(notify_ref):
```
katstevens: Checking my understanding: this line would return true if we'd made a previous script attempt that contained a failure, and we're doing a `--dry-run` attempt first before we re-run for real?

dmscripts/notify_suppliers_of_new_fw_cq_answers.py
```diff
@@ -0,0 +1,81 @@
...
+                else:
+                    try:
+                        notify_client.send_email(
+                            user["emailAddress"],
+                            notify_template_id,
+                            {
+                                "framework_name": framework["name"],
+                                "updates_url":
+                                    f"{get_web_url_from_stage(stage)}/suppliers/frameworks/{framework['slug']}/updates",
+                                "clarification_questions_open": framework["clarificationQuestionsOpen"],
```
katstevens: Is the plan here to use this context var to display optional content in the template? If so we might need to map to `yes`/`no` strings, as we will probably want to show `The deadline to ask clarification new questions has passed` if `clarification_questions_open` is False, and I think the Notify `((var??content))` syntax needs 'yes' or 'no' (I might be wrong).

The existing template

dmscripts/notify_suppliers_of_new_fw_cq_answers.py
risicle: Yeah, so, the `has_been_sent` check is usually done _inside_ the `send_email` method, so we don't have to call it ourselves in the normal case. The only reason we have to call it ourselves is to be able to emit the correct log message for `dry_run` mode.

dmscripts/notify_suppliers_of_new_fw_cq_answers.py
risicle: I really left it open how we pass parameters to the template - what's here is little more than a sketch.

katstevens: OK. I'm happy to get a new template sorted out so we can start testing this and figure out what the Jenkins job will look like.

katstevens: I've added the new template now. I've called the optional var `clarification_questions_closed` so that we can do something like this:

```
clarification_questions_closed = "yes" if framework["clarificationQuestionsOpen"] else "no"
```

risicle: Huzzah. Cool. Will update.

dmscripts/notify_suppliers_of_new_fw_cq_answers.py
```diff
@@ -0,0 +1,83 @@
...
+                    try:
+                        notify_client.send_email(
+                            user["emailAddress"],
+                            notify_template_id,
+                            {
+                                "framework_name": framework["name"],
+                                "updates_url":
+                                    f"{get_web_url_from_stage(stage)}/suppliers/frameworks/{framework['slug']}/updates",
+                                "clarification_questions_closed": (
+                                    "no" if framework["clarificationQuestionsOpen"] else "yes"
```
katstevens: 👍

dmscripts/notify_suppliers_of_new_fw_cq_answers.py
katstevens: (Now done - https://trello.com/c/YTJrDGmH/1210-allow-filtering-by-active-status-in-api-listusers-view)

scripts/notify-suppliers-of-new-fw-cq-answers.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
```
katstevens: Ah - can we make this file executable, to be consistent with our other scripts?

risicle: ah crap

risicle: ok fixed

katstevens: Great - merge away!

PR #465, opened 2020-02-04
 export_framework_results_reasons: fail more obviously when we get unexpected validation errors

risicle:
To supplement https://github.com/alphagov/digitalmarketplace-frameworks/pull/603 and https://trello.com/c/kbTzccXb

We're adding extra constraints to the assessment schemas, constraints which in normal use we shouldn't expect to be broken. But if they are broken for some reason, we'll end up with a validation error that doesn't look as we expect it to, leading to problems when we try to run the "reasons" script.

This change just makes such errors less puzzling if they occur and adds a test for these cases.

### DIFF
--- a/dmscripts/export_framework_results_reasons.py
+++ b/dmscripts/export_framework_results_reasons.py
@@ -16,10 +16,17 @@ DRAFT_STATUSES = [
 ]
 
 
+def get_error_key(error):
+    try:
+        return error.path[0]
+    except (AttributeError, TypeError, IndexError):
+        raise ValueError(f"Unexpected validation error: {error!r}")
+
+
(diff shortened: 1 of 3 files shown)


dmscripts/export_framework_results_reasons.py
```diff
@@ -16,10 +16,17 @@
 ]
 
 
+def get_error_key(error):
+    try:
+        return error.path[0]
+    except (AttributeError, TypeError, IndexError):
+        raise ValueError(f"Unexpected validation error: {error!r}")
```
katstevens: Nice - tested locally with a missing `required` answer:

`ValueError: Unexpected validation error: <ValidationError: "'GAAR' is a required property">` 👍

PR #474, opened 2020-03-11
Add human-readable log output to Jenkins console

katstevens:
https://trello.com/c/AnvspHsL/1526-log-human-readable-stats-on-jenkins-framework-stats-output

Quick and dirty way to get a copy-and-pastable equivalent of the current framework application stats (see last year's performance platform https://www.gov.uk/performance/g-cloud-11-supplier-applications/applications-by-stage-tab/applications-by-stage-tab-applications-by-stage-day), as the current API JSON output is not the easiest to parse!

This way a developer can at least grab the latest Jenkins console output and easily share it with a Product Manager.

### DIFF
--- a/scripts/snapshot-framework-stats.py
+++ b/scripts/snapshot-framework-stats.py
@@ -26,6 +26,33 @@ def get_stats(data_client, framework_slug):
     return data_client.get_framework_stats(framework_slug)
 
 
+def log_human_readable_stats(stats):
+    total_applications = 0
+    made_declaration = 0
+    added_services = 0
+    completed = 0
+    started = 0
+
+    for stat in stats['interested_suppliers']:
+        total_applications += stat['count']
+        if stat['has_completed_services']:
+            if stat['declaration_status'] == 'complete':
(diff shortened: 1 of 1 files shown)


scripts/snapshot-framework-stats.py
```diff
@@ -26,6 +26,33 @@ def get_stats(data_client, framework_slug):
...
+                completed += stat['count']
+            elif stat['declaration_status'] == 'started':
+                added_services += stat['count']
+        else:
+            if stat['declaration_status'] == 'complete':
+                made_declaration += stat['count']
+            else:
+                started += stat['count']
+
+    logger.info(f"Started: {started}")
```
lfdebrux: Could we have one or two blank lines before and after this printout so it's easy to see?

katstevens: I can go one better than that!

lfdebrux: 😁 for a second I was bracing myself for some emojii 😆 

This is good too though!

PR #475, opened 2020-03-12
genericize (for jenkins) notify_suppliers_of_new_fw_cq_answers into notify_fw_interested_suppliers

risicle:
https://trello.com/c/RBJb5rfV

In an attempt to supply some "common context" for the templates this will be pointed at, I've included all of the available timestamps present on the `framework` in a number of different formats.

I've also gone a bit further and included a `_timetoday` format which will give you e.g. `Today at 5pm`, but have tried to make this foolproof by only emitting it for timestamps which _are_ today, _and_ if the timestamp is hour-exact. If you point it at a template that requires it and these conditions are not fulfilled, notify should throw an error and hopefully a user will be made aware of the issue.

Let's see how successful this is.


tests/test_notify_fw_interested_suppliers.py
```diff
@@ -92,16 +94,17 @@ def test_happy_paths(mock_uuid4, run_id, dry_run):
...
-        data_api_client=mock_data_api_client,
-        notify_client=mock_notify_client,
-        notify_template_id="8877eeff",
-        framework_slug="g-cloud-99",
-        dry_run=dry_run,
-        stage="production",
-        logger=mock_logger,
-        run_id=run_id,
-    ) == 0
+    with freeze_time('2000-01-04 01:55:55'):
```
katstevens: Given our important framework dates generally fall in BST, shall we use a summertime-y test date to check the localization?

dmscripts/notify_fw_interested_suppliers.py
```diff
@@ -1,16 +1,57 @@
...
 
-def notify_suppliers_of_new_fw_cq_answers(
+_date_formats = (
+    "displaytimeformat",
+    "dateformat",
+    "datetimeformat",
+)
+
+
+def _maybe_time_today(key, iso_timestamp):
```
katstevens: The word `maybe` in a programming context strikes fear into my heart.

How about `_deadline_time_today(...)`? I think we can be specific, given this script is always going to be sending emails about ongoing framework applications, and any times in these are going to be deadlines.

tests/test_notify_fw_interested_suppliers.py
```diff
@@ -120,18 +94,44 @@ def test_happy_paths(mock_uuid4, run_id, dry_run):
...
+        'clarificationsCloseAtUTC_dateformat': 'Saturday 1 January 2000',
+        'clarificationsCloseAtUTC_datetimeformat': 'Saturday 1 January 2000 at 12:00am GMT',
+        'clarificationsPublishAtUTC_displaytimeformat': '12:00am GMT',
+        'clarificationsPublishAtUTC_dateformat': 'Sunday 2 January 2000',
+        'clarificationsPublishAtUTC_datetimeformat': 'Sunday 2 January 2000 at 12:00am GMT',
+        'applicationsCloseAtUTC_displaytimeformat': '12:00am GMT',
+        'applicationsCloseAtUTC_dateformat': 'Monday 3 January 2000',
+        'applicationsCloseAtUTC_datetimeformat': 'Monday 3 January 2000 at 12:00am GMT',
+        'intentionToAwardAtUTC_displaytimeformat': '12:00am GMT',
+        'intentionToAwardAtUTC_timetoday': 'Today at 12am GMT',
```
katstevens: Can we test a non-exact time as well? Not sure if that's covered elsewhere

katstevens:
Just a reminder to all that the Jenkins job definition for this script will need to be updated before merging, as the script name has changed.

tests/test_notify_fw_interested_suppliers.py
risicle: Iffff I can find the fixture that controls that, sure...

dmscripts/notify_fw_interested_suppliers.py
risicle: I think deadline is even less clear, and I partly mean for `maybe` to stand out...

risicle:
Yes and we'll need to enumerate the template ids too.

dmscripts/notify_fw_interested_suppliers.py
katstevens: OK. Instead, could we change first part of the docstring below (`If the conditions are right`) to `If the given timestamp matches today's local date and is hour-exact`, so we're more precise there?

risicle: ok

tests/test_notify_fw_interested_suppliers.py
risicle: Done

tests/test_notify_fw_interested_suppliers.py
```diff
@@ -120,18 +97,45 @@ def test_happy_paths(mock_uuid4, run_id, dry_run):
...
+        'applicationsCloseAt_dateformat': 'Monday 3 January 2000',
+        'applicationsCloseAt_datetimeformat': 'Monday 3 January 2000 at 12:00am GMT',
+        'intentionToAwardAt_displaytimeformat': '5:00pm BST',
+        'intentionToAwardAt_timetoday': 'Today at 5pm BST',
+        'intentionToAwardAt_dateformat': 'Thursday 29 June 2000',
+        'intentionToAwardAt_datetimeformat': 'Thursday 29 June 2000 at 5:00pm BST',
+        'frameworkLiveAt_displaytimeformat': '5:30pm BST',
+        'frameworkLiveAt_dateformat': 'Thursday 29 June 2000',
+        'frameworkLiveAt_datetimeformat': 'Thursday 29 June 2000 at 5:30pm BST',
+        # no frameworkLiveAt_timetoday because it's non-hour-exact
```
katstevens: 👍

risicle:
I also removed the `...UTC` from the keys because it's misleading seeing as the formatted versions all localize the time.

katstevens:
After looking at the Jenkins updates that will be needed, I'm going to temporarily reinstate the old clarification questions script in addition to this new one. I want to be confident that our existing clarification questions job runs OK as it's currently in use.

katstevens:
> Where would it be most useful to document what variables are now available to Notify templates?

I added the info in the script docstring - given it'll be a dev creating any new templates for this script I think that's enough?

PR #476, opened 2020-03-12
408-export-application-stats

benvand:
https://trello.com/c/i4mL2aBs/408-export-application-stats

* Make application stats script output to file
* General tidying

### DIFF
--- a/scripts/snapshot-framework-stats.py
+++ b/scripts/snapshot-framework-stats.py
@@ -1,12 +1,13 @@
 #!/usr/bin/env python3
 """Usage:
-    snapshot_framework_stats.py <framework_slug> <stage>
+    snapshot_framework_stats.py <stage> <framework_slug> <filename>
 
 Example:
-    ./snapshot_framework_stats.py g-cloud-7 dev
+    ./snapshot_framework_stats.py production g-cloud-12 myfile.json
 """
 
 from docopt import docopt
+import json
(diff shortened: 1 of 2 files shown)


scripts/snapshot-framework-stats.py
```diff
@@ -1,12 +1,13 @@
 #!/usr/bin/env python3
 """Usage:
-    snapshot_framework_stats.py <framework_slug> <stage>
+    snapshot_framework_stats.py <stage> <framework_slug> <filename>
 
 Example:
-    ./snapshot_framework_stats.py g-cloud-7 dev
+    ./snapshot_framework_stats.py production g-cloud-12 myfile.json
```
katstevens: I'd rather avoid `production` in an example, just because this script writes to the DB.

Actually, this is probably a good time to write a 1-line description of what this script does 😸 How about:

`Fetches application stats for a given framework (suppliers, services and users) from the API. The response is stored in an audit event and the output is saved to a JSON file.`

katstevens:
If we're going to change this script and its args anyway (and thus update the Jenkins job), could we move it into the `scripts/framework-applications` folder with the other framework-y scripts?

benvand:
@katstevens 
Forgot to `git co master && git pull` before starting this 🤦‍♂

benvand:
Have made requested changes

PR #485, opened 2020-03-31
One off script for appending licence info to READMEs

katstevens:
https://trello.com/c/Du4aEv02/468-update-readmes-for-our-repos

Our READMEs need some TLC - hardly any of them have Licence info, as recommended by the [GDS Way](https://gds-way.cloudapps.digital/manuals/readme-guidance.html#writing-readmes). There are a lot of repos, so I thought I'd write a script to get the Jenkins github user to do some of the heavy lifting for us.

The branch-creating and PR-creating functions are lifted from the token rotation script.


katstevens:
Is there anything else we want to add in here that's common to all the repos?

katstevens:
> Pretty cool!
> 
> It might be worth adding a line to the docstring mentioning that this command will change their user name and they'll need to change it back later.

Good point - I had to rebase the API one just now and it got all confused!

PR #502, opened 2020-05-19
Add Python version file for Dependabot

lfdebrux:
We have been having issues where Dependabot has been removing importlib-metadata and dependencies from our requirements files, because jsonschema only requires it for python_version<3.8[[1]].

Our apps still use Python 3.6, so we want to run Dependabot on this version of Python, to get requirements for that version of Python. It looks like the way to do this is to create a .python-version file in the repo[[2]].

[1]: https://github.com/Julian/jsonschema/commit/0e6c684a51c5593087b8bb419d54f79db591e7d6
[2]: https://github.com/dependabot/dependabot-core/issues/1455


.python-version
```diff
@@ -0,0 +1 @@
+3.6.9
```
katstevens, line 1: Are we using 3.6.9 on the scripts docker image? I see `python:3.6-slim`

lfdebrux: 3.6-slim currently points to Python 3.6.10, unfortunately however we can't use 3.6.9 for Dependabot just yet:

https://github.com/dependabot/dependabot-core/blob/master/python/lib/dependabot/python/python_versions.rb#L12-L17

PR #503, opened 2020-05-26
Use new find_draft_services_by_framework endpoint on publish-draft-services

katstevens:
https://trello.com/c/kCUjDnH4/48-3-3-jenkins-job-for-publish-services

Uses the new status-filterable `find_draft_services_by_framework` endpoint, hopefully for efficiency savings 💸. I'd like to run this branch against preview to determine the speed gains (if any).

I've done some pretty hefty refactoring of the test coverage for the `publish-draft-services` script. Currently there's 2 'mega-tests' with parametrized scenarios - while these are comprehensive and neat, I found it difficult to determine the expected behaviour for each scenario and be confident that any changes I made to the script itself would still have coverage.

Summary of changes:
- Adds extra unit tests for the public helper functions `publish_draft_service` and `copy_draft_document`, and group them into classes
- Convert parametrized scenarios for `publish_draft_services` into separate unit tests
- Minor tweaks to the `publish_draft_service` function to make it clearer what's being returned 
- Rename `_get_draft_services_iter` to a public (testable) function and add tests
- Finally, replacing the `get_submitted_drafts` function (which will iterate through the JSON for ~40,000 services to determine each one's status, yikes) with the new `find_draft_services_by_framework` generator, which pre-filters the status, getting Postgres to do the heavy lifting.

Comments very welcome, especially if I've missed anything on the test coverage. "I liked the old test structure better because of X" is also a valid comment!


dmscripts/publish_draft_services.py
```diff
@@ -158,12 +156,14 @@ def publish_draft_service(
...
+                previously_published = True
             else:
                 raise e
 
+    return service_id, previously_published
 
-def _get_draft_services_iter(client, framework_slug, draft_ids_file=None):
+
+def get_draft_services_iter(client, framework_slug, draft_ids_file=None):
     supplier_frameworks = find_suppliers_on_framework(client, framework_slug)
```
lfdebrux, line 167: Why do we need to get `supplier_frameworks`? Couldn't we just use `find_draft_services_by_framework_iter` without a supplier id?

katstevens: We need to filter out any suppliers that failed the framework.

lfdebrux: Right got it thanks

tests/test_publish_draft_services.py
```diff
@@ -230,200 +228,6 @@ def test_get_draft_services_for_draft_ids_file_raises_on_supplier_id_mismatch(se
         assert str(excinfo.value) == "Draft service 1's supplier (878004) not on framework 'g-cloud-123'"
 
 
-@pytest.mark.parametrize("dry_run", (False, True,))
-@pytest.mark.parametrize("skip_docs_if_published", (False, True,))
-@mock.patch('dmscripts.publish_draft_services.copy_draft_documents')
-def test_publish_draft_services(copy_draft_documents, skip_docs_if_published, dry_run):
```
lfdebrux, line 18: 🔥

dmscripts/helpers/framework_helpers.py
```diff
@@ -10,14 +10,6 @@
 logger = logging.getLogger("framework_helpers")
 
 
-def get_submitted_drafts(client: DataAPIClient, framework_slug: str, supplier_id: int) -> Sequence[Mapping]:
```
lfdebrux, line 13: 🙌

tests/test_publish_draft_services.py
```diff
@@ -168,6 +168,7 @@ def setup(self):
                 SupplierFrameworkStub(supplier_id=878004, framework_slug="g-cloud-123", on_framework=False).response(),
                 SupplierFrameworkStub(supplier_id=878040, framework_slug="g-cloud-123", on_framework=True).response(),
                 SupplierFrameworkStub(supplier_id=878400, framework_slug="g-cloud-123", on_framework=True).response(),
+                SupplierFrameworkStub(supplier_id=878401, framework_slug="g-cloud-123", on_framework=True).response(),
```
lfdebrux, line 171: Can we add a line about draft service 5 to the comment above?

tests/test_publish_draft_services.py
```diff
@@ -366,55 +349,100 @@ def mock_documents_bucket_copy_side_effect(src_bucket, src_key, target_key, **kw
...
-            target_key="g-cloud-123/documents/171400/2229090909-liver.odt",
-        ),
-        mock.call.copy(
-            acl="public-read",
-            src_bucket="ducky",
-            src_key="g-cloud-123/submissions/171400/379001-mashed.pdf",
-            target_key="g-cloud-123/documents/171400/2229090909-mashed.pdf",
-        ),
-    ]
+    def test_copy_draft_documents_only_logs_on_dry_run(self):
```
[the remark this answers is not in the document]
katstevens: Good call - committed 👍

katstevens:
@lfdebrux Python 3.7+ errors now addressed. I'm going to try and run this against some 'real' (developer) buckets today.

dmscripts/publish_draft_services.py
```diff
@@ -188,9 +192,13 @@ def get_draft_services_iter(client, framework_slug, draft_ids_file=None):
...
-            yield from client.find_draft_services_by_framework_iter(
-                framework_slug, status='submitted', supplier_id=sf['supplierId']
-            )
+            try:
+                yield from client.find_draft_services_by_framework_iter(
+                    framework_slug, status='submitted', supplier_id=sf['supplierId']
+                )
+            except StopIteration:
+                # Handle exception for Python 3.7 and above
+                return None
```
lfdebrux, line 201: This is quite weird, looking at [`make_iter_method`](https://github.com/alphagov/digitalmarketplace-apiclient/blob/2158a624cf52d49ae8596e84fef03e60bce87ae0/dmapiclient/base.py#L25) in digitalmarketplace-apiclient, we shouldn't be raising `StopIteration` in `find_draft_services_by_framework_iter`...

katstevens: I think it's because we're `yield from`-ing.

lfdebrux: I think it's the mock that is causing the error, rather than production code.

We could either leave this `try... except` block inplace

tests/test_publish_draft_services.py
```diff
@@ -1,298 +1,580 @@
...
+            mock.call('g-cloud-123', status='submitted', supplier_id=878400),
+            mock.call('g-cloud-123', status='submitted', supplier_id=878401),
+        ]
 
-    mock_documents_bucket = mock.create_autospec(S3, bucket_name="puddeny-pie")
+    def test_get_draft_services_for_draft_ids_file_includes_not_submitted_services(self):
+        self.mock_data_api_client.get_draft_service.side_effect = [
+            {"services": self.draft_service2},
+            {"services": self.draft_service3}
+        ]
```
lfdebrux, line 218: This is raising `StopIteration` because it is being called more times than expected 🤔

katstevens: Aha! Good thing we are testing with Python 3.7 so this was caught 😸

katstevens: So one of the 2 failing was definitely a problematic mock, the other failure was the `draft_ids_file` not being recognised as a file, and splitting it into 6 'lines' rather than 2  🤦‍♀️ Anyway, commit now fixed up.

PR #507, opened 2020-06-03
fix specialist field in CSV

gidsg:
https://trello.com/c/JDz7V5SO/1704-dos-opportunity-export-csv-missing-specialist-role
Previously this was looking for a field which was not returned by the API.

### DIFF
--- a/dmscripts/export_dos_opportunities.py
+++ b/dmscripts/export_dos_opportunities.py
@@ -42,7 +42,7 @@ def _build_row(brief, brief_responses):
         PUBLIC_BRIEF_URL.format(brief['id']),
         brief['frameworkSlug'],
         brief['lotSlug'],
-        brief.get('specialist', ""),
+        brief.get('specialistRole', ""),
         brief['organisation'],
         remove_username_from_email_address(brief['users'][0]['emailAddress']),
         brief['location'],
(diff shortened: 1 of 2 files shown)


dmscripts/export_dos_opportunities.py
```diff
@@ -42,7 +42,7 @@ def _build_row(brief, brief_responses):
         PUBLIC_BRIEF_URL.format(brief['id']),
         brief['frameworkSlug'],
         brief['lotSlug'],
-        brief.get('specialist', ""),
+        brief.get('specialistRole', ""),
```
lfdebrux, line 45: It might be worth updating the column heading above to also read "Specialist Role"

gidsg: @lfdebrux I don't mind doing this, but my thought was that if people have scripts which rely on the existing field names they would break if we rename an existing field.

lfdebrux: That's a good point. On the other hand their scripts can't have been doing much with a field that is always empty... Not sure. Up to you!

gidsg: I guess that scripts may allow for empty value though, whereas they might trip/break on a different field name, so my preference is to keep it as-is.

PR #510, opened 2020-07-07
Turn off Mailchimp user tracking

katstevens:
https://trello.com/c/Qk3SHoov/

We no longer need to track opens/clicks for individual users on the Mailchimp DOS Opportunities email campaigns (see ticket for details).

We will still be able to track anonymised clicks per campaign via analytics, using the `utm` suffix on the links.

### DIFF
--- a/dmscripts/send_dos_opportunities_email.py
+++ b/dmscripts/send_dos_opportunities_email.py
@@ -68,8 +68,8 @@ def get_campaign_data(lot_name, list_id, framework_name):
             "fb_comments": False
         },
         "tracking": {
-            "opens": True,
-            "html_clicks": True,
+            "opens": False,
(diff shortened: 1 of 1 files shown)


lfdebrux:
For future reference, could you post a link to the Mailchimp documentation where you found the meaning of thse settings? I'm struggling to find it with Google.

katstevens:
> For future reference, could you post a link to the Mailchimp documentation where you found the meaning of these settings? I'm struggling to find it with Google.

[Here's the Mailchimp API reference](https://mailchimp.com/developer/reference/campaigns/) - it's under `campaigns` then you have to expand the `tracking`. It's not obvious and you can't Ctrl+F it 😿

PR #512, opened 2020-07-07
Allow re-run of incomplete applications reminder via supplier ID list

katstevens:
https://trello.com/c/EnPFfpum

Part of the work to make our critical framework scripts more robust against Notify failures.

This script sends a reminder to suppliers with incomplete applications, 1 week before the deadline. It already logs the supplier ID if email sending fails. This PR adds functionality for a developer to re-run the script, passing a comma-separated list of supplier IDs to re-run against.

I've also refactored it a little:
- moved the initialisation of the `logger` and `DataAPIClient` instances to the top level script (to help this script be more consistent with our other scripts, and simplify the test setup). I tried moving the Notify client there as well but it was Unhappy. I may give this another go if people are keen but I was trying not to spend too much time on refactoring!
- moved the tests into a class, so we can use shared set up of mocks where possible
- made the main `notify_suppliers_with_incomplete_applications` function less complex by pulling out the code into a separate function

### DIFF
--- a/dmscripts/notify_suppliers_with_incomplete_applications.py
+++ b/dmscripts/notify_suppliers_with_incomplete_applications.py
@@ -47,6 +47,26 @@ def send_notification(mail_client, message, framework, email, supplier_id, dry_r
     return 0
 
 
+def build_message(sf, framework_slug, data_api_client):
+    message = ''
+    if not sf.get('applicationCompanyDetailsConfirmed', None):
+        message += MESSAGES['unconfirmed_company_details']
+    if not sf.get('declaration', {'status': None}).get('status', None) == 'complete':
+        message += MESSAGES['incomplete_declaration']
+
+    submitted_draft_services, unsubmitted_draft_services = 0, 0
+    for service in data_api_client.find_draft_services_iter(sf['supplierId'], framework=framework_slug):
+        if service.get('status') == 'not-submitted':
+            unsubmitted_draft_services += 1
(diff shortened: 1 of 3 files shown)


dmscripts/notify_suppliers_with_incomplete_applications.py
```diff
@@ -61,22 +74,14 @@ def notify_suppliers_with_incomplete_applications(framework_slug, stage, notify_
...
-        if not sf.get('applicationCompanyDetailsConfirmed', None):
-            message += MESSAGES['unconfirmed_company_details']
-        if not sf.get('declaration', {'status': None}).get('status', None) == 'complete':
-            message += MESSAGES['incomplete_declaration']
+        # Restrict suppliers to those specified in the argument, if given.
+        # While this is inefficient for a small number of supplier IDs, looking up
+        # each supplier individually for a large number of supplier IDs would be worse.
+        if supplier_ids:
+            if sf['supplierId'] not in supplier_ids:
+                continue
```
gidsg, line 82: could this be expressed using `filter` instead?

katstevens: Possibly? I doubt `filter` will gain us anything much performance-wise though. Also we'd have to do some fiddling to get it to work on this iterable, as the client method is doing some magic to paginate through our API's response (hence the `_iter` suffix).

PR #514, opened 2020-07-09
Add tests for Notify suppliers whether application made script

katstevens:
https://trello.com/c/EnPFfpum/

This script already provides a facility for re-running with a list of supplier IDs from a previous failed run. However there were no tests for it 🙀 

I've added some! This meant moving some of the code to a separate module (consistent structure with our other scripts).

### DIFF
--- a/tests/test_notify_suppliers_whether_application_made_for_framework.py
+++ b/tests/test_notify_suppliers_whether_application_made_for_framework.py
@@ -30,8 +30,8 @@ class TestNotifySuppliersWhetherApplicationMade:
             {"supplierId": 712346},
         ]
         self.data_api_client.export_users.return_value = {
-            # This endpoint returns some unusually formmatted keys :(
-            # See https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/users.py#L417
+            # This endpoint returns some unusually formatted keys :(
+            # See https://github.com/alphagov/digitalmarketplace-api/blob/6ad35e526fcf4d76320159a1a4ac97133a1ce13d/app/main/views/users.py#L417  # noqa
             'users': [
                 {'supplier_id': 712345, 'email address': 'user1@example.com', 'application_status': 'no_application'},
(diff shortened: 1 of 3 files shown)


tests/test_notify_suppliers_whether_application_made_for_framework.py
```diff
@@ -0,0 +1,202 @@
...
+                'frameworkLiveAtUTC': "2017-12-31T23:59:59.999999Z"
+            }
+        }
+        self.data_api_client.find_framework_suppliers_iter.return_value = [
+            {"supplierId": 712345},
+            {"supplierId": 712346},
+        ]
+        self.data_api_client.export_users.return_value = {
+            # This endpoint returns some unusually formmatted keys :(
+            # See https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/users.py#L417
```
gidsg, line 34: This link might bit-rot as it points to `master`. Could we link perhaps to a specific hash or PR where this was introduced?

katstevens: Good spot! Will fix

tests/test_notify_suppliers_whether_application_made_for_framework.py
```diff
@@ -0,0 +1,202 @@
...
+            None,
+            None
+        ]
+
+        assert notify_suppliers_whether_application_made(
+            self.data_api_client,
+            self.notify_client,
+            'g-cloud-12',
+            self.logger
+        ) == 1
```
gidsg, line 180: Possibly my lack of context. But I am not sure of the meaning of `0` and `1` in these assertions. Is it possible to explain this at all?

katstevens: Yep, that's a good call - it's the expected number of errors returned by the function (fed into the exit code of the top level script). I'm happy to make this a named var in the test so it's clear - e.g `expected_error_count`.

PR #525, opened 2020-08-21
Get log events in epoch intervals

agalamatis:
trello: https://trello.com/c/e04K4fMI/1833-one-off-script-to-retrieve-logs-from-cloudwatch

### DIFF
--- a/dmscripts/get_log_events.py
+++ b/dmscripts/get_log_events.py
@@ -1,0 +1,112 @@
+import boto3
+import json
+import os
+import uuid
+
+
+class LogRetriever(object):
+    """
+    Encapsulates the log retrieval process.
+    To fetch all log events of a know stream use get_log_events.
+    To fetch all log events from multiple streams within a date range,
+    use get_log_event_in_epoch_range.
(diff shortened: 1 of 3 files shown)


dmscripts/get_log_events.py
```diff
@@ -0,0 +1,112 @@
...
+                logGroupName=self.group,
+                logStreamName=stream,
+                startFromHead=True,
+                nextToken=response['nextForwardToken']
+            )
+            with open(f'{directory}/{counter}.json', 'w') as f:
+                json.dump(response, f)
+        return counter + 1
+
+    def get_log_event_in_epoch_range(self, directory, earliest_epoch, latest_epoch):
```
katstevens, line 97: Could we supply the extra args to the main script and use it conditionally if those are present? Or do we just want to keep it as an option in a Python shell?

agalamatis: Yes, but I'd rather have a new script using the same class, instead of working with optional argument in docopt.

katstevens (approved):
Good stuff 👍

PR #527, opened 2020-08-24
[Experimental] Alternative method for updating draft services and documents

katstevens:
This PR contains two experimental scripts that could provide an alternative way for users to submit G-Cloud services, for example if accessibility problems prevented them from using the usual submission journey.

1. Update existing draft services from a folder of CSV files

The CSV files require a specific naming structure: `<supplier_id>-<lot-slug>-<service-name>.csv`. Lot is required in order to retrieve the correct questions for the service.

The CSV file content also needs to match the format exported by [scripts/framework-applications/generate-questions-csv.py](https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/framework-applications/generate-questions-csv.py). The script also currently hardcodes the number of question columns per lot.

Some service questions will need manual tweaking in the frameworks repo to give a correct match (converting from a markup object to a string).

Pricing also needs special handling, which is normally done by the Supplier FE form input.

2. Update existing draft services from a folder of PDF documents

The PDF files require a specific naming structure: `<supplier_id>-<service-name>-<document-type>.pdf`. Document type (eg `pricing`) is required in order to update the correct field on the draft service with the path to the document.

Both scripts require a match on service name to find an existing draft service - some punctuation/spacing is removed to assist with this. Ambiguous matches are rejected.

Both scripts output the following reports: `unidentified-files` (could not match supplier or draft service), `successful` (successful update) and `failed` (failed validation). 

TODO:
- ~Refactor `parse_csv_and_update_draft_service` into testable function~
- ~Flesh out remaining tests~

Out of scope:
- creating new services (in follow up branch)
- marking services as complete
- DOS services
- updating declarations
- uploading Modern Slavery statement documents
- handling ODF files (in followup branch)
- PDF file size validation or typechecking with `fleep`
- Look at matching questions by id (eg `serviceBenefits`) rather than question text. Would require change to question export script. 
- using a proper logger instead of `print` everywhere 😹

### DIFF
--- a/scripts/oneoff/upload_draft_service_pdfs.py
+++ b/scripts/oneoff/upload_draft_service_pdfs.py
@@ -1,0 +1,267 @@
+#!/usr/bin/env python
+"""
+Experimental script to provide an alternative method of manually uploading documents for draft services,
+for example if a user has accessibility issues that prevent them using the site, they could submit documents
+to CCS via email.
+
+IMPORTANT: ensure you have the correct AWS_PROFILE env variable set for the environment, otherwise the S3 uploads will
+fail.
+
+- For each file in the given folder
+- Look up draft service ID by supplier ID and service name
+- Determine whether doc is pricing/SFIA/T&C/service definition
+- Upload documents to S3, following the standardised naming scheme <draft_id>-<doctype>-<datestamp>.pdf
+- Update draft service with S3 URLs
+
+Any failures will skip and continue to the next document in the folder.
+
+Not included: Modern Slavery Statement documents
+
+Usage: upload_draft_service_pdfs <framework> <stage> [options]
+
+Options:
+    --folder=<folder>                                     Folder to bulk-upload files
+    --dry-run                                             List actions that would have been taken
+    -h, --help                                            Show this screen
+
+"""
+import sys
+import os
+import urllib.parse as urlparse
+from dmapiclient.data import DataAPIClient
+
+sys.path.insert(0, '.')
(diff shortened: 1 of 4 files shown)


katstevens:
Still some bugs to iron out in this around the content loader for partial updates

scripts/oneoff/upload_draft_service_pdfs.py
```diff
@@ -0,0 +1,267 @@
...
+            continue
+        if doc_type is None:
+            print("Unable to determine doctype")
+            unidentifiable_files.append(file_name)
+            continue
+        if draft_id is None:
+            print("Unable to determine Draft ID")
+            unidentifiable_files.append(file_name)
+            continue
+        print(f"Found Supplier ID {supplier_id}, file type {doc_type} and Draft ID {draft_id}")
```
gidsg, line 200: Sometimes filenames have erroneous whitespace which would cause the supplier ID to be invalid.

```suggestion
        # Strip erroneous whitespace from supplier IDs
        supplier_id = supplier_id.strip()
        print(f"Found Supplier ID {supplier_id}, file type {doc_type} and Draft ID {draft_id}")
```

gidsg (approved):
LGTM :+1:

scripts/oneoff/upload_draft_service_pdfs.py
katstevens: This is now handled in the `get_info_from_filename` function where we cast `supplier_id` to an int, hopefully that'll get rid of any pesky whitespace.

PR #535, opened 2020-09-11
Add dates to ITA script

katstevens:
https://trello.com/c/WkNHIRt7/300-05-tell-successful-suppliers-the-result-intention-to-award-11-sept

Our new template needs some extra dates.

### DIFF
--- a/scripts/framework-applications/notify-successful-suppliers-for-framework.py
+++ b/scripts/framework-applications/notify-successful-suppliers-for-framework.py
@@ -68,8 +68,16 @@ if __name__ == '__main__':
 
     prefix = "[Dry Run] " if DRY_RUN else ""
 
+    # Add in any framework-specific dates etc here
+    extra_template_context = {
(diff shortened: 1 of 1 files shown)


scripts/framework-applications/notify-successful-suppliers-for-framework.py
```diff
@@ -66,9 +68,17 @@
 
     prefix = "[Dry Run] " if DRY_RUN else ""
 
+    # Add in any framework-specific dates etc here
+    extra_template_context = {
+        "intentionToAwardAt_dateformat": "12 September 2020",
+        "frameworkLiveAt_dateformat": "28 September 2020",
+    }
```
gidsg, line 75: @katstevens If we have to hardcode these dates should we add a TODO to update them for the next framework?

katstevens: Sure - a TODO will make it a bit more obvious.

gidsg (approved):
📧 🦅 👍

PR #536, opened 2020-09-11
Make the script to notify successful suppliers easier to understand

bjgill:
We ran this earlier today, and noticed some possible improvements.

Fix the doc comment to include all the parameters and recommend providing the list of suppliers.

Also add in logs to the parts that take a while to give the person running the script an indication of progress.

I've manually tested that this gives the expected output when we run the script (with `--dry-run`, of course)


gidsg (approved):
👍

PR #539, opened 2020-09-17
Take updated dependencies

bjgill:
Will allow us to make use of https://github.com/alphagov/digitalmarketplace-apiclient/pull/234/, and other recent changes.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,7 +4,6 @@
 #
 #    pip-compile requirements-dev.in
 #
-appnope==0.1.0            # via ipython
 attrs==19.3.0             # via -c requirements.txt, pytest
 backcall==0.1.0           # via ipython
 certifi==2019.11.28       # via -c requirements.txt, requests
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,6 +4,7 @@
 #
 #    pip-compile requirements-dev.in
 #
+appnope==0.1.0            # via ipython
 attrs==19.3.0             # via -c requirements.txt, pytest
 backcall==0.1.0           # via ipython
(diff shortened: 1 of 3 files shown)


lfdebrux:
You'll need to run `make freeze-requirements` to see the changes reflected in `requirements.txt`.

bjgill:
Thanks - fixed and rebased onto master

requirements-dev.txt
```diff
@@ -17,7 +16,6 @@ entrypoints==0.3          # via flake8
 flake8==3.7.9             # via -r requirements-dev.in
 freezegun==0.3.15         # via -r requirements-dev.in
 idna==2.9                 # via -c requirements.txt, requests
-importlib-metadata==1.5.0  # via -c requirements.txt, pluggy, pytest
```
lfdebrux, line 20: Which version of python were you using, 3.6 or 3.8?

bjgill: Ah-ha. That was what was causing that. I've switched to 3.6, and most of the changes have gone away. `appnope` is still going away, but I don't know why.

lfdebrux: Probably because you're using linux... ["appnope: Disable App Nap on OS X 10.9"](https://pypi.org/project/appnope/). Unfortunately pip-tools isn't clever enough to leave it in for you, and there isn't a way to spoof the environment that I know of. I guess I can just run `freeze-requirements` for you. Why on earth does IPython want appnope anyway? 🤔

lfdebrux: Pushed frozen requirements.

lfdebrux (approved):

requirements-dev.txt
bjgill: That explains that. Thanks.

PR #540, opened 2020-09-18
Allow read-only access to the API client shell

bjgill:
Having direct access to production faintly terrifies me. Sometimes I just need to get information from prod. In those cases, it'd be useful to have protection against accidentally making changes.

Override `__getattr__` for access control. Override `__dir__` so that autocomplete in the iPython shell works.

I've done this on the assumption that API methods that start with "get_" or "find_" are read-only methods.

### DIFF
--- a/scripts/api-clients-shell.py
+++ b/scripts/api-clients-shell.py
@@ -10,6 +10,7 @@ Optional:
   --api-token         Override for the API key (don't decrypt from dm-credentials)
   --search-api-url    Override the implicit SearchAPI URL
   --search-api-token  Override for the SearchAPI key (don't decrypt from dm-credentials)
+  --read-only         Only allow access to API calls that read data
 
 Example:
   ./scripts/api-clients-shell.py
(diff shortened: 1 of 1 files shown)


lfdebrux (approved):
Ooh, I like this very much

scripts/api-clients-shell.py
```diff
@@ -10,6 +10,7 @@
   --api-token         Override for the API key (don't decrypt from dm-credentials)
   --search-api-url    Override the implicit SearchAPI URL
   --search-api-token  Override for the SearchAPI key (don't decrypt from dm-credentials)
+  --read-only         Only allow access to API calls that read data
```
lfdebrux, line 13: Can we have a short flag as well as a long flag for this?

```suggestion
  -r, --read-only         Only allow access to API calls that read data
```

bjgill: Great idea - I've just added a commit to do that.

PR #541, opened 2020-09-18
Adapt countersigning job for e-signatures

gidsg:
https://trello.com/c/UfmJ2EOd/308-adapt-countersigning-jenkins-job-for-e-signature-flow

This makes a number of changes to the script to support e-signature counterpart documents

* Script is responsible for automatically approving the agreement using the API (This would have been done previously in admin-frontend)
* Adapt script to handle multiple documents to facilitate generation of a complete contract PDF

Co-authored-by: Laurence de Bruxelles <laurence.debruxelles@digital.cabinet-office.gov.uk>

Note, this is dependent on sister PR https://github.com/alphagov/digitalmarketplace-agreements/pull/10

### DIFF
--- a/dmscripts/generate_framework_agreement_signature_pages.py
+++ b/dmscripts/generate_framework_agreement_signature_pages.py
@@ -151,7 +151,6 @@ def render_pdf_for_each_html_page(html_pages, html_dir, pdf_dir):
                             2: "4"}
 
             exit_code = subprocess.call(['wkhtmltopdf',
-                                         '--enable-local-file-access',
                                          '--footer-right',
                                          page_numbers.get(index, ""),
                                          '--margin-bottom',
--- a/dmscripts/generate_framework_agreement_signature_pages.py
+++ b/dmscripts/generate_framework_agreement_signature_pages.py
@@ -133,7 +133,7 @@ def render_html_for_suppliers_awaiting_countersignature(rows, framework, templat
         yield html_dir, output_pages
 
 
-def render_pdf_for_each_html_page(html_pages, html_dir, pdf_dir, framework_slug):
+def render_pdf_for_each_html_page(html_pages, html_dir, pdf_dir):
     html_dir = Path(html_dir).resolve()
     pdf_dir = Path(pdf_dir).resolve()
 
@@ -143,9 +143,15 @@ def render_pdf_for_each_html_page(html_pages, html_dir, pdf_dir, framework_slug)
         os.mkdir(pdf_dir)
 
     if len(html_pages) > 1:
-        for html_path in html_pages:
+        for index, html_path in enumerate(html_pages):
             pdf_path = html_dir / (html_path.stem + ".pdf")
-            exit_code = subprocess.call(['wkhtmltopdf', 'file://{}'.format(html_path), pdf_path])
+            # Insert page number for dynamically generated pages
+            footer_page_number = str(index + 1) if index <= 3 else ""
+            exit_code = subprocess.call(['wkhtmltopdf',
+                                         '--footer-right',
+                                         footer_page_number,
+                                         'file://{}'.format(html_path),
+                                         pdf_path])
             if exit_code > 0:
                 logger.error("ERROR {} on {}".format(exit_code, html_path))
                 ok = False
--- a/dmscripts/generate_framework_agreement_signature_pages.py
+++ b/dmscripts/generate_framework_agreement_signature_pages.py
@@ -147,8 +147,8 @@ def render_pdf_for_each_html_page(html_pages, html_dir, pdf_dir):
             pdf_path = html_dir / (html_path.stem + ".pdf")
             # Insert page number for dynamically generated pages
             page_numbers = {0: "1",
-                            2: "3",
-                            3: "4"}
+                            1: "3",
+                            2: "4"}
 
             exit_code = subprocess.call(['wkhtmltopdf',
                                          '--footer-right',
--- a/dmscripts/generate_framework_agreement_signature_pages.py
+++ b/dmscripts/generate_framework_agreement_signature_pages.py
@@ -1,15 +1,17 @@
 import os
 import io
 import shutil
-import re
 import subprocess
+from pathlib import Path
 
 from datetime import datetime
+from PyPDF2 import PdfFileMerger
 
 from dmscripts.helpers.html_helpers import render_html
 from dmscripts.helpers.logging_helpers import get_logger
 from dmscripts.helpers.framework_helpers import (
-    find_suppliers_with_details_and_draft_service_counts
+    find_suppliers_with_details_and_draft_service_counts,
+    framework_supports_e_signature
 )
 from dmscripts.export_framework_applicant_details import get_csv_rows
 
@@ -51,6 +53,7 @@ def save_page(html, supplier_id, output_dir, descriptive_filename_part):
     page_path = os.path.join(output_dir, '{}-{}.html'.format(supplier_id, descriptive_filename_part))
     with io.open(page_path, 'w+', encoding='UTF-8') as htmlfile:
         htmlfile.write(html)
+    return page_path
 
 
 def render_html_for_successful_suppliers(rows, framework, template_dir, output_dir, dry_run=False):
@@ -78,9 +81,19 @@ def render_html_for_successful_suppliers(rows, framework, template_dir, output_d
 
 
 def render_html_for_suppliers_awaiting_countersignature(rows, framework, template_dir, output_dir):
-    template_path = os.path.join(template_dir, 'framework-agreement-signature-page.html')
-    template_css_path = os.path.join(template_dir, 'framework-agreement-signature-page.css')
-    template_countersignature_path = os.path.join(template_dir, 'framework-agreement-countersignature.png')
+    output_dir = Path(output_dir).resolve()
+    html_pages = [Path(template_dir, 'framework-agreement-signature-page.html').resolve()]
+    static_files = [Path(template_dir, 'framework-agreement-signature-page.css').resolve()]
+
+    if not framework_supports_e_signature(framework['slug']):
+        static_files.append(Path(template_dir, 'framework-agreement-countersignature.png'))
+
+    if framework_supports_e_signature(framework['slug']):
+        html_pages.append(Path(template_dir, 'framework-agreement-cover-page.html'))
+        static_files.append(Path(template_dir, 'ccs_logo.png'))
+        static_files.append(Path(template_dir, 'framework-agreement-toc.pdf'))
+        static_files.append(Path(template_dir, 'framework-agreement-boilerplate.pdf'))
+
     for data in rows:
         if data['pass_fail'] == 'fail' or data['countersigned_path'] or not data['countersigned_at']:
             logger.info("SKIPPING {}: pass_fail={} countersigned_at={} countersigned_path={}".format(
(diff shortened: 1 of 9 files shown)


dmscripts/generate_framework_agreement_signature_pages.py
```diff
@@ -95,28 +108,65 @@ def render_html_for_suppliers_awaiting_countersignature(rows, framework, templat
         data['countersigned_at'] = datetime.strptime(
             data['countersigned_at'], '%Y-%m-%dT%H:%M:%S.%fZ'
         ).strftime('%d %B %Y')
+        if data['signed_agreement_returned_at']:
+            data['signed_agreement_returned_at'] = datetime.strptime(
+                data['signed_agreement_returned_at'], '%Y-%m-%dT%H:%M:%S.%fZ'
+            ).strftime('%d %B %Y')
```
bjgill, line 114: Why do we need to update this if it's already set?

scripts/generate-framework-agreement-counterpart-signature-pages.py
```diff
@@ -95,17 +97,30 @@
...
         rows, framework, os.path.join(args['<path_to_agreements_repo>'], 'documents', framework['slug']), html_dir
     )
-    html_pages = os.listdir(html_dir)
-    html_pages.remove('framework-agreement-signature-page.css')
-    html_pages.remove('framework-agreement-countersignature.png')
-    ok = render_pdf_for_each_html_page(html_pages, html_dir, args['<output_folder>'])
+
+    ok = True
+    for html_dir, html_pages in html_to_render:
+        ok = render_pdf_for_each_html_page(html_pages, html_dir, args['<output_folder>'],
```
bjgill, line 121: Nit: would this be clearer with `&=`?

dmscripts/generate_framework_agreement_signature_pages.py
```diff
@@ -78,9 +81,19 @@ def render_html_for_successful_suppliers(rows, framework, template_dir, output_d
...
-    template_css_path = os.path.join(template_dir, 'framework-agreement-signature-page.css')
-    template_countersignature_path = os.path.join(template_dir, 'framework-agreement-countersignature.png')
+    output_dir = Path(output_dir).resolve()
+    html_pages = [Path(template_dir, 'framework-agreement-signature-page.html').resolve()]
+    static_files = [Path(template_dir, 'framework-agreement-signature-page.css').resolve()]
+
+    if not framework_supports_e_signature(framework['slug']):
+        static_files.append(Path(template_dir, 'framework-agreement-countersignature.png'))
+
+    if framework_supports_e_signature(framework['slug']):
```
bjgill, line 91: Nit: can we make this more obvious with `else`?

dmscripts/generate_framework_agreement_signature_pages.py
```diff
@@ -95,28 +108,65 @@ def render_html_for_suppliers_awaiting_countersignature(rows, framework, templat
...
         os.mkdir(pdf_dir)
-    for index, html_page in enumerate(html_pages):
-        html_path = os.path.join(html_dir, html_page)
-        pdf_path = '{}'.format(re.sub(r'\.html$', '.pdf', html_path))
-        pdf_path = '{}'.format(re.sub(html_dir, pdf_dir, pdf_path))
+
+    if len(html_pages) > 1:
+        for html_path in html_pages:
+            pdf_path = html_dir / (html_path.stem + ".pdf")
+            exit_code = subprocess.call(['wkhtmltopdf', 'file://{}'.format(html_path), pdf_path])
```
bjgill, line 142: > `wkhtmltopdf`

Do we need to document this as a requirement of this script?

scripts/generate-framework-agreement-counterpart-signature-pages.py
```diff
@@ -70,6 +70,8 @@
 from dmapiclient import DataAPIClient
 from dmutils.env_helpers import get_api_endpoint_from_stage
 
+AUTOMATED_COUNTERSIGNING_USER_ID = 64041
```
bjgill, line 73: Should we document why we've chosen this number?

bjgill (commented):
Looks good. Just one thing I didn't understand and a couple of things we might want to document.

dmscripts/generate_framework_agreement_signature_pages.py
gidsg: This is to reformat the date for display converting it from the format returned by the API.
This follows the same pattern as https://github.com/alphagov/digitalmarketplace-scripts/blob/528ad08732531041c0467724f788e18435a5f4af/dmscripts/generate_framework_agreement_signature_pages.py#L108-L110

scripts/generate-framework-agreement-counterpart-signature-pages.py
gidsg: updated in 0274a56

dmscripts/generate_framework_agreement_signature_pages.py
gidsg: @bjgill This is documented in the parent script used to call this function: https://github.com/alphagov/digitalmarketplace-scripts/blob/0274a566c88a185833899523870b78c35382cd87/scripts/generate-framework-agreement-counterpart-signature-pages.py#L4-L5

dmscripts/generate_framework_agreement_signature_pages.py
gidsg: good point, fixed in 60c936c

dmscripts/generate_framework_agreement_signature_pages.py
```diff
@@ -95,28 +114,66 @@ def render_html_for_suppliers_awaiting_countersignature(rows, framework, templat
...
+
+        html_dir = output_dir / str(data['supplier_id'])
+        output_pages = []
+        for html_page in html_pages:
+            html = render_html(str(html_page), data)
+            page_path = save_page(html, data['supplier_id'], html_dir, html_page.stem)
+            output_pages.append(Path(page_path).resolve())
+
+        for static_file in static_files:
+            shutil.copyfile(static_file, Path(html_dir, static_file.name))
```
lfdebrux, line 131: It occured to me that we can save a lot of space and copy operations by using symlinks for static files:

```suggestion
            Path(html_dir, static_file.name).symlink_to(static_file.resolve())
```

What do you think?

gidsg: I don't see why not, are we sure this will have the same behaviour on Jenkins?

scripts/generate-framework-agreement-counterpart-signature-pages.py
gidsg: fixed in 4dc6daa

dmscripts/generate_framework_agreement_signature_pages.py
lfdebrux: It should do, I don't think there's anything special about the Jenkins filesystem in this regard.

tests/helpers/test_framework_helpers.py
```diff
@@ -66,6 +67,7 @@ def test_find_suppliers_with_details_and_draft_services(mock_data_client):
         },
         (2, 'framework-slug'): {
             'frameworkInterest': {
+                'agreementId': None,
```
phantas, line 70: how can it be countersigned but have an `agreementId` of `None`?
It's not being tested, but I would still suggest to set it to `1` so the test data is correct.

tests/helpers/test_framework_helpers.py
```diff
@@ -186,6 +203,7 @@ def test_find_suppliers_with_details_and_draft_service_counts(mock_data_client):
                 'onFramework': True,
                 'countersignedPath': 'some/path',
                 'countersignedAt': '2017-01-02T03:04:05.000006Z',
+                'agreementId': None
```
phantas, line 206: Same as above: how can it be countersigned but have an `agreementId` of `None`?
I know it's not being tested, but I suggest setting it to `31385` just to ensure the test data is correct.

tests/helpers/test_framework_helpers.py
```diff
@@ -150,11 +154,22 @@ def test_find_suppliers_with_details_and_draft_services(mock_data_client):
...
                 'onFramework': True,
                 'frameworkSlug': 'g-things-1',
+                'agreementId': ""
             }
         ]
 
 
 def test_find_suppliers_with_details_and_draft_service_counts(mock_data_client):
+    mock_data_client.get_framework_agreement.return_value = {'agreement': {'frameworkSlug': 'g-cloud-12',
+                                                                           'id': 31385,
```
phantas, line 164: shouldn't this be `agreementId` instead of `id`?

tests/helpers/test_framework_helpers.py
```diff
@@ -219,7 +237,11 @@ def test_find_suppliers_with_details_and_draft_service_counts(mock_data_client):
             ),
             'countersignedAt': '',
             'onFramework': True,
+            'signedAgreementReturnedAt': '2020-09-15T15:52:15.677327Z',
+            'signerName': 'gideon',
+            'signerRole': 'developer',
             'frameworkSlug': 'g-things-1',
+            'agreementId': ''
```
phantas, line 244: I think the reason you are getting an empty string here is because you didn't set the `agreementId` on your mock data (possibly that was what `id` was supposed to be?)

tests/helpers/test_framework_helpers.py
```diff
@@ -236,7 +258,11 @@ def test_find_suppliers_with_details_and_draft_service_counts(mock_data_client):
             ),
             'countersignedAt': '',
             'onFramework': False,
+            'signedAgreementReturnedAt': '2020-09-15T15:52:15.677327Z',
+            'signerName': 'gideon',
+            'signerRole': 'developer',
             'frameworkSlug': 'g-things-1',
+            'agreementId': ''
```
phantas, line 265: Same as above

tests/helpers/test_framework_helpers.py
```diff
@@ -253,7 +279,11 @@ def test_find_suppliers_with_details_and_draft_service_counts(mock_data_client):
             ),
             'countersignedAt': '2017-01-02T03:04:05.000006Z',
             'onFramework': True,
+            'signedAgreementReturnedAt': '2020-09-15T15:52:15.677327Z',
+            'signerName': 'gideon',
+            'signerRole': 'developer',
             'frameworkSlug': 'g-things-1',
+            'agreementId': ''
```
phantas, line 286: Same as above

dmscripts/generate_framework_agreement_signature_pages.py
```diff
@@ -95,28 +114,79 @@ def render_html_for_suppliers_awaiting_countersignature(rows, framework, templat
...
         data['includeCountersignature'] = True
-        html = render_html(template_path, data)
-        save_page(html, data['supplier_id'], output_dir, "agreement-countersignature")
-    shutil.copyfile(template_css_path, os.path.join(output_dir, 'framework-agreement-signature-page.css'))
-    shutil.copyfile(
-        template_countersignature_path,
-        os.path.join(output_dir, 'framework-agreement-countersignature.png')
-    )
+
+        html_dir = output_dir / str(data['supplier_id'])
```
phantas, line 123: So, we are changing the path from
`output_dir/<files>` to
`output_dir/1234/<files>`...
I think this is fine, because it's just a temporary location and we are passing the full path every time, as far as I can see, but we may want to do some broader testing on preview to check nothing breaks. since dmscripts is not part of functional/smoke tests.

phantas (dismissed):
Heya, sorry for delay but this was a BIG PR and I had to go though it carefully.
Besides the comments, I would also like to see some tests for:
- that we are only countersigning agreements that require it
- a test that explicitly tests for frameworks on e-signature
- a test that explicitly tests we are backwards compatible with frameworks not on e-signature.
That said, I know we are under a time pressure so use best judgment on what to do now and what to do later?

dmscripts/helpers/framework_helpers.py
```diff
@@ -45,6 +45,17 @@ def find_suppliers_with_details_and_draft_services(
...
 
+def framework_supports_e_signature(framework_slug):
+    """
+    This is a hacky way to determine if a framework supports e-signature,
+    which will be frameworks from G-Cloud-12 onwards
+    # TODO: Do more robust check
+    :param framework_slug:
+    :return: Boolean
+    """
+    return framework_slug in ['g-cloud-12']
```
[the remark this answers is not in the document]
gidsg: OK, have used this same heuristic now in 91828c5

scripts/generate-framework-agreement-counterpart-signature-pages.py
```diff
@@ -95,17 +98,29 @@
     html_dir = tempfile.mkdtemp()
 
     records = find_suppliers_with_details_and_draft_service_counts(client, framework_slug, supplier_ids)
+    records = list(records)
+    for record in records:
+        if framework_supports_e_signature(record['frameworkSlug']) and \
+                (client.get_framework_agreement(record['agreementId'])
+                 ['agreement']['status'] == 'signed'):
```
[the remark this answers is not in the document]
gidsg: OK I tried to improve this a bit in e6aa94b

tests/helpers/test_framework_helpers.py
gidsg: set value in 8abacb5

tests/helpers/test_framework_helpers.py
gidsg: fixed in ab1083c

dmscripts/generate_framework_agreement_signature_pages.py
gidsg: Agree, for now testing on preview/jenkins is the best way to validate this works as expected. Same with symlinking the files.

dmscripts/generate_framework_agreement_signature_pages.py
```diff
@@ -51,6 +53,7 @@ def save_page(html, supplier_id, output_dir, descriptive_filename_part):
     page_path = os.path.join(output_dir, '{}-{}.html'.format(supplier_id, descriptive_filename_part))
     with io.open(page_path, 'w+', encoding='UTF-8') as htmlfile:
         htmlfile.write(html)
+    return page_path
 
 
 def render_html_for_successful_suppliers(rows, framework, template_dir, output_dir, dry_run=False):
```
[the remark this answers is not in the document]
gidsg: Don't think so? Did you have anything specific in mind?

scripts/generate-framework-agreement-counterpart-signature-pages.py
```diff
@@ -95,17 +98,29 @@
...
+            # E-signatures are automatically approved.
+            client.approve_agreement_for_countersignature(record['agreementId'],
```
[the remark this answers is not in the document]
gidsg: Hmm, to be honest I am still not 100% sure of the logic either, but I _think_ and judging by the previous Jenkins runs. We may be protected by this line https://github.com/alphagov/digitalmarketplace-scripts/blob/27288765a66bcb50c568658c6d14a91489546fbe/dmscripts/generate_framework_agreement_signature_pages.py#L104-L105
i.e. once the `countersigned_path` is set (which should be done by the upload script) it skips over the supplier. 
We should definitely test this on preview though.

gidsg:
@phantas I agree with you that ideally we should have tests specifically for this script. However, due to the time constraints here we may have to test manually on preview and add a follow-up card for improving this before DOS-5.

scripts/generate-framework-agreement-counterpart-signature-pages.py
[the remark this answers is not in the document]
lfdebrux: @gidsg is correct, once a PDF has been uploaded the generate script will not create a new one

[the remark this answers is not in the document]
phantas: that means that if the supplier signs again, we won't countersign again (which I think was something we said it would do?). In any case, I think we can off-scope that, as it's new behaviour that wasn't there before.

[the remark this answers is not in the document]
gidsg: @phantas I _believe_ that when a countersign approval happens it changed the status from `signed` to `approved`. If they re-sign it reverts the status to `signed` which would allow the re-approval to take place.

dmscripts/generate_framework_agreement_signature_pages.py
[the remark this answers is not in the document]
phantas: I was just wondering because it makes use of lines 60 and 61 that use the template `framework-agreement-signature-page.html` and css, which, if I understand correctly, is the pre e-signature template? I may be completely way off, so I am sorry if I am wrong.

scripts/generate-framework-agreement-counterpart-signature-pages.py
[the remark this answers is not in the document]
gidsg: If they re-sign it will generate a new agreement, which will not have `countersigned_path` set.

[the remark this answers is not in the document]
phantas: ah great, thanks!

dmscripts/generate_framework_agreement_signature_pages.py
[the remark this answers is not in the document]
gidsg: @phantas good question, this was also confusing for me. I think this method was used for the previous manual signature flow. i.e. to generate the PDF that was printed and signed by the supplier.

As a follow-up. Once DOS-5 goes out we should also have a card to remove all this legacy functionality.

bjgill (approved):
Change looks likely to work. I agree with Paula that this should be tested. However, given the time constraints, I'm happy for that to be deferred for a future story.

scripts/generate-framework-agreement-counterpart-signature-pages.py
```diff
@@ -95,17 +98,29 @@
...
         rows, framework, os.path.join(args['<path_to_agreements_repo>'], 'documents', framework['slug']), html_dir
     )
-    html_pages = os.listdir(html_dir)
-    html_pages.remove('framework-agreement-signature-page.css')
-    html_pages.remove('framework-agreement-countersignature.png')
-    ok = render_pdf_for_each_html_page(html_pages, html_dir, args['<output_folder>'])
+
+    ok = True
+    for html_dir, html_pages in html_to_render:
+        ok &= render_pdf_for_each_html_page(html_pages, html_dir, args['<output_folder>']) and ok
```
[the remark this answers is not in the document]
gidsg: fixed in 3a10504

PR #544, opened 2020-09-22
filter out e-signature suppliers without an agreementID

gidsg:
In the case where a supplier had not signed the agreement and was on an e-signature framework we were erroneously looking up their agreement which would then 404 and cause an exception on the API client. This filters out suppliers without an agreement ID so that subsequent steps are not run for thm.

Co-authored-by: Laurence de Bruxelles <laurence.debruxelles@digital.cabinet-office.gov.uk>

### DIFF
--- a/dmscripts/helpers/framework_helpers.py
+++ b/dmscripts/helpers/framework_helpers.py
@@ -68,11 +68,30 @@ def find_suppliers_with_details_and_draft_service_counts(
     length = len(records)
     records = map_impl(partial(add_supplier_info, client), records)
     records = map_impl(partial(add_framework_info, client, framework_slug), records)
-    records = map_impl(partial(add_agreement_info, client), records)
     records = map_impl(partial(add_draft_counts, client, framework_slug), records)
     records = map_watch(
         records,
         f"fetched details and draft counts for supplier {{count}}/{length}"
+    )
+    return records
+
+
+def find_suppliers_with_signed_framework_agreements(
+    client,
+    framework_slug,
+    supplier_ids=None,
+    map_impl=map,
+):
+    records = find_suppliers(client, framework_slug, supplier_ids)
+    length = len(records)
+    records = map_impl(partial(add_supplier_info, client), records)
+    records = map_impl(partial(add_framework_info, client, framework_slug), records)
+    records = filter(lambda record: bool(record['agreementId']), records)
+    records = map_impl(partial(add_draft_counts, client, framework_slug), records)
+    records = map_impl(partial(add_agreement_info, client), records)
+    records = map_watch(
+        records,
+        f"fetched signed framework agreement for supplier {{count}}/{length}"
     )
     return records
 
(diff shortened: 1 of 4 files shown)


scripts/generate-framework-agreement-counterpart-signature-pages.py
```diff
@@ -97,15 +99,17 @@
...
-                                                          AUTOMATED_COUNTERSIGNING_USER_ID)
+    if framework_supports_e_signature(framework):
+        for record in records:
+            if not record["countersignedAt"]:
+                # E-signatures are automatically approved.
+                countersigned_agreement = client.approve_agreement_for_countersignature(
+                                              record['agreementId'],
+                                              "automated countersignature script",
+                                              AUTOMATED_COUNTERSIGNING_USER_ID)["agreement"]
+                record["countersignedAt"] = countersigned_agreement["countersignedAgreementReturnedAt"]
```
bjgill, line 112: Does this edit get sent back to the API?

dmscripts/helpers/framework_helpers.py
```diff
@@ -77,6 +76,26 @@ def find_suppliers_with_details_and_draft_service_counts(
...
+    supplier_ids=None,
+    map_impl=map,
+):
+    records = find_suppliers(client, framework_slug, supplier_ids)
+    length = len(records)
+    records = map_impl(partial(add_supplier_info, client), records)
+    records = map_impl(partial(add_framework_info, client, framework_slug), records)
+    records = filter(lambda record: bool(record['agreementId']), records)
+    records = map_impl(partial(add_draft_counts, client, framework_slug), records)
+    records = map_impl(partial(add_agreement_info, client), records)
```
bjgill, line 91: This is quite a lot of information to get. Do we need it all?

bjgill (commented):
Looks overall sensible. Just two questions about things I wasn't sure about.

scripts/generate-framework-agreement-counterpart-signature-pages.py
gidsg: this is setting in the `record` what the API has stored and just returned in: https://github.com/alphagov/digitalmarketplace-scripts/blob/94202543a6e883e609e84973449b4871c90ab0fa/scripts/generate-framework-agreement-counterpart-signature-pages.py#L108

bjgill: Am I misunderstanding how this works? What is `record`?

If `record` is just a dictionary, how does our edit to it get into the database?

dmscripts/helpers/framework_helpers.py
gidsg: I believe yes these values are used in the generated document. Although it is possibly something we could check in any future work on this script.

scripts/generate-framework-agreement-counterpart-signature-pages.py
gidsg: `record` is a kind of pimped list of suppliers, which gets augmented with extra details required for the document i.e. in https://github.com/alphagov/digitalmarketplace-scripts/blob/94202543a6e883e609e84973449b4871c90ab0fa/dmscripts/helpers/framework_helpers.py#L85-L96
Because this approval is an additional API call that happens later on in the process and conditionally, we have to update the record object at that point with the `countersignedAt` value, so that subsequent check in: https://github.com/alphagov/digitalmarketplace-scripts/blob/27288765a66bcb50c568658c6d14a91489546fbe/dmscripts/generate_framework_agreement_signature_pages.py#L104 passes.

Does that make more sense?

bjgill (approved):

PR #548, opened 2020-09-23
Tweak counterpart PDF script to be more communicative

lfdebrux:
It would be nice to know what it's doing even without the verbose flag.


gidsg (approved):
👍 ❇️

PR #549, opened 2020-09-23
Change agreement PDF script to be less verbose

lfdebrux:
We only want to see warnings or errors from wkhtmltopdf, we don't need the progress bars.


gidsg (approved):
🥧 👍

PR #551, opened 2020-09-23
Handle case where supplier framework agreement is in bad state

lfdebrux:
Don't stop running if a supplier framework agreement is in a bad state.


gidsg (approved):
👍

PR #552, opened 2020-09-24
Reduce impact of race condition on counterpart agreement signature job

lfdebrux:
Trello: https://trello.com/c/lPJCI12q/1443-race-condition-where-supplier-signs-between-generate-documents-and-upload-documents-scripts-running

This doesn't really fix the issue, but it should reduce the impact until we can fix it.


dmscripts/generate_framework_agreement_signature_pages.py
```diff
@@ -129,11 +129,12 @@ def render_html_for_suppliers_awaiting_countersignature(rows, framework, templat
 
         logger.info(f"generating coutersigned agreement PDF for {data['supplier_id']}")
 
-        html_dir = output_dir / str(data['supplier_id'])
+        name_prefix = f"{data['supplier_id']}-{data['agreementId']}"
```
gidsg, line 132: am getting `KeyError: 'agreementId'` exception on this line when running locally

lfdebrux:
Converted to draft while it's on the backburner.

PR #557, opened 2020-09-25
Update acknowledge publish draft services audit user

lfdebrux:
We changed the user in the publish draft services script in https://github.com/alphagov/digitalmarketplace-scripts/pull/371, we need to reflect that change in this script too.


gidsg (approved):
👍

PR #561, opened 2020-09-30
Stop printing user emails when dry running

bjgill:
Reduce the risk of accidental PII leaks if we dry-run in Jenkins.


lfdebrux (approved):
Good spot!

PR #562, opened 2020-10-01
Unsuspend services when countersigning e-signatures

gidsg:
https://trello.com/c/EQsNsY0R/399-investigate-automatically-un-suspending-suppliers-when-they-sign-agreement-2
Previously we manually unsuspended suppliers who signed the framework agreement after standstill. This made sense in the context of requiring a manual approval of the signed document, whereby the admin team could also unsuspend their services at the same time.

In e-signature flow we automatically approve the signed document, therefore it also makes sense to automatically unsuspend suppliers when approving them in this context.


scripts/generate-framework-agreement-counterpart-signature-pages.py
```diff
@@ -131,7 +136,7 @@ def approve_supplier_framework_agreement(record):
                         record["countersignedAt"] = countersigned_agreement["countersignedAgreementReturnedAt"]
                     except dmapiclient.errors.HTTPError as e:
                         logger.warn(f"failed to countersign agreement {agreement_id} for supplier {supplier_id}: {e}")
-
+                unsuspend_suspended_supplier_services(record, AUTOMATED_SUSPENDING_USER, client, logger, dry_run)
```
bjgill, line 139: What happens if countersigning has failed for this supplier?

dmscripts/helpers/supplier_data_helpers.py
```diff
@@ -336,3 +338,48 @@ def country_code_to_name(country_code):
...
+        event["data"]["serviceId"]
+        for event in
+        client.find_audit_events_iter(
+            audit_type=AuditTypes.update_service_status,
+            data_supplier_id=supplier_id,
+            user=suspending_user
+        )
+    )
+    service_ids = suspended_services_on_framework & services_suspended_by_script
+    # Suspend all services for supplier (the API will de-index the services from search results)
```
bjgill, line 377: Is this true?

dmscripts/helpers/supplier_data_helpers.py
```diff
@@ -336,3 +338,48 @@ def country_code_to_name(country_code):
         return record[code]['item'][0]['name']
 
     raise requests.exceptions.RequestException(register_response)
+
+
+def unsuspend_suspended_supplier_services(record, suspending_user, client, logger, dry_run):
+    """
+    This method checks if the supplier has services suspended by the suspending_user, if so it un-suspends them. It
+    returns None.
```
bjgill, line 346: Given that you've got `:return: None` below, do we need `It returns None` as well?

bjgill (commented):
Looks sensible. Just two questions about comment and another about behaviour

scripts/generate-framework-agreement-counterpart-signature-pages.py
gidsg: Good question I guess if approving fails we should also not unsuspend, as this will leave the supplier in a cleaner state for the subsequent run. In that case I can move them inside the try for approval.

dmscripts/helpers/supplier_data_helpers.py
gidsg: Good question. I am not 100% sure as I took this comment from @lfdebrux's branch https://github.com/alphagov/digitalmarketplace-scripts/blob/83f559f72cc2d04ff2b07de022314e6d04638230/dmscripts/suspend_suppliers_without_agreements.py#L53
A guess @lfdebrux may have more context about where this happens.

bjgill: That comment is present in both `suspend_supplier_services` and `unsuspend_supplier_services`. Is it perhaps a copy-paste error and doesn't make sense here (given that we're not suspending anything)?

gidsg: Good spot. Fixed in 5a7f8ff. I think the edited comment hopefully makes sense as the re-index would need to start indexing these services.

bjgill (approved):

PR #564, opened 2020-10-09
Fix logging for framework application stats snapshot script

lfdebrux:
I noticed that when this script runs we get unformatted log messages such as

    INFO:dmapiclient.base:API {api_method} request on {api_url} finished in {api_time}

See https://ci.marketplace.team/view/Framework%20lifecycle%20jobs/job/hourly-stats-snapshot-digital-outcomes-and-specialists-5-production/27/console for an example of this.

I think this is because we are using the `logging` module directly rather than the dmscripts logging configuration. This commit changes the script to use `dmscripts.helpers.logging.configure_logger`.

### DIFF
--- a/scripts/framework-applications/snapshot-framework-stats.py
+++ b/scripts/framework-applications/snapshot-framework-stats.py
@@ -13,7 +13,6 @@ Example:
 
 from docopt import docopt
 import json
-import logging
 import sys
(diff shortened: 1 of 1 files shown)


gidsg (approved):
👍 🌴

PR #566, opened 2020-10-13
Remove DOS4 questions from DOS5 drafts

alex9smith:
Script to update services from this ticket

https://trello.com/c/lNqAe34j/68-fix-dos-5-draft-services-with-improperly-copied-dos-4-answers

Some questions were removed from DOS5 but not flagged to not be copied from a DOS4 service. Affected draft services would always fail validation because they had answers to questions which don't exist.

This script finds affected draft services and deletes the answers to the removed questions.

### DIFF
--- a/scripts/oneoff/remove-dos4-questions-from-dos5-drafts.py
+++ b/scripts/oneoff/remove-dos4-questions-from-dos5-drafts.py
@@ -12,7 +12,9 @@ from typing import List
 
 
 def get_affected_drafts_services(api_client: DataAPIClient) -> List[dict]:
-    all_drafts = api_client.find_draft_services_by_framework('digital-outcomes-and-specialists-5')["services"]
+    all_drafts = api_client.find_draft_services_by_framework(
+        'digital-outcomes-and-specialists-5', status='not-submitted'
+    )["services"]
 
     return [draft for draft in all_drafts if draft_service_contains_dos4_answer(draft)]
 
@@ -48,10 +50,13 @@ def draft_service_contains_dos4_answer(draft: dict) -> bool:
         "designerAccessibleApplications"
     }
 
-    return len(invalid_answers.union(set(draft.keys()))) > 0 and draft['status'] == 'not-submitted'
+    return invalid_answers.intersection(draft)
 
 
 def remove_dos4_answers(api_client: DataAPIClient, draft: dict, developer_email: str) -> dict:
+    print(f"Supplier id: {draft['supplierId']}")
+    print(f"Draft id: {draft['id']}")
+
     return api_client.update_draft_service(
         draft['id'],
         {
@@ -91,7 +96,7 @@ if __name__ == '__main__':
     parser = argparse.ArgumentParser()
 
     parser.add_argument("stage", type=str)
-    parser.add_argument("developer_email", type=str)
+    parser.add_argument("--updated-by", type=str, dest="developer_email")
 
     args = parser.parse_args()
 
@@ -104,7 +109,3 @@ if __name__ == '__main__':
         remove_dos4_answers(data, draft, args.developer_email) for draft
         in get_affected_drafts_services(data)
     ]
-
-    for draft in updated_drafts:
-        print(f"Supplier id: {draft['supplierId']}")
-        print(f"Draft id: {draft['id']}")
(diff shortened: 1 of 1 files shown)


bjgill:
Did a dry-run of this on preview, and it picked up services to fix.

My plan is to run this first on preview. If that's successful, I'll then run this against staging and production.

scripts/oneoff/remove-dos4-questions-from-dos5-drafts.py
```diff
@@ -0,0 +1,110 @@
...
+
+sys.path.insert(0, '.')
+from dmapiclient import DataAPIClient
+from dmscripts.helpers.auth_helpers import get_auth_token
+from dmutils.env_helpers import get_api_endpoint_from_stage
+
+from typing import List
+
+
+def get_affected_drafts_services(api_client: DataAPIClient) -> List[dict]:
```
lfdebrux, line 14: Typing, nice!

scripts/oneoff/remove-dos4-questions-from-dos5-drafts.py
```diff
@@ -0,0 +1,110 @@
...
+        "developerAccessibleApplications",
+        "contentDesignerAccessibleApplications",
+        "qualityAssuranceAccessibleApplications",
+        "technicalArchitectAccessibleApplications",
+        "webOperationsAccessibleApplications",
+        "serviceManagerAccessibleApplications",
+        "designerAccessibleApplications"
+    }
+
+    return len(invalid_answers.union(set(draft.keys()))) > 0 and draft['status'] == 'not-submitted'
```
lfdebrux, line 51: I like the use of sets here! I'm not sure you need to do `set(draft.keys())` though, because drafts should be a `set`  already 😁  [[1](https://renzolucioni.com/pythons-set-literals/)]

Oh and while we're at it, the use of `len` can also be removed, because any non-empty set is truthy [[2](https://docs.python.org/3.6/library/stdtypes.html#truth-value-testing)].

---

However, I'm not sure the union is what we want here?

> Return a new set with elements from the set and all others.
https://docs.python.org/3.6/library/stdtypes.html#frozenset.union

scripts/oneoff/remove-dos4-questions-from-dos5-drafts.py
```diff
@@ -0,0 +1,110 @@
...
+            "webOperationsAccessibleApplications": None,
+            "serviceManagerAccessibleApplications": None,
+            "designerAccessibleApplications": None
+        },
+        developer_email
+    )
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser()
```
lfdebrux, line 91: It doesn't matter for one-off scripts, but I thought I would flag that in the scripts repo we usually use docopt.org instead of `argparse`

scripts/oneoff/remove-dos4-questions-from-dos5-drafts.py
```diff
@@ -0,0 +1,110 @@
...
+
+    parser.add_argument("stage", type=str)
+    parser.add_argument("developer_email", type=str)
```
lfdebrux, line 94: Can we make this `--updated-by=<developer_email>` to be consistent with other scripts?

lfdebrux (commented):
Looks really nice, a few comments and one thing I'm a bit unsure about

scripts/oneoff/remove-dos4-questions-from-dos5-drafts.py
alex9smith: Ah thanks, I didn't realise that passing a dict into `set` would just take the keys and good point on the truthiness too

You're right, union is definitely wrong 😂  We want the set of values that are common to both sides which is intersection

scripts/oneoff/remove-dos4-questions-from-dos5-drafts.py
```diff
@@ -0,0 +1,111 @@
...
+        "developerAccessibleApplications",
+        "contentDesignerAccessibleApplications",
+        "qualityAssuranceAccessibleApplications",
+        "technicalArchitectAccessibleApplications",
+        "webOperationsAccessibleApplications",
+        "serviceManagerAccessibleApplications",
+        "designerAccessibleApplications"
+    }
+
+    return invalid_answers.intersection(draft)
```
lfdebrux, line 53: We should coerce this to `bool` to make the typing valid

lfdebrux (approved):

scripts/oneoff/remove-dos4-questions-from-dos5-drafts.py
bjgill:
```suggestion
    return bool(invalid_answers.intersection(draft))
```

PR #573, opened 2020-10-27
Add script to scan supplier PDFs for JS content

alex9smith:
Add a script that scans supplier uploaded PDFs stored in S3 for JS or other unusual content types.

Part of ticket https://trello.com/c/SIYebkbk/237-3-scan-service-docs-and-produce-list-of-services-with-non-conforming-pdfs

### DIFF
--- a/scripts/scan-pdf-content.py
+++ b/scripts/scan-pdf-content.py
@@ -1,0 +1,336 @@
+#!/usr/bin/env python3
+"""
+Scan user uploaded PDFs in S3 and flag ones which contain any
+non-text or image content (ie. JS or videos).
+
+Requires a running version of veraPDF-REST (https://github.com/veraPDF/veraPDF-rest)
+
+Usage: scan-pdf-content.py <stage> <framework> <verapdf-url> [--index=<index>]
+
+Options:
+    <stage>                         Stage to target
+    <framework>                     Slug for the framework to scan
+    <verapdf-url>                   The url for the verapdf-rest service to use
+    --index=<index>                 Optional. If supplied, the scan resumes from a previous run, using
+                                        the list of PDFs stored in '/tmp/pdf-scan-queue.json' starting at `index`
+
+    -h, --help                      Show this information
+
+Examples:
+    ./scripts/scan-pdf-content.py staging g-cloud-12 http://localhost:8080
+"""
+import boto3
+import requests
+import csv
+import uuid
+import logging
+import json
+import os
+import threading
+import queue
+from docopt import docopt
+from datetime import datetime
+
+from typing import Tuple, NamedTuple, Optional
+
+
+class ScanResult(NamedTuple):
+    scanned_at: str
+    status_code: int
+    message: str
+    bucket: str
+    framework: str
+    key: str
+
+
+def get_aws_session(stage: str) -> boto3.session.Session:
+    """
+    Start an authenticated session for the supplied stage.
+
+    :param stage: The stage we need credentials for. Must be one of 'production',
+        'staging' or 'preview'.
+    :return: An authenticated boto3.session.Session object
+    """
+    # production and staging data is in the production AWS account
+    if stage == "preview":
+        arn = "arn:aws:iam::381494870249:role/infrastructure"
+    else:
+        arn = "arn:aws:iam::050019655025:role/infrastructure"
+
+    sts_client = boto3.client("sts")
+    session_name = f"PDFContentScan-{str(uuid.uuid4())}"
+    sts_response = sts_client.assume_role(RoleArn=arn, RoleSessionName=session_name)
+
+    return boto3.session.Session(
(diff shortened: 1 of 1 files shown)


scripts/scan-pdf-content.py
```diff
@@ -0,0 +1,336 @@
...
+
+    :param stage: The stage we need credentials for. Must be one of 'production',
+        'staging' or 'preview'.
+    :return: An authenticated boto3.session.Session object
+    """
+    # production and staging data is in the production AWS account
+    if stage == "preview":
+        arn = "arn:aws:iam::381494870249:role/infrastructure"
+    else:
+        arn = "arn:aws:iam::050019655025:role/infrastructure"
```
alex9smith, line 58: Are these the best roles to use for a developer to access S3? I got a permission error when I tried to use the `developer` role, but that might be caused by the limited permissions on my account. 

I looked in the `dm-aws` repo but couldn't see where the policy for the `developer` role was defined to check.

gidsg: I think normally we don't hard code these, instead we set `AWS_PROFILE` env var to direct which account to use.

alex9smith: Ah I think I understand. That'll clean up the code quite a bit if I don't need to pass the authenticated `session` object around. Thanks!

scripts/scan-pdf-content.py
```diff
@@ -0,0 +1,300 @@
...
+import csv
+import logging
+import json
+import os
+import threading
+import queue
+from docopt import docopt
+from datetime import datetime
+
+from typing import Tuple, NamedTuple, Optional
```
lfdebrux, line 36: It's considered [good style](https://google.github.io/styleguide/pyguide.html#313-imports-formatting) to group imports by type, so standard library imports first, then third-party libraries, then our own libraries last. (I realise we don't follow this rule everywhere in our own code, so sorry for picking you up on it!).

```suggestion
import csv
import json
import logging
import os
import threading
import queue
from datetime import datetime
from typing import Tuple, NamedTuple, Optional

import boto3
import requests
from docopt import docopt
```

scripts/scan-pdf-content.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+class ScanResult(NamedTuple):
+    scanned_at: str
+    status_code: int
+    message: str
+    bucket: str
+    framework: str
+    key: str
```
lfdebrux, line 45: Nice!

scripts/scan-pdf-content.py
```diff
@@ -0,0 +1,300 @@
...
+
+def list_pdfs_in_bucket(bucket_name: str, framework: str,) -> list:
+    """
+    Return a list of `ObjectSummary`s for PDFs in the supplied S3 bucket.
+
+    :param bucket_name: The name of the bucket to check
+    :param framework: The name of the framework to scan for
+    :return: A list of S3 objects
+    """
+    s3 = boto3.resource("s3")
```
lfdebrux, line 56: Do we need to include the region name here? 🤔 Genuinely not sure.

scripts/scan-pdf-content.py
```diff
@@ -0,0 +1,300 @@
...
+    :param index: An optional index to resume from. If supplied, a list of PDFs to scan is read from
+                    '/tmp/pdf-scan-queue.json' and the scan is started from position `index`.
+    :return: None
+    """
+    if index is not None:
+        # Resuming from a previous run. Load files to scan from disk and set up index
+        logging.info("Resuming from previous run")
+        pdfs_to_scan = load_pdfs_to_scan_from_file()
+
+        logging.info(f"Resuming scan from position: {index}")
```
lfdebrux, line 231: How will we know what index to start from if resuming?

scripts/scan-pdf-content.py
```diff
@@ -0,0 +1,300 @@
...
+    Create an empty CSV with headers to store results
+
+    :param filename: The name of the output file to create
+    :returns: None
+    """
+    with open(filename, "w") as f:
+        report_writer = csv.writer(f, delimiter=",")
+        report_writer.writerow(
+            ["scanned_at", "bucket", "framework", "key", "status_code", "message"]
+        )
```
lfdebrux, line 274: It might be good to move this further up the file to be close to `write_result`, so you can read them together and more quickly understand how they work together!

lfdebrux (approved):
Looks good to me, just a few stylistic suggestions, feel free to ignore!

scripts/scan-pdf-content.py
alex9smith: It must pick up the region from the auth & profile you use, there's nothing in [the docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#service-resource) about region and the script runs fine

lfdebrux: Awesome 👍

PR #576, opened 2020-11-03
Add --resume to PDF scan script

alex9smith:
Replace the `--index` optional argument with `--resume` which calculates the index to resume from the number of rows in the report CSV. 

Part of https://trello.com/c/SIYebkbk/237-3-scan-service-docs-and-produce-list-of-services-with-non-conforming-pdfs


lfdebrux (approved):
Neat!

PR #580, opened 2020-11-16
Change notify scripts to fail fast if there is a template error

lfdebrux:
For Notify batch scripts an issue with a template such as a missing personalisation will cause all emails that would be sent by the script to fail. In this situation the script should fail fast, rather than try and continue and waste cycles.

This also reduces the possibility of https://trello.com/c/kwVKRjEm/1460-jenkins-job-notify-suppliers-of-framework-application-event-production-keeps-running-even-after-being-aborted happening again, because the script should hopefully stop itself instead of having to be killed.

### DIFF
--- a/dmscripts/notify_buyers_to_award_closed_briefs.py
+++ b/dmscripts/notify_buyers_to_award_closed_briefs.py
@@ -2,7 +2,7 @@
 from datetime import datetime, date, timedelta
 
 import dmapiclient
-from dmutils.email.exceptions import EmailError
+from dmutils.email.exceptions import EmailError, EmailTemplateError
 from dmutils.formats import DATE_FORMAT
 
 from dmscripts.helpers import logging_helpers, brief_data_helpers
@@ -72,6 +72,8 @@ def send_email_to_brief_user_via_notify(notify_client, notify_template_id, user,
                 notify_client.send_email(
                     user['emailAddress'], notify_template_id, email_context_data, allow_resend=False
                 )
+        except EmailTemplateError:
+            raise  # do not try to continue
         except EmailError:
             return user['id']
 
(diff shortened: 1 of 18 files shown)


dmscripts/notify_buyers_to_award_closed_briefs.py
```diff
@@ -72,6 +72,8 @@ def send_email_to_brief_user_via_notify(notify_client, notify_template_id, user,
                 notify_client.send_email(
                     user['emailAddress'], notify_template_id, email_context_data, allow_resend=False
                 )
+        except EmailTemplateError:
+            raise  # do not try to continue
```
bjgill, line 76: Why do we need to explicitly re-raise this?

dmscripts/notify_buyers_when_requirements_close.py
```diff
@@ -49,6 +49,9 @@ def notify_users(email_api_key, stage, brief):
                 extra={'error': e, 'brief_id': brief['id']}
             )
 
+            if isinstance(e, EmailTemplateError):
```
bjgill, line 52: When will this happen? Looking in dmutils, it doesn't look as if `EmailTemplateError` is a subclass of `EmailError`?

bjgill (commented):
Like the change in dmutils. A couple of questions about things I don't quite understand.

dmscripts/notify_buyers_when_requirements_close.py
lfdebrux: Oh farts. That was a mistake! Thanks for spotting. I think will update utils to reflect my original intention.

dmscripts/notify_buyers_to_award_closed_briefs.py
lfdebrux: Because it's a subclass of EmailError, if we don't catch and re-raise it will be handled by the EmailError handler.

dmscripts/notify_buyers_when_requirements_close.py
lfdebrux: Fixed in alphagov/digitalmarketplace-utils#588

lfdebrux:
Added tests

bjgill (approved):
:tada:

PR #586, opened 2020-11-23
Add script to export details about service edits

lfdebrux:
Occasionally we get asked to provide details about historical changes to a service. This script should make answering those requests a little easier.


gidsg (approved):
Nice one 👍 🧷

PR #588, opened 2020-12-04
Switch from Travis CI to Github Actions

alex9smith:
https://trello.com/c/PKG5QnqU/628-05-migrate-scripts-ci-from-travis-to-github-actions

Ubuntu dependencies from Ubuntu dependencies from https://github.com/alphagov/digitalmarketplace-supplier-frontend/pull/1289
Python test template from https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python

Once this is approved I'll switch off the merge requirements for Travis and switch them on for Github actions


gidsg:
We might want to add a caching action similar to https://github.com/alphagov/digitalmarketplace-supplier-frontend/blob/2acb5fbd8a52d16bb8ff3a65e5ec5245a8af57a3/.github/workflows/test.yml#L33-L37

alex9smith:
> We might want to add a caching action

Ooh very cool. I'll do that!

.github/workflows/test.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Setup Python cache
+        uses: actions/cache@v2
+        with:
+          path: venv
+          key: venv-${{ runner.os }}-${{ steps.full-python-version.outputs.version }}-${{ hashFiles('**/requirements-dev.txt') }}
```
gidsg, line 25: Looking at the generated cache key, I don't think it is resolving the second variable. Could we try `${{ matrix.python-version }}` instead?

alex9smith: Sure. Looks like we might need to do the same to the supplier frontend action as well

alex9smith: This looks much better `Cache not found for input keys: venv-Linux-3.7-925d5c0df96c957f702d7e25a0e6eeeb79003666a14110b4f25509d648cc908c`

gidsg (approved):

PR #570, opened 2020-10-19
Add timezone to timestamp logging format

alex9smith:
Add the timezone identifier to timestamps when logging from scripts. This will make it easier to consolidate logs across multiple events or scripts.

Closes ticket https://trello.com/c/wRGIKhhf/1091-05-add-timezone-identifier-to-timestamps-in-logging-format-of-scripts

PR #590, opened 2020-12-15
Remove ',uuu' from logs

bjgill:
We [hoped](https://github.com/alphagov/digitalmarketplace-scripts/pull/570) it would give us the ~microseconds~ milliseconds, but it didn't.


lfdebrux:
If you want microseconds you want `%S.%f` no?

lfdebrux (approved):

bjgill:
Oops - I meant milliseconds. I think microseconds would be a bit excessive. I don't want to spend more time on this, so will merge as-is.

PR #594, opened 2020-12-21
Add admin-only CSV with buyer user details to DOS opportunities export script

lfdebrux:
Ticket: https://trello.com/c/s5IrVY70/2122-ccsrequest-queries-relating-to-the-digital-marketplace-for-dos-team

Currently CSS admins have to cross-reference buyer user details with an opportunity by hand, which is time-consuming for them.

This commit expands the export-dos-opportunities script to generate an additional CSV with extra details about the buyer user. As we don't want to make this information publicly available, this new CSV is uploaded to the reports bucket with private permissions; a change to the admin frontend will be needed to allow admins to download this file.

### DIFF
--- a/dmscripts/export_dos_opportunities.py
+++ b/dmscripts/export_dos_opportunities.py
@@ -1,6 +1,8 @@
 import csv
+from collections import OrderedDict
 from datetime import datetime
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


gidsg (approved):
nice one 👍 🌮

PR #597, opened 2020-12-29
Fix bugs in export-dos-opportunities script

lfdebrux:
* Log messages now show the key for the uploaded file
* Admin-only CSV has the correct permissions to allow download via signed url

There is more detail in the commit message.

### DIFF
--- a/dmscripts/export_dos_opportunities.py
+++ b/dmscripts/export_dos_opportunities.py
@@ -127,13 +127,13 @@ def upload_file_to_s3(
     logger,
 ):
     with open(file_path, 'br') as source_file:
-        acl = "public-read" if public else "private"
+        acl = "public-read" if public else "bucket-owner-full-control"
 
-        logger.info("{}UPLOAD: {} to {}::{} with acl {}".format(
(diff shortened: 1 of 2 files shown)


dmscripts/export_dos_opportunities.py
```diff
@@ -127,13 +127,13 @@ def upload_file_to_s3(
     logger,
 ):
     with open(file_path, 'br') as source_file:
-        acl = "public-read" if public else "private"
+        acl = "public-read" if public else "bucket-owner-full-control"
```
domoscargin, line 130: Presumably there are no cases in which "private" might be appropriate here?

domoscargin (approved):
LGTM

dmscripts/export_dos_opportunities.py
lfdebrux: I don't think so, at least not for the export-dos-opportunities script. This function `upload_file_to_s3` is used only in this module, so it's scope is pretty limited. It would be nice if there was a scripts repo wide `upload_file_to_s3` function, but I decided not to do that today...

PR #598, opened 2020-12-31
add contract title to counterpart notification email

gidsg:
https://trello.com/c/k2lfAXGt/651-2-prepare-e-signature-related-emails
Pass contract title to counterpart notification email.

(Note this will require corresponding changes in the Jenkins job which I will pick up in a subsequent PR once approved.)

### DIFF
--- a/dmscripts/upload_counterpart_agreements.py
+++ b/dmscripts/upload_counterpart_agreements.py
@@ -18,6 +18,7 @@ def upload_counterpart_file(
     file_path,
     dry_run,
     data_api_client,
+    contract_title,
     dm_notify_client=None,
     notify_template_id=None,
     notify_fail_early=True,
@@ -83,6 +84,7 @@ def upload_counterpart_file(
                         "framework_slug": framework["slug"],
                         "framework_name": framework["name"],
                         "supplier_name": supplier_name,
+                        "contract_title": contract_title
                     }, allow_resend=True)
                     logger.debug(f"NOTIFY: sent email to supplier '{supplier_id}' user {hash_string(notify_email)}")
                 else:
--- a/scripts/upload-counterpart-agreements.py
+++ b/scripts/upload-counterpart-agreements.py
@@ -29,7 +29,7 @@ This will:
    * if <notify_key> and <notify_template_id> are provided, will send a notification email to all the supplier's active
      users (and their primaryContactEmail for that framework)
 
-Usage: scripts/upload-counterpart-agreements.py <stage> <documents_directory> <framework_slug> [options]
+Usage: scripts/upload-counterpart-agreements.py <stage> <documents_directory> <framework_slug> <content_path> [options]
 
 Options:
     --dry-run                                   # Don't actually do anything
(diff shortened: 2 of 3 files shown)


dmscripts/upload_counterpart_agreements.py
```diff
@@ -18,6 +18,7 @@ def upload_counterpart_file(
     file_path,
     dry_run,
     data_api_client,
+    contract_title,
```
lfdebrux, line 21: Rather than passing just the contact title why not pass the content loader object? I think that would make it easier if we need more stuff from the content loader later on.

gidsg: @lfdebrux I thought about this but it is inefficient as we then have to re-load this value for every file in the batch. Also simpler this way if we decide the value should live somewhere other than content loader in the future.

lfdebrux: Okay, makes sense.

scripts/upload-counterpart-agreements.py
```diff
@@ -72,8 +73,14 @@
...
     stage = arguments['<stage>']
     data_api_client = DataAPIClient(get_api_endpoint_from_stage(stage), get_auth_token('api', stage))
-    framework = data_api_client.get_framework(arguments['<framework_slug>'])["frameworks"]
+    framework_slug = arguments['<framework_slug>']
+    framework = data_api_client.get_framework(framework_slug)["frameworks"]
     document_directory = arguments['<documents_directory>']
+    content_path = arguments['<content_path>']
+    content_loader = ContentLoader(content_path)
+    content_loader.load_messages(framework_slug, ['e-signature'])
+    contract_title = content_loader.get_message(framework_slug, 'e-signature', 'framework_contract_title')
```
lfdebrux, line 82: What happens if we need to upload a counterpart agreement for an old framework that doesn't have this message? 🤔 Is there a default we can set?

gidsg: @lfdebrux Oh, good point, I'm not sure if we are still signing for G11 and back but I will set a default argument just in case.

gidsg: @lfdebrux addressed in c5b214a

scripts/upload-counterpart-agreements.py
```diff
@@ -72,8 +73,17 @@
 
     stage = arguments['<stage>']
     data_api_client = DataAPIClient(get_api_endpoint_from_stage(stage), get_auth_token('api', stage))
-    framework = data_api_client.get_framework(arguments['<framework_slug>'])["frameworks"]
+    framework_slug = arguments['<framework_slug>']
+    framework = data_api_client.get_framework(framework_slug)["frameworks"]
     document_directory = arguments['<documents_directory>']
+    content_path = arguments['<content_path>']
+    content_loader = ContentLoader(content_path)
+    if framework['isESignatureSupported']:
```
lfdebrux, line 81: Nice!

lfdebrux (approved):

PR #605, opened 2021-01-08
Scripts to identify and delete Trello CSV attachments containing PII

bjgill:
Trello: https://trello.com/c/tq75g1XZ/2148-what-do-we-do-about-historic-lists-of-user-emails-stored-in-trello

Some of our CSV attachments in Trello contain user data (emails, mostly). We want to find and delete all such attachments.

Create a script to get us all our Trello cards with CSV attachments. This needs to be run by someone with access to all our Trello teams/boards. They'll need to get a Trello API key and token and provide them as environment variables.

Create a second script that reads the list of attachments and presents them for a human to determine whether they contain PII. If the human says yes, the script deletes the attachment.

Note - this is all quite rough and of generally poor quality. However, we're only going to run this once, so I think that's OK

### DIFF
--- a/scripts/oneoff/delete_trello_pii.py
+++ b/scripts/oneoff/delete_trello_pii.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python
+"""
+Process the list of Trello attachments produced by get_trello_attachments.py. Delete all attachments that contain
+personal information.
+"""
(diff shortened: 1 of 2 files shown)


lfdebrux (approved):
Smart 🧠

PR #606, opened 2021-01-08
Change script that suspends suppliers to also send a reminder email

alex9smith:
Previously this script suspended suppliers who have not signed the framework agreement and exported a CSV of email addresses to notify.
These email addresses were then uploaded to Notify and a reminder email was sent manually.

In the interests of saving time, reducing the scope for errors and not having personal data on our laptops, modify the script so it also sends the emails.

I have checked to make sure this script isn't used by any Jenkins jobs.

https://trello.com/c/ieH4CTed/699-1-make-script-to-suspend-suppliers-who-havent-signed-also-send-the-email

[Notify email template](https://www.notifications.service.gov.uk/services/95316ff0-e555-462d-a6e7-95d26fbfd091/templates/f4224b66-42cc-45c3-b55a-2cf5cb95792f)
Email sent:
image

### DIFF
--- a/scripts/framework-applications/suspend-suppliers-without-agreements.py
+++ b/scripts/framework-applications/suspend-suppliers-without-agreements.py
@@ -7,38 +7,41 @@ supplier will be unable to sell their services. We should therefore temporarily 
 so that buyers won't see them in search results (G-Cloud) or the supplier cannot apply to an opportunity (DOS).
 Suppliers who have returned an incorrect agreement file ('on-hold') should not be suspended.
 
-This script carries out the same action that a CCS Category user could perform in the admin, but in bulk.
+This script suspends suppliers who have not signed their agreement and sends them a reminder email.
 In the past CCS has provided a list of supplier IDs that should be suspended, however this script can also
 suspend all suppliers who have not returned a framework agreement.
 
 Usage:
     scripts/framework-applications/suspend-suppliers-without-agreements.py
         [-v...] [options]
-        <stage> <framework> <output_dir>
+        <stage> <framework> <notify_api_key> <frameworks_path>
         [--supplier-id=<id>... | --supplier-ids-from=<file>]
     scripts/framework-applications/suspend-suppliers-without-agreements.py (-h | --help)
 
 Options:
     <stage>                     Environment to run script against.
     <framework>                 Slug of framework to generate agreements for.
-    <output_dir>                Output folder for list of email addresses to send notifications to.
+    <notify_api_key>            API key for GOV.UK Notify.
+    <frameworks_path>           Path to digitalmarketplace-frameworks repository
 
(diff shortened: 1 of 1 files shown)


scripts/framework-applications/suspend-suppliers-without-agreements.py
```diff
@@ -88,25 +103,30 @@
...
+        framework_info = client.get_supplier_framework_info(supplier_id, framework_slug)
 
+        logger.info(f"{prefix}Suspending services for supplier {supplier_id}")
+        if not dry_run:
             # Do the suspending
             suspended_service_count = suspend_supplier_services(
                 client, logger, framework_slug, supplier_id, framework_info, dry_run
             )
 
-            if suspended_service_count == 0:
```
alex9smith, line 106: I got rid of this block because I can't see how it does anything. If I'm wrong let me know and I'll put it back

scripts/framework-applications/suspend-suppliers-without-agreements.py
```diff
@@ -88,25 +102,30 @@
...
-            supplier_id = supplier["supplierId"]
-            framework_info = client.get_supplier_framework_info(supplier_id, framework_slug)
+    for supplier in suppliers:
+        supplier_id = supplier["supplierId"]
+        framework_info = client.get_supplier_framework_info(supplier_id, framework_slug)
 
+        logger.info(f"{prefix}Suspending services for supplier {supplier_id}")
+        if not dry_run:
             # Do the suspending
             suspended_service_count = suspend_supplier_services(
```
gidsg, line 115: It doesn't appear we are using this variable assignment? Can we remove it?

alex9smith: Good point. It was logged before, I must have deleted a bit too much. I'll add that logging back in; I think it was useful

scripts/framework-applications/suspend-suppliers-without-agreements.py
```diff
@@ -88,25 +102,31 @@
...
             )
-
-            if suspended_service_count == 0:
-                # We should have logged why already
-                continue
-
-            # Compile a list of email addresses for the supplier (to be sent via Notify) and add to the CSV
-            for supplier_email in get_all_email_addresses_for_supplier(client, framework_info):
-                writer.writerow([supplier_email, supplier_id, suspended_service_count])
+            logger.info(f"{prefix}Suspended {suspended_service_count} services for supplier {supplier_id}")
```
gidsg, line 118: Do we only want to log this where `suspended_service_count` is >0?

alex9smith: I think we also want to log a warning if we do somehow suspend 0 services, as that means something's gone wrong

gidsg (approved):
LGTM 👍

PR #610, opened 2021-01-15
Inform users when the G12 recovery process opens 

bjgill:
Trello: https://trello.com/c/0g9FxjRJ/680-1-as-a-developer-i-can-run-a-script-to-let-all-recovery-users-know-the-recovery-process-is-open

We have drafted an email we're going to send to users when the G12 recovery process opens. The email will tell the users what they need to do. This script will be used to send that email.

Add some helpers to make this a bit easier. 

The Notify template is: https://www.notifications.service.gov.uk/services/95316ff0-e555-462d-a6e7-95d26fbfd091/templates/347e1ed7-ec83-45a0-bb16-832f244f8919. Note that we're currently missing the closing date. There's a story that tracks getting that filled in once we know the dates.

### DIFF
--- a/scripts/oneoff/open_g12_recovery_for_suppliers.py
+++ b/scripts/oneoff/open_g12_recovery_for_suppliers.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python
+"""
+Inform suppliers involved in the G12 recovery that the process is now open for them.
+
+Usage:
+    scripts/oneoff/open_g12_recovery_for_suppliers.py <stage> <notify_api_key> [--dry-run]
+
+Parameters:
+    <stage>                     Environment to run script against.
+    <notify_api_key>            API key for GOV.UK Notify.
+
+Options:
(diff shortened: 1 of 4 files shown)


scripts/oneoff/open_g12_recovery_for_suppliers.py
```diff
@@ -0,0 +1,67 @@
...
+        auth_token=get_auth_token("api", stage),
+    )
+
+    user_emails = [
+        supplier_user
+        for supplier_id in get_g12_suppliers(stage)
+        for supplier_user in get_email_addresses_for_supplier(api_client, supplier_id)
+    ]
+
+    print(user_emails)
```
gidsg, line 55: Do we want to print these? As it would leak PII into logs?

bjgill: Oops - good catch, thanks.

gidsg (approved):

PR #616, opened 2021-01-21
Get one draft ID for each G12 recovery service

bjgill:
Trello: https://trello.com/c/13Ryt5o5

As part of the G12 recovery, we need a list of the IDs of the recovery drafts. This will allow us to hide irrelevant drafts to avoid confusion.

We have a CSV containing the un-normalised service names. For each, get all drafts for that supplier that have a matching service name when normalised. Output the same CSV with the new data in a new column.

Then create new empty drafts for any missing drafts. This depends on https://github.com/alphagov/digitalmarketplace-api/pull/1083

Some services have multiple drafts, which is not what we want. So choose the draft with the most completed answers.

My plan is to merge this, run the scripts, and then update the G12 recovery spreadsheet with the results.

### DIFF
--- a/scripts/oneoff/get_g12_recovery_services.py
+++ b/scripts/oneoff/get_g12_recovery_services.py
@@ -21,12 +21,15 @@ from dmscripts.helpers.auth_helpers import get_auth_token
 def _normalise_service_name(name):
     """
     To be able to safely compare service names. The output is not meant to be human-readable.
+
+    Copied from `upload_draft_service_pdfs.py`
     """
     return (
         name.lower()
         .replace("-", "")
+        .replace("_", "")
+        .replace(":", "")
         .replace(" ", "")
-        .replace("/", ":")
         .replace(".csv", "")
     )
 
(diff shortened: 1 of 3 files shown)


scripts/oneoff/get_g12_recovery_services.py
```diff
@@ -0,0 +1,74 @@
...
+
+def _normalise_service_name(name):
+    """
+    To be able to safely compare service names. The output is not meant to be human-readable.
+    """
+    return (
+        name.lower()
+        .replace("-", "")
+        .replace(" ", "")
+        .replace("/", ":")
```
agalamatis, line 29: I think the colon should be blank.

scripts/oneoff/get_g12_recovery_services.py
```diff
@@ -0,0 +1,74 @@
...
+from dmapiclient import DataAPIClient
+from dmutils.env_helpers import get_api_endpoint_from_stage
+from docopt import docopt
+
+sys.path.insert(0, ".")
+
+from dmscripts.helpers.auth_helpers import get_auth_token
+
+
+def _normalise_service_name(name):
```
agalamatis, line 21: You might need more rules, underscore for example.

agalamatis (approved):
Check my comments, otherwise looks good.

scripts/oneoff/get_g12_recovery_services.py
bjgill: I'd copied from another script that was less comprehensive. I've now updated the normalisation to follow the script you linked. As a result, it found a few new matching drafts.

PR #617, opened 2021-01-22
add csvs and pdfs to .gitignore

gidsg:
https://trello.com/c/0B9MjEdE/738-05-prevent-people-from-accidentally-committing-pdfs-and-csvs
We want to make it more difficult to accidentally commit PII as this repo contains many scripts which output to disk.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,8 @@ venv*
 
 # local environment customizations
 local.nix
+
+# Make it more difficult to accidentally commit PII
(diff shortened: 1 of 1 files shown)


agalamatis (approved):
Why not include odt/ods too?

gidsg:
@agalamatis OK added in 2139a11

gidsg:
@agalamatis OK added in 2139a11

PR #618, opened 2021-01-26
Unify sending emails to all users of suppliers

bjgill:
Trello: https://trello.com/c/1V1ho9Cx

We want to improve the way in which we send emails. The first step is to try and commonise how we send emails to make future improvements easier.

These two scripts send emails in exactly the same way, so make for a quick start.


bjgill:
Just a quick thing I did in some spare time on Friday. It's a draft because it's not yet ready, but I don't want to forget about it.

bjgill:
We're doing it a different way now (see email engine)

PR #624, opened 2021-02-11
Install awscli in seperate virtual environment to other Python packages

lfdebrux:
Currently we have awscli in our requirements files with other Python packages, however this makes upgrading very hard because awscli always requires a specific version of botocore, and this causes lots of depenendency conflicts with boto3. To get around this we instead install awscli into it's own venv with it's own version of botocore using [pipx].

[pipx]: https://pipxproject.github.io/pipx/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,13 @@ RUN wget --quiet -O /usr/local/bin/docopts.sh https://raw.githubusercontent.com/
     && echo 'd117d3290def71d6a7fdc5b67efddfc3a9299f146bb4f98f96e322222957d1ce /usr/local/bin/docopts.sh' | sha256sum -c - \
     && chmod 0755 /usr/local/bin/docopts.sh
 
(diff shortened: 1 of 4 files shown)


bjgill (commented):
What uses awscli? Are developers going to need to make this same change in their local environments?

gidsg:
>Are developers going to need to make this same change in their local environments?

@bjgill I think it's OK as we suggest a global install of awscli locally https://alphagov.github.io/digitalmarketplace-manual/developing-the-digital-marketplace/developer-setup.html#bootstrapping-a-mac

bjgill (approved):

lfdebrux:
> What uses awscli? Are developers going to need to make this same change in their local environments?

It's used by the script to validate CloudTrail logs https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/validate-cloudtrail-logs.sh

PR #626, opened 2021-02-12
Add oneoff script export-framework-search-filters

lfdebrux:
Ticket: https://trello.com/c/XAgEQjUw/2171-export-list-of-g-cloud-12-filters-and-categories

CCS want to be able to see the current search filters and categories for G12 so they can make improvements for G13. This commit adds a script to collate the information from the frameworks repo into one spreadsheet.

Note to reviewers: a lot of the code is copied from the buyer frontend with minimal changes so it's easier to tell if the buyer frontend diverges in future.

PR #635, opened 2021-02-23
add script to export services top level filter categories

gidsg:
Add script to support data request https://govuk.zendesk.com/agent/tickets/4469876

This borrows heavily from https://github.com/alphagov/digitalmarketplace-scripts/pull/626 for the logic to derive available filters and sub-filters.

### DIFF
--- a/scripts/oneoff/export-service-top-level-categories.py
+++ b/scripts/oneoff/export-service-top-level-categories.py
@@ -1,0 +1,167 @@
+#!/usr/bin/env python3
+
+"""
+List a spreadsheet with services and their associated top-level filter categories
+Note this only works for cloud-software lot as this is the only lot with sub-filter categories which are grouped
+
+Usage:
+    ./scripts/oneoff/export-service-top-level-categories.py [options] <framework_slug> <lot> <stage>
+
+Options:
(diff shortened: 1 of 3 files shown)


scripts/oneoff/export-service-top-level-categories.py
```diff
@@ -0,0 +1,167 @@
...
+        for option in section["filters"]:
+            yield ["", "", str(option["label"]), ""]
+            # category filters can have sub-categories
+            # (but not sub-sub-categories)
+            if "children" in option:
+                for suboption in option["children"]:
+                    yield ["", "", "", suboption["label"]]
+
+# the following functions were copied from the buyer frontend with minor alterations
+# https://github.com/alphagov/digitalmarketplace-buyer-frontend/blob/a83163/app/main/presenters/search_presenters.py
```
lfdebrux, line 91: I know I didn't do this originally, but it might be better if we move these functions to a separate module, seeing as we're using them in more than one place now?

gidsg: @lfdebrux good thinking, I have extracted the shared functions in 922aadd.

lfdebrux: Thanks!

lfdebrux (approved):

PR #636, opened 2021-02-23
use static fixtures for registers lookup

gidsg:
https://trello.com/c/Hey9OS9t/805-2-mitigate-registers-switch-off

In order to mitigate registers switch off, we can store the data as static fixtures to maintain the existing interface.


dmscripts/helpers/supplier_data_helpers.py
```diff
@@ -329,17 +326,25 @@ def get_supplier_ids_from_args(args):
     return supplier_ids
 
 
-@lru_cache()
```
bjgill, line 332: Why are we removing the cache?

gidsg: Ah, good spot 👀 
I removed this as I assumed it was something to do with caching the HTTP response but now I see it is a more generic thing, I have restored it now in 7e05fc0

bjgill (approved):
Changes all look sensible.

Possibly worth recording how you generated the two JSON files? I assume you just downloaded them from Registers today?

gidsg:
> I assume you just downloaded them from Registers today?

@bjgill yes, literally that, I can denote it in an extended commit message as JSON doesn't support comments.

PR #637, opened 2021-02-24
Add script to email G12 suppliers who didn't submit any services

alex9smith:
Borrowed heavily from https://github.com/alphagov/digitalmarketplace-scripts/compare/bg-close-g12

[Notify template](https://www.notifications.service.gov.uk/services/95316ff0-e555-462d-a6e7-95d26fbfd091/templates/7ff14f15-b869-49b1-b2cc-a576fb054a45)

If this script was going to stay around for longer than a few days it should probably be combined with the other email script that notifies suppliers who **did** submit at least one draft service. It's not so I think having two scripts to send the emails is fine.

https://trello.com/c/DteCbA85/828-1-as-a-supplier-i-get-an-email-telling-me-submitted-no-draft-services

image


bjgill (approved):
:tada:

PR #640, opened 2021-02-25
Add script to email G12 suppliers who submitted at least one service

alex9smith:
Borrowed heavily from #637.

[Notify template](https://www.notifications.service.gov.uk/services/95316ff0-e555-462d-a6e7-95d26fbfd091/templates/d8907d3f-164a-4929-bbe6-5d174f04ac2f)
I had to add a new variable to the template to allow for the logic to send the framework agreement message to the one supplier who has not yet signed. I also modified it so we say `1 service` and `2 services`.

https://trello.com/c/0YicABRj/793-2-as-a-supplier-i-get-an-email-telling-me-how-many-draft-services-i-submitted

Email to most suppliers:
image

Email to the supplier who hasn't signed:
image

### DIFF
--- a/scripts/oneoff/notify-g12-suppliers-with-submitted-services.py
+++ b/scripts/oneoff/notify-g12-suppliers-with-submitted-services.py
@@ -1,0 +1,116 @@
+#!/usr/bin/env python
+"""
+Send an email to suppliers involved in the G12 recovery process who completed at least one draft service.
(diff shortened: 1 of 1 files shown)


bjgill (approved):
Looks good.

Quite deeply indented (5), which makes this a bit more difficult to read, but that's not really a problem for a one-use script.

PR #641, opened 2021-02-26
Add script to evaluate G12 recovery supplier declaration

alex9smith:
We have one supplier in the G12 recovery process who has not yet signed the framework agreement. In order for them to be able to do this, we need to evaluate their answers to the declaration questions and mark them as being on G-Cloud 12 if they pass.

This script borrows heavily from `digitalmarketplace-scripts/blob/master/scripts/framework-applications/mark-definite-framework-results.py` to do so for this single supplier.

Note: Their declaration *looks* fine from a quick manual comparison against the schema. Therefore I've not written code to export their declaration answers if they don't pass automatically. If that's needed, we'll probably want to borrow from
`scripts/framework-applications/export-framework-results-reasons.py`

https://trello.com/c/QwO6dH3A/794-1-make-sure-recovery-supplier-not-yet-on-framework-can-sign-framework-agreement

### DIFF
--- a/scripts/oneoff/mark-g12-supplier-status.py
+++ b/scripts/oneoff/mark-g12-supplier-status.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python
+"""
+We have one supplier in the G12 recovery process who has not yet signed the framework agreement. In order for them
+to do this, we need to evaluate their answers to the declaration questions and mark them as being on G-Cloud 12 if
+they pass.
+
+This script borrows heavily from
+digitalmarketplace-scripts/blob/master/scripts/framework-applications/mark-definite-framework-results.py
+to do so for this single supplier.
+
+Usage: mark-g12-supplier-status.py <stage> --updated-by=<updater> --dry-run
+"""
+import sys
+import logging
+import json
+from docopt import docopt
+from dmapiclient import DataAPIClient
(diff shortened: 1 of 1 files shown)


scripts/oneoff/mark-g12-supplier-status.py
```diff
@@ -0,0 +1,81 @@
...
+import json
+from docopt import docopt
+from dmapiclient import DataAPIClient
+from dmutils.env_helpers import get_api_endpoint_from_stage
+sys.path.insert(0, ".")
+
+from dmscripts.mark_definite_framework_results import _passes_validation, pass_supplier
+from dmscripts.helpers.auth_helpers import get_auth_token
+
+UNSIGNED_SUPPLIER_ID = 712034
```
gidsg, line 24: I guess more typically this would be an argument to the script but since this should be really and once ever thing, seems OK to hardcode.

gidsg (approved):

scripts/oneoff/mark-g12-supplier-status.py
alex9smith: Yes, I definitely agree. My thinking here was that I didn't want to run the risk of passing the wrong ID as an argument since we're doing something a bit weird

PR #642, opened 2021-02-26
Only error if all emails invalid

bjgill:
I think the previous behaviour was probably wrong. We do officially support multiple email addresses in this field - the hint text is "You can add more than one email address. Use a comma between each email address."

Users sometimes enter things that aren't emails - for instance names or URLs. This happens quite commonly - I've seen [4 such awarded brief responses in the past month](https://ci.marketplace.team/job/notify-suppliers-of-awarded-briefs-production/).

In no cases have we taken any action as a result of only some of the email addresses being invalid. So I don't think it's useful to fail the script in this case. Instead, downgrade to a warning. We will still error if there is no valid email address.

I've raised a story to improve our input validation: https://trello.com/c/Yld4r8Px/2178-users-can-enter-invalid-email-addresses


gidsg (approved):
Makes sense to me, should cut down a lot of alerting noise 👍

PR #644, opened 2021-03-01
Add framework state to Jenkins functional test report

alex9smith:
Adds a new column `framework_state` to the report. Different tests are run in different framework states, so this will allow us to make like-to-like comparisons.

This is only the approximate framework state as reported by the framework closing, live and expiry dates from the API. We can't get the actual state of the DMp on preview or staging at a point in time because we overwrite the audit events table whenever we apply a data dump from production.

The approximation is probably fine for staging since the state in staging is the same as production the vast majority of the time, and the state in production is managed to the timetable dictated by the dates from CCS.

For preview, we often have the environment one step ahead of staging and production, which means this data may be less helpful. We'll still be able to use it to compare preview builds against other preview builds, but if we're comparing the difference in time taken between different framework states then we'll need to use only data from staging.

https://trello.com/c/oeYMogTp/807-3-measure-functional-test-average-speed-to-provide-baseline

### DIFF
--- a/scripts/oneoff/get-jenkins-functional-test-build-data.py
+++ b/scripts/oneoff/get-jenkins-functional-test-build-data.py
@@ -13,13 +13,20 @@ Options:
     --filename=<file>           Optional. The output filename to use. Defaults to 'functional_test_report.csv'
 """
 
+import sys
 import os
 import csv
 import logging
 import requests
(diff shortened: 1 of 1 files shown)


lfdebrux (approved):
This makes sense to me

PR #645, opened 2021-03-02
Escape dictionary for logging

bjgill:
It needs to be escaped, otherwise CustomLogFormatter tries to double-format it. This is harmless, but it does produce an extra log message: https://github.com/alphagov/digitalmarketplace-utils/commit/d4a0591962d613cdb7f1a3c7fe995bb4f9bab233

Escape the dictionary to reduce the log spam.


lfdebrux (approved):
Weird hack, but cool!

PR #646, opened 2021-03-03
add one off script to notify g12 recovery successful suppliers

gidsg:
https://trello.com/c/FBgpLztJ/845-1-make-successful-services-live

Script to notify successful suppliers as part of G12 recovery.


alex9smith (approved):
LGTM 👍

PR #660, opened 2021-04-14
Add email engine framework for writing scripts that send emails

lfdebrux:
Adds a framework for writing scripts that send emails that aims to be

- Reliable
- Repeatable
- Auditable

The theory is based around a suggestion from @alex9smith and @bjgill, when running the email engine logs email notifications that are outstanding and sent are recorded in a human readable log file. If the script fails partway through sending emails, it can read back its own log file to pick up where it left off.

To use the email engine, the script provides a [generator function](https://wiki.python.org/moin/Generators) that calls the Digital Marketplace API and yields notifications to be sent; `dmscripts.email_engine.email_engine` handles everything else, including resuming failed runs, logging, and command line argument handling:

```python
from dmscripts.email_engine import email_engine

def notifications(stage, **kwargs):
    # generate emails to be sent by calling the DMp API
    data_api_client = DataAPIClient.from_stage(stage)
    for ... in data_api_client.find_some_iter(...):
        yield {
            "email_address": ...,
            "template_id": ...,
            "personalisation": ...
        }

if __name__ == "__main__":
    email_engine(notifications)
```

More examples and details are in the [docstring for `dmscripts.email_engine.__init__`](https://github.com/alphagov/digitalmarketplace-scripts/pull/660/files#diff-ee887f99f520d6592d7f5ba6a1df87ac4376e2a062b8807ff373b00996eb4901).

### Implementation notes

Most of the logic is in the module [dmscripts.email_engine.queue](https://github.com/alphagov/digitalmarketplace-scripts/pull/660/files#diff-9287bcd73e3dca0c283d89224b5f28404ae67b890df97b9576ce400028d56dca), encapsulated in a function called `run`.

### Outstanding

- [x] Add documentation to modules
- [x] Add proper commit messages
- [x] Add example email script using email engine
- [ ] Add example Jenkins job

### DIFF
--- a/dmscripts/email_engine/cli.py
+++ b/dmscripts/email_engine/cli.py
@@ -1,0 +1,159 @@
+"""Standard command line arguments for an email script
+
+For notes on how this module works for a user running a script, have a look at
+the docstring for the email_engine module.
+
+Scripts that send emails tend to have a lot of options in common; this module
+gathers them together to reduce repetition and increase consistency.
+
+The main function is `argument_parser_factory()`: it takes a bunch of keyword
+arguments that configure how the command line will appear. It returns a
+subclass of `argparse.ArgumentParser`, so you can configure the command line
+more if necessary before parsing args.
+
+For some command line options, we want to allow values to be supplied by
+environment variable as well as a flag; the `EnvDefaultAction` class supports
+this.
+
+Running this module will show the default command line arguments::
+    python dmscripts/email_engine/cli.py
+"""
+from pathlib import Path
+import argparse
+import os
+import sys
+
+
+__all__ = ["argument_parser_factory"]
+
+
+def argument_parser_factory(
+    *,
+    reference: str = None,
+    logfile: Path = None,
+    **kwargs
+) -> argparse.ArgumentParser:
+    """Create an ArgumentParser to read options for an email script
+
+    Feel free to add command line arguments to this function, you may want to
+    put them behind a kwarg flag if they are not needed for all scripts.
+    Remember to document them here so that script writers know about them, you
+    can avoid duplication by not adding them to the function signature though.
+
+    :param reference: default reference if none is supplied at command line,
+        if None the script name will be used
+    :param logfile: default path to the logfile if none is supplied at command line,
+        if None /tmp/<reference>.log will be used
+
+    :param description: Text to display before the argument help
+    :param epilog: Text to display after the argument help
+
+    :param stage_required: If true then parser will raise an exception if stage
+        cannot be found from the environment or command line (default: true).
+    :param notify_api_key_required: If true then parser will raise an exception
+        if a Notify API key cannot be found from the environment or command line
+        (default: true).
+    :param notify_template_id_required: Some scripts may expect to a template
+        ID be supplied on the command line, rather than have it hardcoded. If
+        notify_template_id_required is true then the command line arguments will
+        include a required flag to specify the template ID (default: false).
+    """
+
+    p = _ArgumentParser(
+        description=kwargs.get("description"),
+        epilog=kwargs.get("epilog"),
+        formatter_class=argparse.RawDescriptionHelpFormatter
+    )
+
+    p.add_argument(
+        "-n", "--dry-run", action="store_true", help="Do not send notifications."
+    )
+    p.add_argument(
+        "-v",
+        "--verbose",
+        action="store_true",
+        help="Print more detail about what the script is doing.",
+    )
+
+    p.add_argument(
+        "--stage",
+        action=EnvDefault,
+        choices=["local", "preview", "staging", "production"],
+        envvar="DM_ENVIRONMENT",
+        help="Stage/environment to make API calls to. Can also be set with environment variable DM_ENVIRONMENT.",
+        required=kwargs.get("stage_required", True),
+    )
+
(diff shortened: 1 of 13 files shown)


scripts/notify-suppliers-of-awarded-briefs.py
```diff
@@ -1,75 +1,48 @@
...
+Alternatively, when supplied with a list of BriefResponse IDs, it will notify
+just the suppliers for those responses.
+"""
 
+__examples__ = """
 Examples:
-    ./scripts/notify-suppliers-of-awarded-briefs.py preview myToken notifyToken t3mp1at3id --awarded_at=2017-10-27
-    ./scripts/notify-suppliers-of-awarded-briefs.py preview myToken notifyToken t3mp1at3id --dry-run --verbose
-
+    ./scripts/notify-suppliers-of-awarded-briefs.py \\
```
gidsg, line 12: Is the double-slash here intentional?
When I try and run these examples I am just getting the usage prompt e.g.
```
$ ./scripts/notify-suppliers-of-awarded-briefs.py --stage=preview --notify-api-token=notifyToken --notify-template-id=t3mp1at3id --awarded_at=2017-10-27
Usage:
    notify-suppliers-of-awarded-briefs.py <stage> <govuk_notify_api_key> <govuk_notify_template_id>
        [options]
```
Sorry if I'm missing something obvious?

lfdebrux: The double-slash is so that Python doesn't treat it as escaping the newline character, so that when it is printed the newline is present.

The error you're getting is very weird... Are you running the examples with this PR checked out?? 🤔

scripts/notify-suppliers-of-awarded-briefs.py
```diff
@@ -1,75 +1,48 @@
...
+        --stage=preview --notify-api-token=notifyToken --notify-template-id=t3mp1at3id --awarded_at=2017-10-27
```
gidsg, line 13:
```suggestion
        --stage=preview --notify-api-key=notifyToken --notify-template-id=t3mp1at3id --awarded_at=2017-10-27
```
This argument in the example is incorrect.

scripts/notify-suppliers-of-awarded-briefs.py
gidsg: Ah yes, I think I was confused by the force push, it improved once I reset!

scripts/notify-suppliers-of-awarded-briefs.py
lfdebrux: Thanks!

scripts/notify-suppliers-of-awarded-briefs.py
```diff
@@ -1,75 +1,48 @@
...
+just the suppliers for those responses.
+"""
 
+__examples__ = """
 Examples:
-    ./scripts/notify-suppliers-of-awarded-briefs.py preview myToken notifyToken t3mp1at3id --awarded_at=2017-10-27
-    ./scripts/notify-suppliers-of-awarded-briefs.py preview myToken notifyToken t3mp1at3id --dry-run --verbose
-
+    ./scripts/notify-suppliers-of-awarded-briefs.py \\
+        --stage=preview --notify-api-key=notifyToken --notify-template-id=t3mp1at3id --awarded_at=2017-10-27
```
gidsg, line 13:
```suggestion
        --stage=preview --notify-api-key=notifyToken --notify-template-id=t3mp1at3id --awarded-at=2017-10-27
```
argument is incorrect.

lfdebrux: Fixed, thanks

scripts/notify-suppliers-of-awarded-briefs.py
lfdebrux: Ah, I found a better way to get around the escaping, I've made it a raw-string.

gidsg:
From my experience testing this it may be desirable to include the hash of the arguments in the reference, this is because during debugging one may reasonably expect changing the arguments would change the behaviour whereas in the current implementation the reference log overrides argument changes.

lfdebrux:
> From my experience testing this it may be desirable to include the hash of the arguments in the reference, this is because during debugging one may reasonably expect changing the arguments would change the behaviour whereas in the current implementation the existence of the reference log overrides argument changes.

I've added this functionality, thanks for the suggestion.

dmscripts/email_engine/cli.py
```diff
@@ -127,6 +131,16 @@ def argument_parser_factory(
...
 
 
+def append_hash_of_argv(reference: str) -> str:
+    # Add a hash of the command line arguments to the reference so running the
+    # same script with different arguments results in a different reference.
+    args = [reference] + sorted(sys.argv[1:])
+    arghash = hashlib.blake2b(
+        " ".join(args).encode(), digest_size=4
+    ).hexdigest()
+    return reference + "-" + arghash
```
gidsg, line 141: we could use a format string here?

lfdebrux: Changed

gidsg (approved):
Works for me in testing.

Nice improvement 💌 👍

lfdebrux:
Hmm, not sure merging this is a good idea just yet; it may break the Jenkins job to send emails to suppliers about awarded briefs.

PR #661, opened 2021-04-16
Copy uncopied G12 service documents

bjgill:
https://govuk.zendesk.com/agent/tickets/4533790

During the G12 recovery, we published 17 services without copying their service documents to the documents bucket. So the document links did not work.

Find the affected documents, copy them to the correct bucket, and update the affected services with the new URL.

I've done a dry-run of this, and it looked as if it should do the right thing.

Copies extensively from `scripts/framework-applications/publish-draft-services.py`.

### DIFF
--- a/scripts/oneoff/copy-uncopied-service-documents.py
+++ b/scripts/oneoff/copy-uncopied-service-documents.py
@@ -4,13 +4,17 @@ import sys
 from urllib.parse import urljoin
 
 from dmapiclient import DataAPIClient
+from dmutils.env_helpers import (
+    get_api_endpoint_from_stage,
+    get_assets_endpoint_from_stage,
+)
 from dmutils.s3 import S3
 
 sys.path.insert(0, ".")
 
-from dmutils.env_helpers import get_api_endpoint_from_stage, get_assets_endpoint_from_stage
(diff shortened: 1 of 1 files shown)


lfdebrux (approved):
The code you've written makes sense to me, but I am a bit worried about why it was necessary to write it.

Would you mind documenting on this PR why you couldn't use the existing `framework-applications/publish-draft-services.py` script?

bjgill:
You mentioned on Slack that this should maybe be counted as an incident and have an incident review. On reflection, I agree - I think there are some things we can probably learn from this, and that this is probably a P3 incident.

To answer your question: the distinction is subtle, but significant

What `publish-draft-services.py` does is look at all submitted drafts. It publishes each and then copies the documents.

I _could_ run `publish-draft-services.py` again for the affected services. However, I think that would revert the service data to match what was submitted in the draft, which would revert any service edits made in the interim.

So I created this script that only does the file copying and service updating to ensure that I don't need to worry about messing with other fields in the services.

lfdebrux:
@bjgill good reasoning. I hadn't thought of that. Let's merge it 🚀

bjgill:
:+1: It ran successfully on the second attempt - I needed to make a small tweak to get it working. I've added a commit with the changes I made. Merging

PR #666, opened 2021-04-28
Create script to redact data in a brief response

bjgill:
A brief response has data that we need to redact - see https://govuk.zendesk.com/agent/tickets/4540337. To ensure the data is completely removed, we also need to redact all related audit events.

There is no way to do this via the API. We don't want to introduce such a mechanism because it would be a dangerous back-door.

We have agreed this with CDIO IA and also CCS. See the linked Zendesk ticket above for the paper trail.

Note - this is very dangerous. If you want to run this in future, you should only run it after testing on non-production environments, and with the agreement of
IA.

### DIFF
--- a/scripts/oneoff/redact_brief_response.sql
+++ b/scripts/oneoff/redact_brief_response.sql
@@ -1,0 +1,36 @@
+/*
+This script is to redact a brief response and also all related audit messages.
+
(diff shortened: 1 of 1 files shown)


bjgill:
Will test against Preview, and then get it reviewed

bjgill:
Worked against Preview - let's get this merged and then run it live!

gidsg (approved):

PR #668, opened 2021-04-30
Pass notification hash to Notify

alex9smith:
We want to send a unique reference to Notify for every email we send. The reference should be reproducible to allow us to re-create and look up an email's status in our logs at a future date. This is useful when investigating 2nd line requests, or in the face of a potential legal challenge from a supplier.

Scripts not using the email engine already construct this unique reference.

Add a property `EmailNotification.sha256_hash` which computes the SHA256 hash on the string representation of the object.

`EmailNotification` already implements the `__hash__()` method, however this is salted with a random value which changes between Python processes[[1]] and therefore cannot be used to re-create a reference at a future date.

https://trello.com/c/jsjBYAIm/1572-emailengine-does-not-use-unique-reference-hash-per-email

[1]:https://docs.python.org/3/reference/datamodel.html#object.__hash__

### DIFF
--- a/dmscripts/email_engine/__init__.py
+++ b/dmscripts/email_engine/__init__.py
@@ -208,10 +208,9 @@ def email_engine(
         def send_email_notification(
             notification: EmailNotification,
         ) -> NotificationResponse:
-            # note that all notifications will have the same reference
             return notify_client.send_email_notification(
                 **notification,
-                reference=reference,
+                reference=f"{reference}-{notification.sha256_hash}",
             )
 
     try:
--- a/dmscripts/email_engine/typing.py
+++ b/dmscripts/email_engine/typing.py
@@ -10,6 +10,8 @@ reading it back into memory in a human readable fashion.
 from ast import literal_eval
 from typing import Callable, Dict, Generator, Union
 
+from dmutils.email.helpers import hash_string
+
 
 class EmailNotification(dict):
     """A typed, hashable, serder-able, frozen dict subclass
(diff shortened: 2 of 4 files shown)


dmscripts/email_engine/__init__.py
```diff
@@ -208,10 +208,9 @@ def send_email_notification(
         def send_email_notification(
             notification: EmailNotification,
         ) -> NotificationResponse:
-            # note that all notifications will have the same reference
             return notify_client.send_email_notification(
                 **notification,
-                reference=reference,
+                reference=f"{reference}-{notification.sha256_hash}",
```
gidsg, line 213: Do we still want the reference prefix in this case? Would it make future discovery harder if you lost the reference but could derive the hash?

alex9smith: I think we do want the reference, because then in our logs we can see what script sent the email as well as the hash of the data sent to Notify.

dmscripts/email_engine/typing.py
```diff
@@ -61,6 +63,23 @@ def from_str(cls, s: str) -> "EmailNotification":
         """Parse a dict literal representation of a notification"""
         return cls(**literal_eval(s))
 
+    @property
+    def sha256_hash(self) -> str:
```
bjgill, line 67: Why do we need to change the way we create the reference? I thought we already had `dmutils.email.DMNotifyEmail.get_reference`?

alex9smith: The email engine doesn't use `dmutils.email.DMNotifyEmail`, it goes straight to the Notify client. See https://github.com/alphagov/digitalmarketplace-scripts/blob/main/dmscripts/email_engine/__init__.py#L130

alex9smith: I wonder if we could use the one from dmutils though, let me check for the differences

alex9smith: Ok, I think I have it working with `dmutils.email.DMNotifyEmail` and the hash method we already use. Just need to sort out some tests

alex9smith:
I'll mark this as a draft until I've had time to sort out the tests using `dmutils.email.DMNotifyEmail`

alex9smith:
We decided in the tech meeting to use the new method of hashing I've written for now since it's interfering with our ability to log individual email statuses. I'll make a new card on the tech board that covers the switch to `dmutils.email.DMNotifyEmail`

dmscripts/email_engine/__init__.py
bjgill: Gideon - I believe both Kibana and Splunk are clever enough to find "prefix-reference" if you search for "reference"

bjgill (approved):

PR #669, opened 2021-05-06
Get buyers for research who've used a lot

bjgill:
https://trello.com/c/haekzoqW/941-10-recruitment-for-dos-ir35-buyers-of-specialists-research

The admin frontend will give us a list of all buyers who've opted in to being contacted about user research. However, for this particular bit of research, we only want to contact those who've used the Digital Specialists lot.

Just a one-off script for now. If we end up having to run this again, we should consider moving the functionality into admin frontend.


gidsg (approved):
⛽ 👍

PR #679, opened 2021-06-10
Upgrade to python 3.9

gidsg:
https://trello.com/c/PqMIVIGW/13-upgrade-apps-to-python-36-38
Upgrade to python 3.9. Doing this with scripts first as it is lower risk as not a Paas-deployable app.

Have tested the docker image builds and runs correctly locally

## Notes
* You may need to update `pyenv` to pick up the python version
* you may need to delete your `venv` folder to pick up the new python version


bjgill (approved):
Looks good. I assume this will be fine on Jenkins? Are you planning to stop CI against older Python versions?

gidsg:
> I assume this will be fine on Jenkins?

@bjgill good point. For jenkins scripts that use the docker image they will inherit this. For others that just clone and run a script I guess they will still use whatever Jenkins is running.

bjgill:
Are you planning to update Jenkins to 3.9 now? If not, we might want to retain CI for 3.6 for the moment

PR #687, opened 2021-06-16
Stop using python 3.9-only features

bjgill:
In review-dependabot-prs.py, we were using a new parameter that was not available in Python 3.6 (https://docs.python.org/3.6/library/subprocess.html). Revert to the old way of capturing output (that still works in 3.9).

I can confirm that the script now runs on Python 3.6


alex9smith (approved):
This has also fixed the problem that I thought was to do with quotes in the search string - it must have been related to how `subprocess` called `gh`!

bjgill:
Hurrah - I'll add a note to the Trello ticket to suggest the use of this script

PR #692, opened 2021-06-30
Sort approved service edits in the API

bjgill:
Previously, we had to do some rather hacky search to get audit events acknowledged in the interval of interest. This was because they were sorted by creation date by default. We made an assumption that events would always be acknowledged within 5 days, which might be a risky assumption over the end of year holiday period.

Take advantage of a new API feature (https://github.com/alphagov/digitalmarketplace-api/pull/1147) to get the audit events already sorted by the date they were acknowledged. This significantly simplifies the logic, slightly improves performance, and removes some potential edge cases.

Just finishing off something I noticed last time I was on support. I've tested this locally, and it doesn't obviously break anything.

Also move the script to a more sensible directory.

### DIFF
--- a/scripts/export-approved-service-edits.py
+++ b/scripts/export-approved-service-edits.py
@@ -17,7 +17,7 @@ Example:
 
     Count approved service edits in October 2020
 
-    ./scripts/oneoff/export-approved-service-edits.py 2020-10 2020-11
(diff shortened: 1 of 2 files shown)


gidsg (approved):
nice 👍

gidsg:
👀  @bjgill looks like there's a linter error.

PR #694, opened 2021-07-07
Add initial script to generate data for the DMp

alex9smith:
The first stage of my firebreak project.

We've discovered that the API will start against a local database as long as the required tables exist - it doesn't matter if there's no data in any of the tables. This means we can use the API client to insert our randomised data and benefit from all the validation and error handling that the API brings.

Currently this script only adds a single buyer user account, but it sets up a proof of concept for future steps.

https://trello.com/c/RNssJN4U/99-generate-data-for-locally-developing-the-dmp-rather-than-importing-cleaned-production-data

### DIFF
--- a/dmscripts/generate_database_data.py
+++ b/dmscripts/generate_database_data.py
@@ -1,6 +1,5 @@
 import random
 import string
-import os
 from dmapiclient import DataAPIClient
 
 from typing import Optional
@@ -17,7 +16,7 @@ USER_ROLES = [
 ]
 
 
-DEFAULT_PASSWORD = os.getenv("DM_DEFAULT_PASSWORD", default=None)
+DEFAULT_PASSWORD = "Password1234"
 
 
 def generate_email_address(username: Optional[int] = None) -> str:
--- a/dmscripts/generate_database_data.py
+++ b/dmscripts/generate_database_data.py
@@ -1,0 +1,48 @@
+import random
+import string
+import os
+from dmapiclient import DataAPIClient
+
+from typing import Optional
+
+USER_ROLES = [
+    'admin',
+    'admin-ccs-category',
+    'admin-ccs-sourcing',
+    'admin-manager',
+    'admin-framework-manager',
+    'admin-ccs-data-controller',
+    'buyer',
+    'supplier',
+]
+
+
+DEFAULT_PASSWORD = os.getenv("DM_DEFAULT_PASSWORD", default=None)
+
+
+def generate_email_address(username: Optional[int] = None) -> str:
+    """Generate an email of the form 123456@user.marketplace.team"""
+    if username is not None:
+        return f"{username}@user.marketplace.team"
+    else:
+        return f"{random.randrange(1, 10 ** 6):06}@user.marketplace.team"
+
+
+def random_string_of_length(n: int = 10) -> str:
+    return ''.join(random.choice(string.ascii_letters) for _ in range(n))
+
+
+def generate_user(data: DataAPIClient, role: Optional[str] = None) -> dict:
+    """Generate a new user with randomised data and store it in the API"""
+
+    if role not in USER_ROLES:
+        raise ValueError(f"role: {role} not a valid user role")
+    user = {
+        "name": random_string_of_length(10),
+        "email_address": generate_email_address(),
+        "password": DEFAULT_PASSWORD,
+        "role": role,
+    }
+
+    data.create_user(user=user)
+    return user
--- a/dmscripts/generate_database_data.py
+++ b/dmscripts/generate_database_data.py
@@ -32,7 +32,7 @@ def random_string_of_length(n: int = 10) -> str:
     return ''.join(random.choice(string.ascii_letters) for _ in range(n))
 
 
-def generate_user(data: DataAPIClient, role: Optional[str] = None) -> dict:
+def generate_user(data: DataAPIClient, role: str) -> dict:
     """Generate a new user with randomised data and store it in the API"""
 
     if role not in USER_ROLES:
--- a/scripts/generate-database-data.py
+++ b/scripts/generate-database-data.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 """
-Fill a PostgresSQL database with enough randomly generated data to run the DMp. Currently only creates a single buyer
-user account.
+Populate your local database with enough randomly generated data to run the DMp using the API. Currently only creates a
+single buyer user account.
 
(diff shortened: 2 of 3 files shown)


dmscripts/generate_database_data.py
```diff
@@ -0,0 +1,48 @@
...
+USER_ROLES = [
+    'admin',
+    'admin-ccs-category',
+    'admin-ccs-sourcing',
+    'admin-manager',
+    'admin-framework-manager',
+    'admin-ccs-data-controller',
+    'buyer',
+    'supplier',
+]
```
bjgill, line 17: Nit: would an [enum](https://docs.python.org/3/library/enum.html) help here?

dmscripts/generate_database_data.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+DEFAULT_PASSWORD = os.getenv("DM_DEFAULT_PASSWORD", default=None)
+
+
+def generate_email_address(username: Optional[int] = None) -> str:
+    """Generate an email of the form 123456@user.marketplace.team"""
```
bjgill, line 24: Why do we want emails of this form?

dmscripts/generate_database_data.py
```diff
@@ -0,0 +1,48 @@
...
+        return f"{username}@user.marketplace.team"
+    else:
+        return f"{random.randrange(1, 10 ** 6):06}@user.marketplace.team"
+
+
+def random_string_of_length(n: int = 10) -> str:
+    return ''.join(random.choice(string.ascii_letters) for _ in range(n))
+
+
+def generate_user(data: DataAPIClient, role: Optional[str] = None) -> dict:
```
bjgill, line 35: What happens if the role isn't provided?

scripts/generate-database-data.py
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env python3
+"""
+Fill a PostgresSQL database with enough randomly generated data to run the DMp. Currently only creates a single buyer
```
bjgill, line 3: Is this description true, given that you're using the data API?

dmscripts/generate_database_data.py
```diff
@@ -0,0 +1,48 @@
...
+    'admin-ccs-sourcing',
+    'admin-manager',
+    'admin-framework-manager',
+    'admin-ccs-data-controller',
+    'buyer',
+    'supplier',
+]
+
+
+DEFAULT_PASSWORD = os.getenv("DM_DEFAULT_PASSWORD", default=None)
```
bjgill, line 20: What happens if `DM_DEFAULT_PASSWORD` is not in the environment?

tests/test_generate_database_data.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @pytest.mark.parametrize(
+        "role,expected_error", [(role, None if role in USER_ROLES else ValueError) for role in TEST_ROLES])
+    def test_generate_user_validates_roles(self, role, expected_error):
+        if expected_error is not None:
+            with pytest.raises(expected_error):
+                generate_user(self.api_client, role=role)
+
+        else:
+            generate_user(self.api_client, role=role)
```
bjgill, line 29: Would this test be simpler if separated out into two tests for the two cases?

tests/test_generate_database_data.py
```diff
@@ -0,0 +1,29 @@
...
+    def setup_method(self, method):
+        self.api_client_patch = mock.patch("dmapiclient.DataAPIClient", autospec=True)
+        self.api_client = self.api_client_patch.start()
+
+    def teardown_method(self, method):
+        self.api_client_patch.stop()
+
+    def test_generate_user_has_correct_keys(self):
+        buyer = generate_user(data=self.api_client, role="buyer")
+        assert set(buyer.keys()) == {"name", "email_address", "password", "role"}
```
bjgill, line 19: Nit: I don't think the `set` is needed - the output of `keys` is already set-like:

```
In [6]: {1:2, 3:4}.keys() == {3, 1}                                                                                                                                                            
Out[6]: True
```

dmscripts/generate_database_data.py
bjgill: Also, our default non-prod password is no secret (e.g. https://github.com/alphagov/digitalmarketplace-scripts/blob/main/scripts/create_test_users.py#L16). So you could include it here if you wanted to...

bjgill (approved):
Overall approach looks sensible.

A few questions about the details, but nothing serious. Happy for you to merge when you've made whatever changes you think necessary.

dmscripts/generate_database_data.py
alex9smith: That's one of the email domains that admin accounts are allowed to be on a non-production environment (see https://github.com/alphagov/digitalmarketplace-api/blob/main/config.py#L31) and also the one we already change user accounts to when we clean production data - https://github.com/alphagov/digitalmarketplace-aws/blob/main/scripts/clean-db-dump.sql#L6.

I'll add to this docstring to explain why

dmscripts/generate_database_data.py
alex9smith: I think there would be some benefits but, unless I'm misunderstanding, enum can't include dashes so I'd need to have something else as well to convert the enum value into the valid roles. I think that's adding too much complexity for what we gain

dmscripts/generate_database_data.py
alex9smith: Good catch - this is leftover from some earlier refactoring and the role should be required

dmscripts/generate_database_data.py
alex9smith: Ah thanks, I didn't realise it wasn't a secret. I'll just set the value rather than trying to hide it in an environment variable

tests/test_generate_database_data.py
alex9smith: True. Then I can also get rid of the potentially confusing list comprehension in the parameters

scripts/generate-database-data.py
alex9smith: I've updated it. This description is from when we thought the script would need to interact directly with the database

dmscripts/generate_database_data.py
daniele-occhipinti: Am I right the Enum would look like this?
```
class UserRoles(Enum):
    ADMIN = 'admin',
    ADMIN_CSS_SOURCING = 'admin-ccs-sourcing',
   .....
```

PR #697, opened 2021-07-27
Sort dependabot PRs by title

bjgill:
Dependabot PR titles are of the form "Bump X from Y to Z". So by sorting by title, we'll group together updates to the same dependency. This should make it easier to handle all PRs for one common dependency at the same time.


alex9smith (approved):
This is a great idea!

PR #700, opened 2021-08-11
Fix data generation script

alex9smith:
This PR makes two changes that mean `scripts/generate-database-data.py` can now run against a database with tables set up but no data.

Firstly, fix a mistake with the email address key when using the API client to create a user.

Secondly, add a new function to add a domain to the list of approved buyer email domains. This has to happen before we try to add any buyer user accounts.


alex9smith:
@daniele-occhipinti I refactored `create_buyer_email_domain_if_not_present` so we can test it. This introduces a extra API call to get the current allowed domains, but overall I think it's also neater than what we had before.

daniele-occhipinti:
Good refactoring - it looks much better now.

daniele-occhipinti (approved):
LGTM and good test coverage.

PR #702, opened 2021-08-12
Set all frameworks to expired

alex9smith:
Applying only the database migrations to an empty database creates several old frameworks in various states. This causes problems when the frontend apps try to load content for frameworks we don't have content for (eg. G-Cloud 5).

Set all existing frameworks  to expired before we start adding new data.


daniele-occhipinti (commented):
It looks good to me. Just waiting for the dmp-contained env to be up so that I can test it and approve the PR

daniele-occhipinti:
Tested and works.

daniele-occhipinti (approved):
tested and works

PR #704, opened 2021-08-12
Fix adding a buyer email domain

alex9smith:
`data.get_buyer_email_domains_iter` returns a list of dictionaries, not just approved domains, so this function raised an error if the domain name was already in the database.

The structure of the returned dictionary: 
```
IPython 7.12.0 -- An enhanced Interactive Python. Type '?' for help.

local rw
In [1]: [i for i in data.get_buyer_email_domains_iter()]
Out[1]: [{'domainName': 'user.marketplace.team', 'id': 1}]
```


daniele-occhipinti (approved):
Tested the script and the tests and it works.

PR #705, opened 2021-08-12
Add a live g-cloud framework to generated data

alex9smith:
Based on G-Cloud 12 as that's the most recent G-Cloud framework at time of writing.


daniele-occhipinti (approved):
LGTM

PR #706, opened 2021-08-12
Create script to add external collaborator

bjgill:
We have a lot of repos, so manually adding an external collaborator for each is painful. Create this script to invite someone as an external collaborator to all Digital Marketplace repositories. We'll use this to give CCS techops people the ability to create/merge PRs to our repos before we move the repos over to their organisation.

Whilst this will save us a bunch of time, it will still be painful for the person we've invited, since they'll need to accept each invitation manually. Though they could automate this if they wanted to (https://docs.github.com/en/rest/reference/repos#list-repository-invitations-for-the-authenticated-user).


alex9smith (approved):
LGTM

PR #707, opened 2021-08-16
Change test data to open GCloud12 before making it live

daniele-occhipinti:
Problem we faced: with GCloud being live we couldn't add suppliers to the framework. This PR first opens GCloud12 and then make it live. This lays the ground for being able to add suppliers to the framework between the framework being open and live.

I have added/edited unit tests. I've also tested this on an empty db and it seems to work.

### DIFF
--- a/dmscripts/generate_database_data.py
+++ b/dmscripts/generate_database_data.py
@@ -73,7 +73,7 @@ G_CLOUD_FRAMEWORK = {
             'unitSingular': 'service'
         }
     ],
-    'status': 'live',
+    'status': '[STATUS]',
     'name': 'G-Cloud 12',
     'slug': 'g-cloud-12',
     'variations': {}
@@ -121,7 +121,17 @@ def set_all_frameworks_to_expired(data: DataAPIClient) -> None:
             data.update_framework(framework_slug=framework["slug"], data={"status": "expired"})
 
 
-def add_live_g_cloud_framework(data: DataAPIClient) -> None:
+def make_gcloud_12_live(data: DataAPIClient) -> None:
+    data.update_framework(
+        framework_slug=G_CLOUD_FRAMEWORK["slug"],
+        data={
+            "status": "live",
+            "clarificationQuestionsOpen": False
+        }
+    )
+
+
+def open_gcloud_12(data: DataAPIClient) -> None:
     existing_frameworks = [f["slug"] for f in data.find_frameworks().get("frameworks", [])]
     if G_CLOUD_FRAMEWORK["slug"] not in existing_frameworks:
         data.create_framework(
@@ -131,7 +141,7 @@ def add_live_g_cloud_framework(data: DataAPIClient) -> None:
             lots=[lot["slug"] for lot in G_CLOUD_FRAMEWORK["lots"]],
             has_further_competition=False,
             has_direct_award=True,
-            status="live"
+            status="open"
         )
 
         data.update_framework(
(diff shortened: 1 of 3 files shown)


dmscripts/generate_database_data.py
```diff
@@ -73,7 +73,7 @@
             'unitSingular': 'service'
         }
     ],
-    'status': 'live',
```
bjgill, line 76: If we're not using this, does it still need to be here?

dmscripts/generate_database_data.py
```diff
@@ -121,7 +121,17 @@ def set_all_frameworks_to_expired(data: DataAPIClient) -> None:
             data.update_framework(framework_slug=framework["slug"], data={"status": "expired"})
 
 
-def add_live_g_cloud_framework(data: DataAPIClient) -> None:
+def make_gcloud_12_live(data: DataAPIClient) -> None:
+    data.update_framework(
+        framework_slug=G_CLOUD_FRAMEWORK["slug"],
+        data={
+            "status": "live",
+            "clarificationQuestionsOpen": False
```
bjgill, line 129: As per https://alphagov.github.io/digitalmarketplace-manual/content-and-frameworks/framework-lifecycle-for-developers.html#id6, I don't think we need to disable clarification questions when we make the framework live.

tests/test_generate_database_data.py
```diff
@@ -70,40 +71,53 @@ def test_set_all_frameworks_to_expired(self):
         self.api_client.update_framework.assert_any_call("g-cloud-7", {"status": "expired"})
 
 
-class TestCreateGCloudFramework(TestGenerateDataBase):
+class TestOpenGcloud12(TestGenerateDataBase):
```
bjgill, line 74:
```suggestion
class TestOpenGCloud12(TestGenerateDataBase):
```

Minor nit - 'G-Cloud' gets Camel-cased to `GCloud`

tests/test_generate_database_data.py
```diff
@@ -70,40 +71,53 @@ def test_set_all_frameworks_to_expired(self):
...
+
+        self.api_client.update_framework.assert_called_with(
+            framework_slug=G_CLOUD_FRAMEWORK["slug"],
+            data={
+                "clarificationQuestionsOpen": True
+            }
+        )
+
+
+class TestMakeGcloud12Live(TestGenerateDataBase):
```
bjgill, line 113:
```suggestion
class TestMakeGCloud12Live(TestGenerateDataBase):
```

Again, I think this is the preferred way to camel-case G-Cloud

bjgill (commented):
Looks sensible - just a few minor things

dmscripts/generate_database_data.py
daniele-occhipinti: No, it does not need to be there. However, `G_CLOUD_FRAMEWORK` represents a complete json representation of the object as the API returns it. If we remove that one key, that won't be true anymore, so it is not clear what `G_CLOUD_FRAMEWORK` anymore. That's the only reason why I kept it.
Do you think it is a strong enough reason to keep that @bjgill?

tests/test_generate_database_data.py
daniele-occhipinti: Make sense.

dmscripts/generate_database_data.py
bjgill: That's a good reason to keep it, but should we keep it as a valid value in that case?

dmscripts/generate_database_data.py
daniele-occhipinti: You are right. Fixed in https://github.com/alphagov/digitalmarketplace-scripts/pull/707/commits/773a71f9a5f23401c72e5f7486e93d56261db776

dmscripts/generate_database_data.py
daniele-occhipinti: That's reasonable. Fixed in https://github.com/alphagov/digitalmarketplace-scripts/pull/707/commits/a089913c04dc2f4753e4fbb6050cdbe875258759

daniele-occhipinti:
@bjgill I believe I've addressed all your comments. If you are happy, could you approve please?

bjgill (approved):
:rocket:

PR #708, opened 2021-08-17
Tidy up scripts/create_test_users.py

bjgill:
Tidy up a script we used for the DOS 5 performance testing. Stop hard-coding the user - it's no longer necessary, and is confusing people.

Also run a formatter, and mark a potential secret as safe.


daniele-occhipinti (approved):
LGTM. Good tidy-up!

PR #714, opened 2021-08-31
Change Github organisation to CCS

alex9smith:
It seems that `gh` doesn't follow the redirect for the repos we've moved from `alphagov` to `Crown-Commercial-Service` when using it like this. Update the org name so this script still works.


phantas (approved):
LGTM!

PR #716, opened 2021-09-02
Generate email list for 'coming' email

bjgill:
https://trello.com/c/xkosxLyt/128-2-automatically-extract-the-correct-set-of-emails-from-mailchimp

We send an email 1-2 weeks before a new framework opens to let people know about the new framework: https://trello.com/c/qqyRSPut/21-announce-that-new-framework-is-coming-1-2-weeks-before. The process for sending this email is currently very manual (see linked Trello card). For DOS 5, we made a mistake during the manual process. We exported the wrong list, so didn't notify everyone we ought to. Create a script to get the list of email addresses. This should make the process considerably quicker, easier, and safer in future.

I've tested this on Preview - all looks correct. We'll need to update the documentation after this gets merged.

Relies on https://github.com/Crown-Commercial-Service/digitalmarketplace-utils/pull/689

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -45,12 +45,12 @@ importlib-metadata==1.5.0
     #   -c requirements.txt
     #   pluggy
     #   pytest
-ipython==7.12.0
-    # via -r requirements-dev.in
 ipython-genutils==0.2.0
     # via
     #   -r requirements-dev.in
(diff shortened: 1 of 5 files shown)


alex9smith (approved):
Looks good to me! I like the switch to using sets

PR #720, opened 2021-09-16
Get list of DMP repos from API

bjgill:
We no longer use the seal now that we're in the CCS org. So we can instead use the API to get all repos associated with the root DMP GitHub team.

I've tested that this appears to work correctly.


bjgill:
Hmm. This is picking up repos in archived repos. That's inconvenient. Easily fixed, though.

tim-s-ccs (approved):

PR #721, opened 2021-10-06
Create script to review bot PRs

tim-s-ccs:
- Identical to the script for dependabot but changed to allow for multiple bot PRs to be reviewed

As we get PRs for snyk as well as dependabot, I thought it might be worth refactoring the script so that it could review PRs for multiple bots.

The only differences between dependabot PRs and snyk PRs are that they have different authors and slightly different labelling. The config for these bots is kept in the `BOTS_CONFIG` and this is used to differentiate them.

Other than that, the script remains unchanged other than exiting if one of the options given is not in the `BOTS_CONFIG`.

### DIFF
--- a/scripts/review-bot-prs.py
+++ b/scripts/review-bot-prs.py
@@ -1,14 +1,29 @@
 #!/usr/bin/env python
 """
-Show a human all Dependabot PRs that look ready to be merged. If the human accepts, approve and merge the PRs.
+Show a human bot PRs that look ready to be merged. If the human accepts, approve and merge the PRs.
 
 Requires GitHub's CLI tool: https://github.com/cli/cli
+
+Usage:
+    scripts/review-bot-prs.py (snyk | dependabot)
 """
 
 import json
 import subprocess
+from docopt import docopt
 
 ORGANISATION = "Crown-Commercial-Service"
+
+BOTS_CONFIG = {
+    'dependabot': {
+        'login': 'dependabot',
+        'label_str': 'label:dependencies'
+    },
+    'snyk': {
+        'login': 'snyk-bot',
+        'label_str': 'author:snyk-bot'
+    }
+}
 
 
 def get_digital_marketplace_repos():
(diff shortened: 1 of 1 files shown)


scripts/review-bot-prs.py
```diff
@@ -1,15 +1,37 @@
...
+    - dependabot
+    - snyk
 """
 
+import sys
 import json
 import subprocess
+from docopt import docopt
+
+sys.path.insert(0, '.')
```
bjgill, line 20: Is this needed? I think we only need to add this when we're importing from somewhere else within this repo.

scripts/review-bot-prs.py
```diff
@@ -1,15 +1,37 @@
...
 ORGANISATION = "Crown-Commercial-Service"
 
+BOTS_CONFIG = {
+    'dependabot': {
+        'login': 'dependabot',
+        'label_str': 'label:dependencies'
+    },
+    'snyk': {
+        'login': 'snyk-bot',
+        'label_str': 'author:snyk-bot'
```
bjgill, line 31: Cunning :tada:

bjgill (commented):
A great idea. Just a few questions about the implementation details.

scripts/review-bot-prs.py
```diff
@@ -73,6 +95,20 @@ def eligible_for_semiautomated_merge(pr):
...
+
+    BOT = arguments['<bot>']
+
+    if BOT not in BOTS_CONFIG:
+        print("FAILED: bot not in list, available bots are:")
+
+        for bot in BOTS_CONFIG:
+            print(f"  - {bot}")
+
+        exit(1)
```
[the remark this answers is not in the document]
tim-s-ccs: I did not know we could do this, I will implement it

tim-s-ccs:
> A great idea. Just a few questions about the implementation details.

Thank you for the feedback, I have implemented the changes suggested

PR #735, opened 2021-11-17
Update script to get supplier accounts for suppliers who have copyable services

tim-s-ccs:
Commits:
- Make sure only suppliers who have returned the agreement are added
- Add ability to specify a limit for the number of suppliers requires
- Add ability to create file with suppliers

After talking with Toyin, we found a slight issue with the script in that it would include suppliers who had not agreed the previous framework and could therefore not copy services for it. To fix this I added the argument `agreement_returned=True` so we only got valid suppliers.

As Naresh wants a variable number of accounts for each framework, I added the option to specify a limit to the number of supplier accounts. And so that we don't have to copy it all from the console, I changed it to output to a CSV file.

### DIFF
--- a/scripts/get-suppliers-with-copyable-services.py
+++ b/scripts/get-suppliers-with-copyable-services.py
@@ -5,18 +5,28 @@ Get up to 1000 suitable suppliers and then remove their data for the new framewo
 in tests.
 
 Usage:
-    scripts/get-suppliers-with-copyable-services.py <stage> <new-framework> <copy-from-framework>
+    scripts/get-suppliers-with-copyable-services.py <stage> <new-framework> <copy-from-framework> [options]
+
+Options:
+    --limit=<limit>             Limit the number of suppliers to find (default is 1000)
+    --output-dir=<output_dir>   Directory to write to CSV file, if not given it will print to the console
 """
 import sys
 
 from dmapiclient import DataAPIClient, HTTPError
 from dmutils.env_helpers import get_api_endpoint_from_stage
 from docopt import docopt
+import os
 
 sys.path.insert(0, ".")
 
 from dmscripts.helpers.auth_helpers import get_auth_token
 from dmscripts.helpers.updated_by_helpers import get_user
+
+try:
+    import unicodecsv as csv
+except ImportError:
+    import csv
 
 # We don't want to use _all_ the suppliers. There should be ~5000 eligible suppliers.
 SUPPLIER_LIMIT = 1000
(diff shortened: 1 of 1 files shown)


scripts/get-suppliers-with-copyable-services.py
```diff
@@ -5,19 +5,29 @@
 in tests.
 
 Usage:
-    scripts/get-suppliers-with-copyable-services.py <stage> <new-framework> <copy-from-framework>
+    scripts/get-suppliers-with-copyable-services.py <stage> <new-framework> <copy-from-framework> [options]
+
+Options:
+    --limit=<limit>             Limit the number of suppliers to find (default is 1000)
```
bjgill, line 11:
```suggestion
    --limit=<limit>             Limit the number of suppliers to find [default: 1000]
```

Nit: docopt can handle defaults itself: https://github.com/docopt/docopt#option-descriptions-format

bjgill (commented):
Looks good - good catch on the count. Just a couple of docopt nits and a question about the imports.

Was this originally three separate commits? The commit message looks a bit odd. Not a blocker, though.

scripts/get-suppliers-with-copyable-services.py
```diff
@@ -5,19 +5,29 @@
...
 
 sys.path.insert(0, ".")
 
 from dmscripts.helpers.auth_helpers import get_auth_token
 from dmscripts.helpers.updated_by_helpers import get_user
 
+try:
+    import unicodecsv as csv
+except ImportError:
+    import csv
```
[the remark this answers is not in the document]
tim-s-ccs: As I had not written a script to import a CSV before I looked at how it was done in other places. However, having looked and read a bit more I've changed it just to import the `csv` module.

scripts/get-suppliers-with-copyable-services.py
```diff
@@ -27,39 +37,66 @@
 if __name__ == "__main__":
     arguments = docopt(__doc__)
     stage = arguments['<stage>']
+    OUTPUT_DIR = arguments['--output-dir']
+
     if stage == "production":
         raise Exception("This script is not safe to run in production")
+    if not OUTPUT_DIR:
+        raise Exception("You must specify an output directory for the file")
```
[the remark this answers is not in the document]
tim-s-ccs: I have changed it to be mandatory

scripts/get-suppliers-with-copyable-services.py
tim-s-ccs: I did not know I could do this, thanks for the link it was very useful

scripts/get-suppliers-with-copyable-services.py
```diff
@@ -36,30 +40,48 @@
...
-        except HTTPError as e:
-            if not (e.status_code == 404 and "has not registered interest" in e.message):
-                raise
+            count += 1
 
-        try:
-            user = next(u for u in data_api_client.find_users_iter(supplier_id=supplier_id) if u['active'])
-        except StopIteration:
-            continue
+            print(f"Suppliers found: {count}/{SUPPLIER_LIMIT}", end='\r')
```
bjgill, line 84: Why the `end` here?

(not a problem that needs to be fixed - I'm just curious)

bjgill (approved):
Looks good 🎉

scripts/get-suppliers-with-copyable-services.py
tim-s-ccs: It is a counter so I can see the progress. The end means it will just overwrite the current count with the updated count and means we don't get 1000 lines in the terminal.

It looks something like this:
Screenshot 2021-11-18 at 11 12 23

bjgill: That's neat - I didn't know about that. Thanks!

PR #737, opened 2021-11-18
Add NFT to list to access api client shell

tim-s-ccs:
This will allow us to use the `api-clients-shell.py` script in the NFT environment.


bjgill (approved):
🚀

PR #738, opened 2021-11-19
Create script to get suppliers who have not signed the framework agreement

tim-s-ccs:
Using the script in `mark-definite-framework-results.py` I was able to determine some logic to find the suppliers we were after.

I used what I had done in https://github.com/Crown-Commercial-Service/digitalmarketplace-scripts/pull/735 for the spreadsheet generation.

### DIFF
--- a/scripts/get-suppliers-with-unsigned-framework-agreements.py
+++ b/scripts/get-suppliers-with-unsigned-framework-agreements.py
@@ -16,7 +16,6 @@ import sys
 from dmapiclient import DataAPIClient
 from dmutils.env_helpers import get_api_endpoint_from_stage
 from docopt import docopt
-from collections import Counter
 import os
 import csv
 
@@ -27,16 +26,6 @@ from dmscripts.helpers.updated_by_helpers import get_user
 
 
 DEFAULT_PASSWORD = "Password1234"
-
-
-def _assess_draft_services(framework_slug, supplier_id,):
-    # A supplier must have at least 1 submitted service
-    counter = Counter()
-
-    for draft_service in data_api_client.find_draft_services_by_framework_iter(framework_slug, supplier_id=supplier_id):
-        counter[draft_service["status"]] += 1
-
-    return counter
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


scripts/get-suppliers-with-unsigned-framework-agreements.py
```diff
@@ -0,0 +1,99 @@
...
+
+DEFAULT_PASSWORD = "Password1234"
+
+
+def _assess_draft_services(framework_slug, supplier_id,):
+    # A supplier must have at least 1 submitted service
+    counter = Counter()
+
+    for draft_service in data_api_client.find_draft_services_by_framework_iter(framework_slug, supplier_id=supplier_id):
+        counter[draft_service["status"]] += 1
```
bjgill, line 37: This works, so feel free to leave as-is. But the API has a built-in way of doing this: `find_draft_services_by_framework(framework_slug, supplier_id=supplier_id, status="submitted")["meta"]["total"]`

scripts/get-suppliers-with-unsigned-framework-agreements.py
```diff
@@ -0,0 +1,99 @@
...
+
+if __name__ == "__main__":
+    arguments = docopt(__doc__)
+    stage = arguments['<stage>']
+    framework_slug = arguments['<framework_slug>']
+    OUTPUT_DIR = arguments['<output_dir>']
+    SUPPLIER_LIMIT = int(arguments['--limit'])
+
+    if stage == "production":
+        raise Exception("This script is not safe to run in production")
```
bjgill, line 50: Is this true for this script?

bjgill (approved):
Looks sensible - a couple of minor nits, but nothing significant - feel free to merge.

I'm just surprised that there are so many suppliers who're eligible to sign but haven't yet. I guess that's their loss.

scripts/get-suppliers-with-unsigned-framework-agreements.py
tim-s-ccs: Yes, as this will be collecting supplier accounts, we wont want/need a list of supplier emails

scripts/get-suppliers-with-unsigned-framework-agreements.py
tim-s-ccs: I didn't know about this as I have taken the logic from another script. I will implement this as there's no reason to have the repeated code in this case

tim-s-ccs:
> Looks sensible - a couple of minor nits, but nothing significant - feel free to merge.
> 
> I'm just surprised that there are so many suppliers who're eligible to sign but haven't yet. I guess that's their loss.

There actually aren't that many . The limit comes from the request we got but for G-Cloud there are 15 and DOS5 there are 22. But I think that data will be useful to the testers.

bjgill (approved):

PR #740, opened 2021-11-26
Script for getting suppliers who've only applied to older frameworks.

tim-s-ccs:
- Create script for getting suppliers who have not applied to a new framework

Another request for data from the testing team, this script creates a CSV of suppliers who have not applied to a new framework after the last they have applied for.

For example, if you were to run:
```
./scripts/get-suppliers-on-last-specified-framework.py nft g-cloud-8
```

The list of supplier accounts you get will have applied to g-cloud-8 but not:
- g-cloud-9
- g-cloud-10
- g-cloud-11
- g-cloud-12


bjgill (approved):
That's a cunning way of getting the list of suppliers with minimal API requests - I like it.

PR #741, opened 2021-11-30
Create script which will create PRs for the database restoration

tim-s-ccs:
- Create script which will create PRs for the database restoration

This is meant to be used within Jenkins, I'm not 100% sure it will work for the NFT environment but I will try it out.
However, I have checked that all the steps work locally with my own Git config.

I've also made it so that the output is silent expect for the PR it creates.

### DIFF
--- a/scripts/nft-database-backup-config.sh
+++ b/scripts/nft-database-backup-config.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+# This script will create a PR against the aws-nft repository, setting the NFT environemnt to use the new databse.
+#
+# Syntax: ./scripts/nft-database-backup-config-pr.sh [live|maintenance]
+
+set -e
+
+MODE=$(echo "${1}" | tr "[:upper:]" "[:lower:]")
+DB_NAME="digitalmarketplace_api_db"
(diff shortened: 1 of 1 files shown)


bjgill:
> > This looks sensible. You've copied extensively from https://github.com/Crown-Commercial-Service/digitalmarketplace-scripts/blob/main/scripts/maintenance-mode-pr.sh, so I'm happy this is likely to work.
> > The difficult part will be getting credentials for a bot user onto the NFT Jenkins in a secure manner. Do we want to use the same bot user as for DMP 1.0? I was wondering if we wanted to create a new bot user?
> 
> The NFT Jenkins is already using credentials for somewhere, which is how it logs into GPaaS. However, these are different credentials as the login being used belongs to Leo.

Yes - the NFT Jenkins currently has Leo's credentials for GOV.UK PaaS. Do we want to create a dedicated bot user for Github access for the NFT? That would allow us to avoid exposing credentials for the current bot account which has admin access to all the DMP 1.0 repos.

bjgill:
> Yes, we should create a separate bot with separate access. Did you want to do it or shall I?

Feel free - I'm a bit busy today.

PR #747, opened 2021-12-13
Find active suppliers since a specified date

tberey:
This script was needed to find all active registered suppliers, since a given date.

The script parses a date, provided by the user (if not default date of today is provided). This date is then compared against the Suppliers list, where active is true. If the date is more recent than the date the user provided, the record is captured and outputted. A count is taken and also provided.


tberey:
This branch can't yet be merged in, because of an outstanding question:

1. This script outputs a file, with the results in... Perhaps I should just print it all directly to the console/terminal log, rather than out to a txt file?

tberey:
@tim-s-ccs the script itself works, it just seems that either:
`arguments = docopt(__doc__)`, which I am importing and initialising to grab arguments in this way `arguments['--date']`

it cold also be the initialisation of the data object `data_api_client = DataAPIClient(get_api_endpoint_from_stage(environ["STAGE"].lower()), get_auth_token("api", environ["STAGE"].lower()))`

Both of these I have copied and pasted from another script, but errors are gone once these are removed (but they are needed, for the script to work)

bjgill:
I can take a look, but a quick question on the goal before I start: are we trying to get active suppliers created since the date, or all suppliers who have been active since the date?

tberey:
@bjgill we are trying to get active suppliers created since the date

scripts/find-active-suppliers-since-a-date.py
```diff
@@ -0,0 +1,62 @@
...
+    arguments = docopt(__doc__)
+    stage = arguments['<stage>']
+    user_date = arguments['--date']
+
+    data_api_client = DataAPIClient(
+        get_api_endpoint_from_stage(stage), get_auth_token("api", stage), user=get_user()
+    )
+
+    counter = 0
+    date_format = "%Y-%m-%d"
```
bjgill, line 40: Would `datetime_helpers` help here (and elsewhere where you're parsing dates/datetimes)?

scripts/find-active-suppliers-since-a-date.py
```diff
@@ -0,0 +1,62 @@
...
+The script parses a date, provided by the user (if not default date of today is provided).
+This date is then compared against the Suppliers list, where active is true.
+If the date is more recent than the date the user provided, the record is captured and outputted.
+A count is taken and also provided.
+
+Usage:
+    scripts/find-active-suppliers-since-a-date.py <stage> [options]
+
+Options:
+    --date=YYYY-MM-DD       Limit the number of suppliers to find [default: <today's date>]
```
bjgill, line 14: If I try to use the default I get:

```
$ scripts/find-active-suppliers-since-a-date.py preview
ValueError: time data "<today's date>" does not match format '%Y-%m-%d'
```

scripts/find-active-suppliers-since-a-date.py
```diff
@@ -0,0 +1,62 @@
...
+    with open(os.path.join(
+        'data',
+        'suppliers_registered_%s.csv' % datetime.now().strftime("%H:%M:%S")
+    ), 'w', newline='') as file:
+        writer = csv.writer(file, delimiter=',')
+        writer.writerow(["email_address", "name", "supplierName", "supplierID"])
+
+        for user in data_api_client.find_users_iter(role='supplier'):
+            if user['active'] and datetime.strptime(
+                user['createdAt'].split('T', 1)[0], date_format
```
bjgill, line 52: It looks as if we're filtering based on the `createdAt` of the user. What happens if a user is created today for a supplier that's existed for several years? Should they be included?

scripts/find-active-suppliers-since-a-date.py
```diff
@@ -0,0 +1,62 @@
...
+    data_api_client = DataAPIClient(
+        get_api_endpoint_from_stage(stage), get_auth_token("api", stage), user=get_user()
+    )
+
+    counter = 0
+    date_format = "%Y-%m-%d"
+    formatted_date = datetime.strptime(user_date, date_format) or datetime.now().strftime(date_format)
+
+    with open(os.path.join(
+        'data',
```
bjgill, line 44: What happens if the person running this script doesn't have a pre-existing `data` directory?

bjgill (commented):
Did a review of the overall design and structure. A few questions below - mostly a few edge cases I noticed. Once we're happy with that, I can take a more detailed look.

Also two things about the PR:
1. it's not obvious from the PR description why we need to create this script. Could we link to the associated ticket?
2. the commit description is a bit bare. GDS has some good suggestions for how to write a good commit message: https://gds-way.cloudapps.digital/standards/source-code.html#commit-messages

scripts/find-active-suppliers-since-a-date.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python3
+"""
+This script was needed to find all active registered suppliers, since a given date.
```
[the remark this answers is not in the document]
tberey: Would mark as resolved, but it is subjective - I would say it is descriptive enough now, if you agree?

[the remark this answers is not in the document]
tberey: p.s. There was no ticket, this was request by Naresh in a chat, that he needed certain information

scripts/find-active-suppliers-since-a-date.py
tberey: I don't know the overall system well enough to distinguish what the difference means, for the above scenarios.
I'm just using the date of creation for a user that is an active supplier, and it seemed to return what was needed.
The `find_suppliers` payload returned does not contain any date fields.

scripts/find-active-suppliers-since-a-date.py
```diff
@@ -1,17 +1,18 @@
...
+where active is true.
+If the creation date of the user is more recent than the compared date the user provided,
+the record is captured and outputted. A count is also taken and outputted.
 
 Usage:
     scripts/find-active-suppliers-since-a-date.py <stage> [options]
 
 Options:
-    --date=YYYY-MM-DD       Limit the number of suppliers to find [default: <today's date>]
+    --date=YYYY-MM-DD       Limit the number of suppliers to find [default: ]
```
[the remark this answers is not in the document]
tberey: Yeah I didn't realise that until I'd seen the `ValueError: time data "<today's date>" does not match format '%Y-%m-%d'` message - I thought it was just a comment block with dev instructions 😂 

Much appreciated, my last push has fixed that 👍

tberey:
@bjgill last push should have addressed/fixed most, if not all, of the issues you highlighted

bjgill (approved):
As discussed a couple of days ago, I think the new AWS Cloudwatch dashboard now includes this information.

Feel free to merge or close.

PR #756, opened 2022-01-06
Add support for all known non-prod 1.5 environments

bjgill:
There are a lot more non-prod environments in 1.5 than in 1.0. Add support for all the ones I know about. Also change the way the prompt is constructed - we'll no longer need to add each new non-production environment to the prompt explicitly.


bjgill:
Oops. Wrong repo.

PR #779, opened 2022-04-28
Automate creation of new AWS tokens for preview and production

bjgill:
https://trello.com/c/hsMTMFpi/2445-rotate-aws-access-key

We have a process to rotate our AWS tokens, which we do every 3 months: https://crown-commercial-service.github.io/digitalmarketplace-manual/2nd-line-runbook/rotate-keys.html#rotating-aws-access-keys

This isn't too much work, since there are only two AWS environments we need to do it for. However, Apply to Supply adds a new AWS environment for each environment. That will make the rotation process a lot more painful. So it now makes sense to try to automate the process to make it easier to run for more environments.

Enhance rotate-api-tokens.sh to be able to handle creating a new AWS token and adding it into credentials. The user can only have 2 tokens at any time, so delete an inactive token, and panic if there are already two active tokens.

My plan is to use this in a Jenkins job alongside a command to tidy up after we've re-released all the apps.

I created https://github.com/Crown-Commercial-Service/digitalmarketplace-credentials/pull/485 and https://github.com/Crown-Commercial-Service/digitalmarketplace-credentials/pull/486 by running this script.


tim-s-ccs (approved):

bjgill:
I've tested this successfully to rotate the AWS access token for preview.

Once merged, I'll start looking at Jenkinsification.

tim-s-ccs (approved):

